PR #3, opened 2015-08-13
Fixed: ParseUser.getCurrentUser() returns a non authenticated user when local datastore is used

lukas1994:
Repro project: https://www.dropbox.com/s/m6vl215mg3nj8uo/Parse-Starter-Project-1.9.4.zip
1- change keys
2- click SIGN UP
-> make sure a user & a post are created
3- click GETCURRENTUSER
-> make sure user and session are shown
4- click ISAUTHENTICATED()
-> make sure it returns true
5- now click QUERY DATA WITH INCLUDE("USER")
6- repeat step 3 and 4 and notice the issue

### DIFF
--- a/Parse/src/main/java/com/parse/ParseUser.java
+++ b/Parse/src/main/java/com/parse/ParseUser.java
@@ -501,6 +501,16 @@ public class ParseUser extends ParseObject {
   }
 
   @Override
+  /* package */ void setState(ParseObject.State newState) {
+    // Avoid clearing sessionToken when updating the current user's State via ParseQuery result
+    if (isCurrentUser() && getSessionToken() != null
+            && newState.get("sessionToken") == null) {
+      newState = newState.newBuilder().put("sessionToken", getSessionToken()).build();
+    }
+    super.setState(newState);
+  }
+
+  @Override
   /* package */ void validateDelete() {
     synchronized (mutex) {
       super.validateDelete();
--- a/Parse/src/test/java/com/parse/ParseUserTest.java
+++ b/Parse/src/test/java/com/parse/ParseUserTest.java
@@ -1059,6 +1059,41 @@ public class ParseUserTest {
     verify(currentUserController, times(1)).setAsync(user);
   }
 
+  @Test
+  public void testDontOverwriteSessionTokenForCurrentUser() throws Exception {
+    ParseUser.State sessionTokenState = new ParseUser.State.Builder()
+            .sessionToken("sessionToken")
+            .put("key0", "value0")
+            .put("key1", "value1")
+            .isComplete(true)
+            .build();
+    ParseUser.State newState = new ParseUser.State.Builder()
+            .put("key0", "newValue0")
+            .put("key2", "value2")
+            .isComplete(true)
+            .build();
+    ParseUser.State emptyState = new ParseUser.State.Builder().isComplete(true).build();
+
+    ParseUser user = ParseObject.from(sessionTokenState);
+    user.setIsCurrentUser(true);
+    assertEquals(user.getSessionToken(), "sessionToken");
+    assertEquals(user.getString("key0"), "value0");
+    assertEquals(user.getString("key1"), "value1");
+
+    user.setState(newState);
+    assertEquals(user.getSessionToken(), "sessionToken");
+    assertEquals(user.getString("key0"), "newValue0");
+    assertNull(user.getString("key1"));
+    assertEquals(user.getString("key2"), "value2");
+
+    user.setIsCurrentUser(false);
+    user.setState(emptyState);
+    assertNull(user.getSessionToken());
+    assertNull(user.getString("key0"));
+    assertNull(user.getString("key1"));
+    assertNull(user.getString("key2"));
+  }
+
   //endregion
 
   //region testUnlinkFromAsync


grantland:
I'm not convinced that this is where the actual problem lies, since we shouldn't be overwriting `sessionToken` when merging from LDS when the JSON isn't complete. Could you add some unit tests that reproduces the problem so I could see more into why this problem occurs?

lukas1994:
If the server returns the query that contains a reference to the ```currentUser``` the following happens:
The ```ParseDecoder``` calls ```ParseObject.fromJSON()``` with ```isComplete == true```.
Then ```ParseObject.createWithoutData()``` gets called.
If the local datastore is enabled it gets ```currentUser``` from the store (by object id).
Finally, we call ```currentUser.mergeFromServer()``` with ```isComplete == true``` which overwrites ```currentUser``` with the data from the server (which doesn't contain the sessionToken, ...).

There are different ways to fix this. We need to add a special case somewhere: ```ParseUser``` is the best place I guess. Another fix would be to set ```isComplete == false``` appropriately.

What do you think?

lukas1994:
One interesting thing to note is that the server normally returns the ```sessionToken``` if you query for users but if you query for posts that contain a pointer to a user then it doesn't return the ```sessionToken```.

grantland:
If that's the case, it seems like this current modification will have some unintended side affects such as stale data such as keys that were removed on the server persisting after a `fetch`.

We'd probably want to have a more precise fix that is conditional upon the `ParseUser` instancing being current and only persisting the `sessionToken` locally if there was no `sessionToken` in the new state.

Additionally, this should be verified with actual code that hits the server, but the PR should contain a unit test that prevents this from happening in the future.

Parse/src/main/java/com/parse/ParseUser.java
```diff
@@ -501,6 +501,15 @@ private void restoreAnonymity(Map<String, String> anonymousData) {
   }
 
   @Override
+  void setState(ParseObject.State newState) {
```
grantland: For future PRs, we use `/* package */` to designate package protected methods.

Parse/src/main/java/com/parse/ParseUser.java
```diff
@@ -501,6 +501,18 @@ private void restoreAnonymity(Map<String, String> anonymousData) {
...
+  /* package */ void setState(ParseObject.State newState) {
+    // Avoid overwriting session token
+    State.Builder builder = newState.newBuilder();
+    if (isCurrentUser() && getSessionToken() != null &&
+            newState.get("sessionToken") == null) {
+      builder.put("sessionToken", getSessionToken());
+    }
+
+    super.setState(builder.build());
+  }
```
grantland: It seems like we'd be creating unnecessary `State` objects with every `setState()`. Mind only creating a new `State` if necessary?

Parse/src/main/java/com/parse/ParseUser.java
```diff
@@ -501,6 +501,18 @@ private void restoreAnonymity(Map<String, String> anonymousData) {
   }
 
   @Override
+  /* package */ void setState(ParseObject.State newState) {
+    // Avoid overwriting session token
+    State.Builder builder = newState.newBuilder();
+    if (isCurrentUser() && getSessionToken() != null &&
+            newState.get("sessionToken") == null) {
```
grantland: nit: multi-line conditionals should have the conditional `&&` on the new line.

Parse/src/main/java/com/parse/ParseUser.java
```diff
@@ -501,6 +501,18 @@ private void restoreAnonymity(Map<String, String> anonymousData) {
   }
 
   @Override
+  /* package */ void setState(ParseObject.State newState) {
+    // Avoid overwriting session token
```
grantland: Let's add a little more description here. Something like:

    // Avoid clearing sessionToken when updating the current user's State via ParseQuery result

Parse/src/test/java/com/parse/ParseUserTest.java
```diff
@@ -1059,6 +1061,29 @@ public void testUpgradeToRevocableSessionAsync() throws Exception {
...
+  public void testDontOverwriteSessionTokenForCurrentUser() throws Exception {
+    ParseUser.State sessionTokenState = new ParseUser.State.Builder()
+            .sessionToken("sessionToken")
+            .build();
+    ParseUser.State emptyState = new ParseUser.State.Builder().build();
+
+    ParseUser user = ParseObject.from(sessionTokenState);
+    user.setIsCurrentUser(true);
+
+    Assert.assertEquals(user.getSessionToken(), "sessionToken");
```
grantland: nit: you should be able to use static imports.

Parse/src/test/java/com/parse/ParseUserTest.java
```diff
@@ -1059,6 +1061,29 @@ public void testUpgradeToRevocableSessionAsync() throws Exception {
...
+
+    user.setState(emptyState);
+
+    Assert.assertEquals(user.getSessionToken(), "sessionToken");
+
+    user.setIsCurrentUser(false);
+
```
grantland: nit: whitespace

Parse/src/test/java/com/parse/ParseUserTest.java
```diff
@@ -1059,6 +1061,29 @@ public void testUpgradeToRevocableSessionAsync() throws Exception {
...
+    user.setState(emptyState);
+
```
grantland: nit: whitespace

Parse/src/test/java/com/parse/ParseUserTest.java
```diff
@@ -1059,6 +1061,29 @@ public void testUpgradeToRevocableSessionAsync() throws Exception {
...
+            .build();
+    ParseUser.State emptyState = new ParseUser.State.Builder().build();
+
+    ParseUser user = ParseObject.from(sessionTokenState);
+    user.setIsCurrentUser(true);
+
+    Assert.assertEquals(user.getSessionToken(), "sessionToken");
+
+    user.setState(emptyState);
+
```
grantland: nit: whitespace

Parse/src/test/java/com/parse/ParseUserTest.java
```diff
@@ -1059,6 +1061,29 @@ public void testUpgradeToRevocableSessionAsync() throws Exception {
...
+  @Test
+  public void testDontOverwriteSessionTokenForCurrentUser() throws Exception {
+    ParseUser.State sessionTokenState = new ParseUser.State.Builder()
+            .sessionToken("sessionToken")
+            .build();
+    ParseUser.State emptyState = new ParseUser.State.Builder().build();
+
+    ParseUser user = ParseObject.from(sessionTokenState);
+    user.setIsCurrentUser(true);
+
```
grantland: nit: whitespace

grantland:
This is looking much better! Just an optimization and a few minor nits.

Parse/src/main/java/com/parse/ParseUser.java
```diff
@@ -501,6 +501,18 @@ private void restoreAnonymity(Map<String, String> anonymousData) {
...
   @Override
+  /* package */ void setState(ParseObject.State newState) {
+    // Avoid clearing sessionToken when updating the current user's State via ParseQuery result
+    if (isCurrentUser() && getSessionToken() != null
+            && newState.get("sessionToken") == null) {
+      super.setState(newState.newBuilder().put("sessionToken", getSessionToken()).build());
+    }
+    else {
+      super.setState(newState);
+    }
```
grantland: This logic seems a tad messy, could we simplify it to the following?

```java
if (/* ... */) {
  newState = // ...
}
super.setState(newState);
```

Parse/src/test/java/com/parse/ParseUserTest.java
```diff
@@ -1059,6 +1059,25 @@ public void testUpgradeToRevocableSessionAsync() throws Exception {
...
+            .sessionToken("sessionToken")
+            .build();
+    ParseUser.State emptyState = new ParseUser.State.Builder().build();
+
+    ParseUser user = ParseObject.from(sessionTokenState);
+    user.setIsCurrentUser(true);
+    assertEquals(user.getSessionToken(), "sessionToken");
+
+    user.setState(emptyState);
+    assertEquals(user.getSessionToken(), "sessionToken");
```
grantland: Could we add some data in `emptyState` that we would want to merge in and verify that it's getting merged in correctly? A key that was in `sessionTokenState` that should be removed, a key that should be overwritten, and a key that should be added?

grantland:
Sorry, I realized that there are some additional use cases that we should cover in the tests.

grantland:
Looks awesome, thanks!

Also same comments RE commits, etc. as I made on your other diff.

PR #5, opened 2015-08-13
Fix querying for a ParseObject with JSONArray key

grantland:
`OfflineQueryLogic` was missing the proper code path to compare
a value against a `JSONArray`.

Added tests for comparing a value against a `List`.

### DIFF
--- a/Parse/src/test/java/com/parse/OfflineQueryLogicTest.java
+++ b/Parse/src/test/java/com/parse/OfflineQueryLogicTest.java
@@ -510,6 +510,56 @@ public class OfflineQueryLogicTest {
 
   //endregion
 
+  //region compare
+
+  @Test
+  public void testCompareList() throws Exception {
+    ParseObject object = new ParseObject("SomeObject");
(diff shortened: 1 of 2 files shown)


Parse/src/test/java/com/parse/OfflineQueryLogicTest.java
```diff
@@ -510,6 +510,44 @@ public void testMatchesWithin() throws ParseException {
...
+    List<Integer> list = new ArrayList<>();
+    list.add(1);
+    list.add(2);
+    list.add(3);
+    object.put("list", list);
+
+    OfflineQueryLogic logic = new OfflineQueryLogic(null);
+
+    ParseQuery.State<ParseObject> query = new ParseQuery.State.Builder<>("SomeObject")
+        .whereEqualTo("list", 2)
```
hallucinogen: add test `.whereEqualTo("list", 4)` and make sure it returns `false`.

hallucinogen:
LGTM. One nits

PR #25, opened 2015-08-17
Fixed #7 - Keep `WeakReference` to parent `ParseObject` in `ParseRelation`

### LINKED ISSUE #7: Keep `WeakReference` to parent `ParseObject` in `ParseRelation`
Will also need to keep reference to parent `objectId` and `className`, so we can recreate a pointer to parent if parent was GC'd.

sowrabh:
As suggested in the issue, keeping reference to parent objectId, className and recreating parent pointer when the parent is GC'ed.

### DIFF
--- a/Parse/src/main/java/com/parse/ParseRelation.java
+++ b/Parse/src/main/java/com/parse/ParseRelation.java
@@ -8,6 +8,7 @@
  */
 package com.parse;
 
+import java.lang.ref.WeakReference;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


Parse/src/main/java/com/parse/ParseRelation.java
```diff
@@ -193,7 +208,13 @@ public void remove(T object) {
   }
 
   /* package for tests */ ParseObject getParent() {
-    return parent;
+    if(parent == null){
+      return null;
+    }
+    if(parent.get() == null){
+      return ParseDecoder.get().decodePointer(parentClassName, parentObjectId);
```
grantland: Let's just use `ParseObject.createWithoutData(String, String)` instead here so it'll be more concise.

grantland:
This looks awesome! I just have a minor change that you could squash into your current commit and I can merge this in.

sowrabh:
@grantland Thanks for the quick review. Addressed and squashed. Please have a look.

grantland:
Thanks for your contribution!

grantland:
@sowrabh If you're interested we'd love to send you some Parse swag as a thank you for your contribution! Just send me an email with your shipping address and t-shirt size, my email is on my GitHub profile page. Thanks again!

sowrabh:
@grantland Thanks a bunch. Sent you the details on email.

PR #29, opened 2015-08-17
Push Translation

lukas1994:
added locale/language detection
localeIdentifier gets automatically populated whenever an installation is saved/updated

### DIFF
--- a/Parse/src/main/java/com/parse/ParseInstallation.java
+++ b/Parse/src/main/java/com/parse/ParseInstallation.java
@@ -11,10 +11,12 @@ package com.parse;
 import android.content.Context;
 import android.content.pm.PackageInfo;
 import android.content.pm.PackageManager;
+import android.text.TextUtils;
 
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.Locale;
 import java.util.TimeZone;
 
 import bolts.Continuation;
@@ -36,12 +38,13 @@ public class ParseInstallation extends ParseObject {
   private static final String KEY_DEVICE_TOKEN = "deviceToken";
   private static final String KEY_PUSH_TYPE = "pushType";
   private static final String KEY_TIME_ZONE = "timeZone";
+  private static final String KEY_LOCALE = "localeIdentifier";
   private static final String KEY_APP_VERSION = "appVersion";
   /* package */ static final String KEY_CHANNELS = "channels";
 
   private static final List<String> READ_ONLY_FIELDS = Collections.unmodifiableList(
       Arrays.asList(KEY_DEVICE_TYPE, KEY_INSTALLATION_ID, KEY_DEVICE_TOKEN, KEY_PUSH_TYPE,
-          KEY_TIME_ZONE, KEY_APP_VERSION, KEY_APP_NAME, KEY_PARSE_VERSION,
+          KEY_TIME_ZONE, KEY_LOCALE, KEY_APP_VERSION, KEY_APP_NAME, KEY_PARSE_VERSION,
           KEY_APP_IDENTIFIER));
 
   // TODO(mengyan): Inject into ParseInstallationInstanceController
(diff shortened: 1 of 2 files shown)


lukas1994:
see https://github.com/ParsePlatform/Parse-SDK-Android/issues/28

Parse/src/main/java/com/parse/ParseInstallation.java
```diff
@@ -180,25 +184,65 @@ private void updateVersionInfo() {
...
+    String country = locale.getCountry();
+
+    if (TextUtils.isEmpty(language)) {
+      return;
+    }
+
+    // rewrite depreciated two-letter codes
+    if (language.equals("iw")) language = "he"; // Hebrew
+    if (language.equals("in")) language = "id"; // Indonesian
+    if (language.equals("ji")) language = "yi"; // Yiddish
```
grantland: Let's add a unit test to verify that we're correctly persisting these 3 locales.

Parse/src/test/java/com/parse/ParseInstallationTest.java
```diff
@@ -184,6 +194,8 @@ public void testUpdateBeforeSave() throws Exception {
     // Make sure we update device info
     assertEquals("android", installation.getString(KEY_DEVICE_TYPE));
     assertEquals(installationId.get(), installation.getString(KEY_INSTALLATION_ID));
+    // Make sure we update the locale identifier
+    assertEquals("en-US", installation.getString(KEY_LOCALE_IDENTIFIER));
```
grantland: Let's specify a locale with `Locale.setDefaultLocale(Locale)` since we might not always be running on a `en-US` device.

Parse/src/main/java/com/parse/ParseInstallation.java
```diff
@@ -180,25 +184,65 @@ private void updateVersionInfo() {
...
+
+    String localeString = language;
+
+    if (!TextUtils.isEmpty(country)) {
+      localeString = String.format(Locale.US, "%s-%s", language, country);
+    }
```
grantland: Let's add a unit test for this use case.

Parse/src/main/java/com/parse/ParseInstallation.java
```diff
@@ -180,25 +184,65 @@ private void updateVersionInfo() {
...
+    
+    if (!localeString.equals(get(KEY_LOCALE))) {
+      performPut(KEY_LOCALE, localeString);
+    }
```
grantland: Let's also verify that we update the locale if it's already set.

Parse/src/test/java/com/parse/ParseInstallationTest.java
lukas1994: I did this in ```setUp()```. Do you prefer doing it in the test?

grantland: ah missed that! yeah let's do that so it's a bit more contextual.

Parse/src/test/java/com/parse/ParseInstallationTest.java
```diff
@@ -57,6 +59,28 @@ public void setUp() {
   public void tearDown() {
     ParseObject.unregisterSubclass(ParseInstallation.class);
     ParseCorePlugins.getInstance().reset();
+    ParsePlugins.reset();
+  }
+
+  private void mocksForUpdateBeforeSave() {
```
grantland: nit: make `static` and move to the bottom of the class

Parse/src/test/java/com/parse/ParseInstallationTest.java
```diff
@@ -57,6 +59,28 @@ public void setUp() {
   public void tearDown() {
     ParseObject.unregisterSubclass(ParseInstallation.class);
     ParseCorePlugins.getInstance().reset();
+    ParsePlugins.reset();
```
grantland: Set default Locale back to whatever it was before

grantland:
Two minor changes and we're good to go!

grantland:
LGTM deferring to @hallucinogen to merge.

For future PRs, please submit changes as new commits so it's easier to review changes in between updates as well as @ me since I don't get notifications when you update your PR. When it's all ready we'll ask you to squash. Thanks!

PR #32, opened 2015-08-18
Fix: Unable to receive null response from Cloud Code

grantland:
`res.success(null)` from Cloud Code caused `ParseCloud.callFunctionInBackground()` to return a `JSONObject`, which would result in `ClassCastException` in developer code since they would normally expect a `String` or something.

This change fixes that.

/cc @hallucinogen @wangmengyan95 @nlutsenko

### DIFF
--- a/Parse/src/main/java/com/parse/ParseCloudCodeController.java
+++ b/Parse/src/main/java/com/parse/ParseCloudCodeController.java
@@ -8,7 +8,6 @@
  */
 package com.parse;
 
-import org.json.JSONException;
 import org.json.JSONObject;
 
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


Parse/src/main/java/com/parse/ParseCloudCodeController.java
```diff
@@ -46,11 +45,7 @@ public T then(Task<JSONObject> task) throws Exception {
   /* package for test */ Object convertCloudResponse(Object result) {
     if (result instanceof JSONObject) {
       JSONObject jsonResult = (JSONObject)result;
-      try {
-        result = jsonResult.get("result");
-      } catch (JSONException e) {
-        return result;
-      }
```
grantland: `jsonResult.get("result")` would throw a `JSONException`, which caused this function to return `{}` instead of `null`

wangmengyan95:
LGTM

PR #33, opened 2015-08-20
Update ParseLogInterceptor to handle gzip stream

wangmengyan95:
1. Update ParseLogInterceptor to handle GZIP response body stream
2. Decouple ProxyInputStream (Split the proxy logic and log response logic)
3. Clean ParseHttpRequest and ParseHttpResponse
4. Add test

### DIFF
--- a/Parse/src/main/java/com/parse/ParseHttpResponse.java
+++ b/Parse/src/main/java/com/parse/ParseHttpResponse.java
@@ -29,6 +29,10 @@ import java.util.Map;
 
     /* package */ abstract T self();
 
+    public Init() {
+      this.headers = new HashMap<>();
+    }
+
     public T setStatusCode(int statusCode) {
       this.statusCode = statusCode;
       return self();
@@ -50,7 +54,17 @@ import java.util.Map;
     }
 
     public T setHeaders(Map<String, String> headers) {
-      this.headers = Collections.unmodifiableMap(new HashMap<>(headers));
+      this.headers = new HashMap<>(headers);
+      return self();
+    }
+
+    public T addHeaders(Map<String, String> headers) {
+      this.headers.putAll(headers);
+      return self();
+    }
+
+    public T addHeader(String key, String value) {
+      this.headers.put(key, value);
       return self();
     }
 
@@ -84,7 +98,7 @@ import java.util.Map;
     this.content = builder.content;
     this.totalSize = builder.totalSize;
     this.reasonPhrase = builder.reasonPhrase;
-    this.headers = builder.headers;
+    this.headers = Collections.unmodifiableMap(new HashMap<>(builder.headers));
     this.contentType = builder.contentType;
   }
 
--- a/Parse/src/main/java/com/parse/ParseLogInterceptor.java
+++ b/Parse/src/main/java/com/parse/ParseLogInterceptor.java
@@ -14,13 +14,18 @@ import org.json.JSONException;
 import org.json.JSONObject;
 
 import java.io.ByteArrayOutputStream;
-import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.PipedInputStream;
+import java.io.PipedOutputStream;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.Callable;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.ReentrantLock;
+import java.util.zip.GZIPInputStream;
+
+import bolts.Task;
 
 // TODO(mengyan): Add java doc and make it public before we launch it
 /** package */ class ParseLogInterceptor implements ParseNetworkInterceptor {
@@ -37,12 +42,16 @@ import java.util.concurrent.locks.ReentrantLock;
   private final static String KEY_BODY = "Body";
   private final static String KEY_STATUS_CODE = "Status-Code";
   private final static String KEY_REASON_PHASE = "Reason-Phase";
+  private final static String KEY_ERROR = "Error";
   private final static String TYPE_REQUEST = "Request";
   private final static String TYPE_RESPONSE = "Response";
+  private final static String TYPE_ERROR = "Error";
 
   private final static String IGNORED_BODY_INFO = "Ignored";
 
-  private static abstract class Logger {
+  private static final String GZIP_ENCODING = "gzip";
+
+  /* package for tests */ static abstract class Logger {
     public static String NEW_LINE = "\n";
 
     // The reason we need a lock here is because since multiple network threads may write to the
@@ -80,9 +89,16 @@ import java.util.concurrent.locks.ReentrantLock;
 
   private static class LogcatLogger extends Logger {
 
+    private static int MAX_MESSAGE_LENGTH = 4000;
+
     @Override
     public void write(String str) {
-      Log.i(TAG, str);
+      int start = 0;
+      while (start < str.length()) {
+        int end = Math.min(start + MAX_MESSAGE_LENGTH, str.length());
+        Log.i(TAG, str.substring(start, end));
+        start = end;
+      }
     }
 
     @Override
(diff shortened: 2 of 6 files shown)


wangmengyan95:
1. Update ParseLogInterceptor to handle GZIP response body stream
2. Decouple ProxyInputStream (Split the proxy logic and log response logic)
3. Clean ParseHttpRequest and ParseHttpResponse
4. Add test

Parse/src/main/java/com/parse/ParseHttpResponse.java
```diff
@@ -50,7 +54,12 @@ public T setReasonPhase(String reasonPhase) {
     }
 
     public T setHeaders(Map<String, String> headers) {
-      this.headers = Collections.unmodifiableMap(new HashMap<>(headers));
+      this.headers = new HashMap<>(headers);
+      return self();
+    }
```
grantland: Should we change this to:

```java
public T addHeaders(Map<String, String> headers) {
  this.headers.addAll(headers);
  return self();
}
```

Leaving it as `setHeaders` where it can clear all old values seem like it could be confusing.

Parse/src/main/java/com/parse/ParseLogInterceptor.java
```diff
@@ -93,67 +109,60 @@ public void writeLine(String str) {
   }
 
   private static class ProxyInputStream extends InputStream {
+    // Helper stream to proxy the original input stream to other input stream
+    InputStream originalInput;
+    PipedInputStream proxyInput;
+    PipedOutputStream proxyOutput;
```
grantland:
```java
private
```

Parse/src/main/java/com/parse/ParseLogInterceptor.java
```diff
@@ -37,12 +42,16 @@
...
   private final static String KEY_REASON_PHASE = "Reason-Phase";
+  private final static String KEY_ERROR = "Error";
   private final static String TYPE_REQUEST = "Request";
   private final static String TYPE_RESPONSE = "Response";
+  private final static String TYPE_ERROR = "Error";
 
   private final static String IGNORED_BODY_INFO = "Ignored";
 
-  private static abstract class Logger {
+  private static final String GZIP_ENCODING = "gzip";
```
grantland: nit: `ENCODING_GZIP`

Parse/src/main/java/com/parse/ParseLogInterceptor.java
```diff
@@ -168,44 +177,97 @@ private static String formatBytes(byte[] bytes, String contentType) {
     } else if (contentType.contains("text")) {
       return new String(bytes).trim();
     } else {
-      throw new IllegalStateException("We can not print this " + contentType);
+      return "Not supported contentType";
```
grantland: Maybe "[No preview available for contentType]"?

Parse/src/main/java/com/parse/ParseLogInterceptor.java
```diff
@@ -93,67 +109,60 @@ public void writeLine(String str) {
...
-      needsToBePrinted = isContentTypePrintable(response.getContentType());
+      // We need to make sure we read and write proxyInput/Output in separate thread, otherwise
+      // there will be deadlock.
+      Task.call(new Callable<Void>() {
+        @Override
+        public Void call() throws Exception {
+          callback.done(proxyInput, null);
+          return null;
+        }
+      }, ParseExecutors.io());
```
grantland: This seems like an odd hack, do we know why?

Parse/src/main/java/com/parse/ParseLogInterceptor.java
```diff
@@ -240,19 +310,34 @@ private void logRequestInfo(final String requestId,
...
-    String requestContentType = request.getBody().getContentType();
-    if (isContentTypePrintable(requestContentType)) {
-      ByteArrayOutputStream output = new ByteArrayOutputStream();
-      request.getBody().writeTo(output);
-      return formatBytes(output.toByteArray(), requestContentType);
-    } else {
-      return IGNORED_BODY_INFO;
+    // Body
+    if (responseBodyInfo != null) {
+      logger.writeLine(KEY_BODY, responseBodyInfo);
```
grantland: We'll be writing out "Ignored", is this correct?

Parse/src/main/java/com/parse/ParseLogInterceptor.java
grantland: Actually, will we ever hit this in a normal circumstance? If not, shouldn't this still be `IllegalStateException`?

Parse/src/main/java/com/parse/ParseLogInterceptor.java
grantland: Is there a more elegant way to do this? A Callback and background thread seems super hacky.

wangmengyan95: Yep, it is really a little bit hacky. But I do not find a better way to do that.
I check the  [Stetho](https://github.com/facebook/stetho/blob/master/stetho/src/main/java/com/facebook/stetho/inspector/network/GunzippingOutputStream.java) , it seems it also use the a future + callable hacky way to do that. Any good ideas?

Parse/src/main/java/com/parse/ParseHttpResponse.java
wangmengyan95: kk

Parse/src/main/java/com/parse/ParseLogInterceptor.java
wangmengyan95: Yep, ideally we should never go into this state.

Parse/src/main/java/com/parse/ParseLogInterceptor.java
wangmengyan95: Nice catch!!!

wangmengyan95: e... Why? If `responseBodyInfo` is not null, we should just print `responseBodyInfo`.

Parse/src/main/java/com/parse/ParseLogInterceptor.java
grantland: Hrm, the only other option I could think of would be to have the `PipedInputStream` as a property that you must read on a different thread, but it's still pretty ugly, so this is fine.

Parse/src/main/java/com/parse/ParseLogInterceptor.java
grantland: We sometimes pass in `IGNORED_BODY_INFO`, which is `!= null`

wangmengyan95: Sorry misunderstand your first question :), yep, when we get the unprintable body(like a image), we print "Ignored" here. Do you think we need to change this?

Parse/src/main/java/com/parse/ParseLogInterceptor.java
wangmengyan95: Yes, I will revise this if I find some better solution in the future. But let's keep this for now.

Parse/src/main/java/com/parse/ParseLogInterceptor.java
wangmengyan95: For `proxyInput` and `proxyOutput`, since we use them in a try/catch block, it is not so easy to make them final, so I will make them private and make `originalInput` private and final.

Parse/src/main/java/com/parse/ParseLogInterceptor.java
```diff
@@ -93,67 +109,60 @@ public void writeLine(String str) {
   }
 
   private static class ProxyInputStream extends InputStream {
+    // Helper stream to proxy the original input stream to other input stream
+    private final InputStream originalInput;
+    private PipedInputStream proxyInput;
+    private PipedOutputStream proxyOutput;
```
grantland: You can get around the problem with final and try/catch by using a local var in the constructor and setting the instance vars after the try/catch.

grantland:
Last nit, just squash and LGTM

PR #34, opened 2015-08-20
Don't share locks across UI thread and long running processes

grantland:
`setConnected` uses the same lock we use for synchronizing disk I/O, so there's a possibility
that we can block the UI thread on disk I/O, so we're going to bump the lock usage to a
different thread.

Related Bug: https://developers.facebook.com/bugs/838833329491346/


wangmengyan95:
LGTM

PR #42, opened 2015-08-21
Use OKHttp interceptor list when we use ParseOkHttpClient

wangmengyan95:


Parse/src/main/java/com/parse/ParseOkHttpClient.java
```diff
@@ -146,6 +156,124 @@ public ParseOkHttpClient(int socketOperationTimeout, SSLSessionCache sslSessionC
     return okHttpRequestBuilder.build();
   }
 
+  private ParseHttpRequest getParseHttpRequest(Request okHttpRequest) {
```
grantland: `static`?

Parse/src/main/java/com/parse/ParseOkHttpClient.java
```diff
@@ -146,6 +156,124 @@ public ParseOkHttpClient(int socketOperationTimeout, SSLSessionCache sslSessionC
...
+    // Set Header
+    for (Map.Entry<String, List<String>> entry : okHttpRequest.headers().toMultimap().entrySet()) {
+      parseRequestBuilder.addHeader(entry.getKey(), entry.getValue().get(0));
+    }
+
+    // Set Body
+    if (okHttpRequest.body() instanceof CountingOkHttpRequestBody) {
+      CountingOkHttpRequestBody okHttpBody = (CountingOkHttpRequestBody) okHttpRequest.body();
+      parseRequestBuilder.setBody(okHttpBody.getParseHttpBody());
+    }
```
grantland: What if we have a different kind of body?

Parse/src/main/java/com/parse/ParseOkHttpClient.java
```diff
@@ -146,6 +156,124 @@ public ParseOkHttpClient(int socketOperationTimeout, SSLSessionCache sslSessionC
...
+          @Override
+          public ParseHttpResponse proceed(ParseHttpRequest parseRequest) throws IOException {
+            // Use OKHttpClient to send request
+            Request okHttpRequest = ParseOkHttpClient.this.getRequest(parseRequest);
+            Response okHttpResponse = okHttpChain.proceed(okHttpRequest);
+            okHttpResponseCapture.set(okHttpResponse);
+            return getResponse(okHttpResponse);
+          }
+        });
+        final Response okHttpResponse = okHttpResponseCapture.get();
```
grantland: Are we guaranteed this will be set by the time we get here?

Parse/src/main/java/com/parse/ParseOkHttpClient.java
wangmengyan95: We also have `Request getRequest(ParseHttpRequest parseRequest)` which is not a static method, to make it consistent I think non-static is better?

Parse/src/main/java/com/parse/ParseOkHttpClient.java
wangmengyan95: We can guarantee the body here is always `CountingOkHttpRequestBody `, any ideas how we can   do cast here?

Parse/src/main/java/com/parse/ParseOkHttpClient.java
wangmengyan95: Yep, we can guarantee this, since `chain.proceed()` will always be called in `parseNetworkInterceptor.intercept()`.

grantland:
Seems like the tests are failing

grantland:
LGTM

PR #48, opened 2015-08-23
Add ParseStethoInterceptor

wangmengyan95:
Currently rely on manually test, using three http clients and check normal and GZIP response.


Parse/src/main/java/com/parse/ParseStethoInterceptor.java
```diff
@@ -0,0 +1,279 @@
...
+            .setHeaders(response.getAllHeaders())
+            .setReasonPhase(response.getReasonPhrase())
+            .setStatusCode(response.getStatusCode())
+            .setContent(responseStream)
+            .build();
+      }
+    }
+    return response;
+  }
+}
```
grantland: nit: newline at EOF

grantland:
LGTM, let's get Josh looking at this.

In addition, do you think we should put this and `LogcatInterceptor` in a separate module? That way developers don't have to ship this extra code, but can add another Maven dependency when they want to use it?

wangmengyan95:
Discuss offline, will split `ParseLogInterceptor` and `ParseStethoInterceptor` to a separate module.

Parse/src/main/java/com/parse/ParseStethoInterceptor.java
```diff
@@ -0,0 +1,278 @@
+package com.parse;
```
grantland: Add license headers

PR #52, opened 2015-08-25
Remove isLazy property in ParseUser

lukas1994:
We always have the invariant:
```lazy == true``` <==> ```objectId == null && authData.get("anonymous") != null```

So we can remove the additional property.

For testing: to get the "lazy" user behavior we need to make sure that the user has no ```objectId``` and we need to set "anonymous" ```authData``` (see ```setLazy()``` in ```ParseACLTest.java```).

### DIFF
--- a/Parse/src/test/java/com/parse/ParseACLTest.java
+++ b/Parse/src/test/java/com/parse/ParseACLTest.java
@@ -18,6 +18,9 @@ import org.mockito.invocation.InvocationOnMock;
 import org.mockito.stubbing.Answer;
 import org.skyscreamer.jsonassert.JSONCompareMode;
 
+import java.util.HashMap;
+import java.util.Map;
+
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNull;
@@ -141,7 +144,7 @@ public class ParseACLTest {
     ParseACL acl = new ParseACL();
     acl.setReadAccess("userId", true);
     ParseUser unresolvedUser = new ParseUser();
-    unresolvedUser.isLazy = true;
+    setLazy(unresolvedUser);
     acl.setReadAccess(unresolvedUser, true);
     // Mock decoder
     ParseEncoder mockEncoder = mock(ParseEncoder.class);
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: Parse/src/test/java/com/parse/ParseACLTest.java [hunk 1]
```diff
@@ -202,17 +205,18 @@ public void testResolveUserWithNewUser() throws Exception {
...
     ParseUser unresolvedUser = new ParseUser();
-    unresolvedUser.isLazy = true;
+    setLazy(unresolvedUser);
     // This will set the unresolvedUser in acl
     acl.setReadAccess(unresolvedUser, true);
     acl.setWriteAccess(unresolvedUser, true);
 
     unresolvedUser.setObjectId("test");
     acl.resolveUser(unresolvedUser);
+    unresolvedUser.setObjectId("test");
```
grantland: What's the reasoning we need this?

Parse/src/test/java/com/parse/ParseACLTest.java
```diff
@@ -202,17 +205,18 @@ public void testResolveUserWithNewUser() throws Exception {
...
 
     assertNull(acl.getUnresolvedUser());
-    assertFalse(acl.getReadAccess(unresolvedUser));
-    assertFalse(acl.getWriteAccess(unresolvedUser));
+    assertTrue(acl.getReadAccess(unresolvedUser));
+    assertTrue(acl.getWriteAccess(unresolvedUser));
```
grantland: Just to clarify this change, it looks like we were testing an illegal scenario originally, since there's no way for a `ParseUser` to be lazy and have an objectId, right?

grantland:
@lukas1994 just a few questions

Parse/src/test/java/com/parse/ParseACLTest.java
lukas1994: I think the person who wrote the test thought that ```unresolvedUser``` is still unresolved?
It definitely doesn't make sense.

Parse/src/test/java/com/parse/ParseACLTest.java
lukas1994: Sorry, I'll remove this.

grantland:
LGTM

If you could squash and rebase on master, I can merge!

PR #53, opened 2015-08-25
Download ParseFile content to file instead of memory

wangmengyan95:


Parse/src/main/java/com/parse/ParseAWSRequest.java
```diff
@@ -19,14 +19,18 @@
  * Request returns a byte array of the response and provides a callback the progress of the data
  * read from the network.
  */
-/** package */ class ParseAWSRequest extends ParseRequest<byte[]> {
+/** package */ class ParseAWSRequest extends ParseRequest<Void> {
 
-  public ParseAWSRequest(Method method, String url) {
+  // The temp file is used to save the ParseFile content when we fetch it from server
+  private File tempFile;
```
grantland: nit: `final`

Parse/src/main/java/com/parse/ParseFileController.java
```diff
@@ -125,44 +132,53 @@ public void clearCache() {
...
         }
         if (cancellationToken != null && cancellationToken.isCancelled()) {
           return Task.cancelled();
         }
 
+        // Generate the temp file path for caching ParseFile content based on ParseFile's url
+        // The reason we do not write to the cacheFile directly is because there is no way we can
+        // verify if a cacheFile is complete or not. If download is interrupted in the middle, next
+        // time when we download the ParseFile, since cacheFile has already existed, we will return
+        // this incomplete cacheFile
```
grantland: This makes sense since although we use a `TaskQueue` in `ParseFile`, there could potentially be multiple instances of the same `ParseFile` without LDS.

Would it make sense to instead keep a mapping of URL -> `Task<File>` so that we return the `Task<File>` if the same URL is attempted to be fetched multiple times? This would reduce the amount of requests as well as remove the potential error when we update the file pointer when a developer is reading it due to concurrent fetches on the same `ParseFile`.

Parse/src/main/java/com/parse/ParseFileController.java
```diff
@@ -56,6 +56,13 @@ public File getCacheFile(ParseFile.State state) {
     return new File(cachePath, state.name());
   }
 
+  /* package for tests */ File getTempFile(ParseFile.State state) {
+    if (state.url() == null) {
+      return null;
+    }
+    return new File(cachePath, ParseDigestUtils.md5(state.url()));
```
grantland: If we end up keeping this, we'd be clashing on the temporary file name. We'd probably want something unique like `UUID.randomUUID()`.

Parse/src/main/java/com/parse/ParseAWSRequest.java
```diff
@@ -19,14 +19,18 @@
  * Request returns a byte array of the response and provides a callback the progress of the data
  * read from the network.
  */
-/** package */ class ParseAWSRequest extends ParseRequest<byte[]> {
+/** package */ class ParseAWSRequest extends ParseRequest<Void> {
```
grantland: Would it be easier if we made the generic `File`?

Parse/src/main/java/com/parse/ParseFileController.java
wangmengyan95: Yep, it definitely make sense. Will do that.

Parse/src/main/java/com/parse/ParseAWSRequest.java
wangmengyan95: kk

Parse/src/main/java/com/parse/ParseAWSRequest.java
wangmengyan95: Basically, what `ParseAWSRequest` do is to read response stream and write data to a tempFile. Since we have already injected the tempFile (which means the caller already holds the file pointer), I change `byte[]` to `Void`. But if you think return the tempFile is cleaner, I can change this.

Parse/src/main/java/com/parse/ParseFileController.java
wangmengyan95: Can we guarantee the url of a ParseFile should always be different? If so, why we will have clash. But I will make the change, since UUID is definitely much safer. Just curious about the question.
For clash, I think 
`  public File getCacheFile(ParseFile.State state) {
    return new File(cachePath, state.name());
  }`
will have clash, since for ParseFile create without a given name, `state.name` will always be file. Is that correct？

Parse/src/main/java/com/parse/ParseFileController.java
wangmengyan95: It is useful but we have already had cacheFile. Before we fetch, we check whether the cacheFile exist or not. So instead of checking file exists or not in disk, you want to keep a in-memory map to make the checking faster?

grantland: Never mind, I thought this comment was with regards to asynchronously downloading a `ParseFile`, but it's not.

Parse/src/main/java/com/parse/ParseFileController.java
grantland: The current implementation is fine, but why not just use state.url + (.tmp) or something? We should also remove all temp files for a given file that exist at some point, maybe when we try to re-download?

wangmengyan95: Here I just follow what we do for iOS, but I think state.url + (.tmp) is fine. I will use this.
Clear all tmp before re-download make sense, will do that.

Parse/src/main/java/com/parse/ParseAWSRequest.java
```diff
@@ -19,14 +19,18 @@
...
+  private final File tempFile;
+
+  public ParseAWSRequest(Method method, String url, File tempFile) {
     super(method, url);
+    this.tempFile = tempFile;
   }
 
   @Override
-  protected Task<byte[]> onResponseAsync(ParseHttpResponse response,
+  protected Task<Void> onResponseAsync(ParseHttpResponse response,
```
grantland: We should probably do the operations in this method on `ParseExecutors.io()` since we're doing disk i/o now or else we'd be on `ParseRequest.NETWORK_EXECUTOR` by default.

Parse/src/main/java/com/parse/ParseFileController.java
```diff
@@ -56,6 +56,13 @@ public File getCacheFile(ParseFile.State state) {
     return new File(cachePath, state.name());
   }
 
+  /* package for tests */ File getTempFile(ParseFile.State state) {
+    if (state.url() == null) {
+      return null;
+    }
+    return new File(cachePath, state.url()+".tmp");
```
grantland: nit: spaces around `+`

Parse/src/main/java/com/parse/ParseFileController.java
```diff
@@ -125,44 +132,62 @@ public void clearCache() {
...
+        // this incomplete cacheFile
+        final File tempFile = getTempFile(state);
+
+        return Task.call(new Callable<Void>() {
+          @Override
+          public Void call() throws Exception {
+            // If the last download is interrupted, there may be incomplete temp file in disk. So
+            // we need to clear it first.
+            ParseFileUtils.deleteQuietly(tempFile);
+            return null;
```
grantland: Let's actually remove this part and instead come up with a proper solution to deal with the staging and cache.

grantland:
It seems like we're going to have to have a proper solution for managing our cache. I'll create a new issue for it.

grantland:
Created #98.

Might be worthwhile to move staging into it's own directory so that we can keep cache size separate of cache size. In addition it would allow us to merge this without needing to depend on #98 since the cache manager would clear staging files without a one-off condition.

Parse/src/main/java/com/parse/ParseAWSRequest.java
grantland: ping ^

wangmengyan95: oops, forgot this point, sorry.

grantland:
Can you rebase master? It looks like there are conflicts with master.

grantland:
LGTM

PR #56, opened 2015-08-25
Refactored ParseACL to use POJOs instead of JSON objects

lukas1994:


lukas1994:
The new ```Permissions``` class got quite big, should we put it in a new file @grantland ?

Parse/src/main/java/com/parse/ParseACL.java
```diff
@@ -60,21 +61,63 @@ public static void setDefaultACL(ParseACL acl, boolean withAccessForCurrentUser)
...
+    /* package */ static Permissions createPermissionsFromJSONObject(JSONObject object) {
+      boolean read = object.optBoolean(READ_PERMISSION, false);
+      boolean write = object.optBoolean(WRITE_PERMISSION, false);
+      return new Permissions(read, write);
+    }
+
+    /* package */ boolean isEmpty() {
+      return !(read || write);
+    }
+  }
```
grantland: Could we move this static class either above the instance variables or at the bottom of this class?

grantland:
The `Permissions` class seems fine, esp since the abstraction is isolated to `ParseACL`.

I do think it might be worthwhile to make `Permissions` immutable, sorry I didn't have this comment in your initial inquiry. Thoughts @hallucinogen?

lukas1994:
Yeah that's a good idea. I'll make it immutable.

grantland:
a nit a small optimization

grantland:
LGTM, please squash and I can merge!

PR #63, opened 2015-08-26
Revert "Remove isLazy property in ParseUser"

grantland:
Reverts ParsePlatform/Parse-SDK-Android#52

This has an unintended consequence of breaking resolving lazy users via [`signUpAsync`][1] and [`linkWithAsync`][2]. This is due to stripping anonymity [here][3] and [here][4] before calling into [`saveAsync` which has a conditional for `isLazy()`][5], which now wrongly return `false` because we stripped anonymity.

Reverting for now, but a better solution would be to clean up the `resolveLaziness` mess.

[1]: https://github.com/ParsePlatform/Parse-SDK-Android/blob/master/Parse/src/main/java/com/parse/ParseUser.java#L634
[2]: https://github.com/ParsePlatform/Parse-SDK-Android/blob/master/Parse/src/main/java/com/parse/ParseUser.java#L1238
[3]: https://github.com/ParsePlatform/Parse-SDK-Android/blob/master/Parse/src/main/java/com/parse/ParseUser.java#L630
[4]: https://github.com/ParsePlatform/Parse-SDK-Android/blob/master/Parse/src/main/java/com/parse/ParseUser.java#L1229
[5]: https://github.com/ParsePlatform/Parse-SDK-Android/blob/master/Parse/src/main/java/com/parse/ParseUser.java#L480

### DIFF
--- a/Parse/src/main/java/com/parse/CachedCurrentUserController.java
+++ b/Parse/src/main/java/com/parse/CachedCurrentUserController.java
@@ -241,6 +241,8 @@ import bolts.Task;
                 if (current != null) {
                   synchronized (current.mutex) {
                     current.setIsCurrentUser(true);
+                    current.isLazy = current.getObjectId() == null
+                        && ParseAnonymousUtils.isLinked(current);
                   }
                   return current;
                 }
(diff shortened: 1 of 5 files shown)


hallucinogen:
LGTM. Will reopen internal task.

lukas1994:
alternative: https://github.com/ParsePlatform/Parse-SDK-Android/pull/72

grantland:
Abandoned in favor of #72

PR #70, opened 2015-08-26
Add ParseFileHttpBody and ParseCountingFileHttpBody

wangmengyan95:
1. Add `ParseFileHttpBody` and test
2. Add `ParseCountingFileHttpBody` and test
3. Revise `ParseHttpBody.getContent()`, right now it should throw `IOException`
4. Change all size related variables in `ParseRequest` and `ParseResponse` from `int` to `long`

### DIFF
--- a/Parse/src/main/java/com/parse/ParseApacheHttpClient.java
+++ b/Parse/src/main/java/com/parse/ParseApacheHttpClient.java
@@ -212,7 +212,7 @@ import java.util.Map;
   private static class ParseApacheHttpEntity extends InputStreamEntity {
     private ParseHttpBody parseBody;
 
-    public ParseApacheHttpEntity(ParseHttpBody parseBody) {
+    public ParseApacheHttpEntity(ParseHttpBody parseBody) throws IOException {
       super(parseBody.getContent(), parseBody.getContentLength());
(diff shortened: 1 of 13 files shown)


Parse/src/main/java/com/parse/ParseApacheHttpClient.java
```diff
@@ -212,7 +212,7 @@ public ParseApacheHttpClient(int socketOperationTimeout, SSLSessionCache sslSess
   private static class ParseApacheHttpEntity extends InputStreamEntity {
     private ParseHttpBody parseBody;
 
-    public ParseApacheHttpEntity(ParseHttpBody parseBody) {
+    public ParseApacheHttpEntity(ParseHttpBody parseBody) throws IOException {
```
grantland: Any reason for this change?

wangmengyan95: Since parseBody.getContent() will try to create a fileStream, so we need to throw IOException here.

grantland: :+1:

grantland:
LGTM

PR #71, opened 2015-08-26
Clean warnings in ParseUserTest

grantland:


Parse/src/test/java/com/parse/ParseUserTest.java
```diff
@@ -165,12 +165,12 @@ public void testSignUpAsyncWithObjectIdSetAndAuthDataSet() throws Exception {
     ParseUser partialMockUser = spy(user);
     doReturn(Task.<Void>forResult(null))
         .when(partialMockUser)
-        .saveAsync(anyString(), any(Task.class));
+        .saveAsync(anyString(), Matchers.<Task<Void>>any());
```
wangmengyan95: What is the warning and why use Matchers will solve the warning?

grantland: Incorrect generic since `saveAsync` requires `Task<Void>` but `any(Task.class)` is equivalent to `Task`.

PR #72, opened 2015-08-27
Fix signUp and linkWith with an anonymous user

grantland:
Fixes integration test failures caused by #52

We strip anonymity in `signUp` and `linkWith` before calling into
`saveAsync` where there's a fork in logic for `isLazy()`. Since we now
determine `isLazy()` based off `objectId != null &&
ParseAnonymousUtils.isLinked(this)`,
`isLazy()` returns the incorrect value for these logic paths.

The solution is to get the value of `isLazy()` before stripping anonymity
and passing it through to `saveAsync()`.

* Callers of resolveLazinessAsync never use it's return value, might as well make it `Void`
* Make signUp more robust by reverting current user's password on failure if that's how it originally was
* Make signUp more robust by reverting the current user's password on success

### DIFF
--- a/Parse/src/main/java/com/parse/ParseUser.java
+++ b/Parse/src/main/java/com/parse/ParseUser.java
@@ -475,26 +475,28 @@ public class ParseUser extends ParseObject {
 
   @Override
   /* package */ Task<Void> saveAsync(String sessionToken, Task<Void> toAwait) {
-    synchronized (mutex) {
-      Task<Void> task;
-      if (isLazy()) {
-        task = resolveLazinessAsync(toAwait).makeVoid();
-      } else {
-        task = super.saveAsync(sessionToken, toAwait);
-      }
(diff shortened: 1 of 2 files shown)


grantland:
Still writing a unit test that will simulate this behavior for the future.

/cc @lukas1994 @hallucinogen

grantland:
So it doesn't seem like there's a good way to write a unit test as the scenario is more of a integration test scenario, especially since I added `saveAsync(sessionToken, isLazy, toAwait)`.

lukas1994:
LGTM
Do we have integration tests for this?

hallucinogen:
So this is basically the same as our initial behavior, except `isLazy` is being injected into `saveAsync`?

grantland:
@lukas1994 yeah we have integration tests for this. they're currently failing and this makes them pass

@hallucinogen pretty much

hallucinogen:
Sounds good then.

PR #74, opened 2015-08-27
Update ParseCountingByteArrayHttpBodyTest

wangmengyan95:
We forget to assert when we check the Semaphore


grantland:
Good catch!

PR #77, opened 2015-08-27
Implement ParseAuthenticationManager

grantland:
* Decouples managing ParseAuthentication providers, convert deauthenticate to
async.
* Convert ParseAuthenticationProvider into a interface.

### DIFF
--- a/Parse/src/main/java/com/parse/ParseAuthenticationManager.java
+++ b/Parse/src/main/java/com/parse/ParseAuthenticationManager.java
@@ -1,0 +1,77 @@
+/*
+ * Copyright (c) 2015-present, Parse, LLC.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+package com.parse;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import bolts.Continuation;
+import bolts.Task;
+
+/*** package */ class ParseAuthenticationManager {
+
+  private final Object lock = new Object();
+  private final Map<String, ParseAuthenticationProvider> authenticationProviders = new HashMap<>();
+  private final ParseCurrentUserController controller;
+
+  public ParseAuthenticationManager(ParseCurrentUserController controller) {
+    this.controller = controller;
+  }
+
+  public void register(ParseAuthenticationProvider provider) {
+    final String authType = provider.getAuthType();
+    if (authType == null) {
+      throw new IllegalArgumentException("Invalid authType: " + null);
+    }
+
+    synchronized (lock) {
+      if (authenticationProviders.containsKey(authType)) {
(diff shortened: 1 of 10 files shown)


Parse/src/main/java/com/parse/ParseAuthenticationManager.java
```diff
@@ -0,0 +1,77 @@
...
+  private final Object lock = new Object();
+  private final Map<String, ParseAuthenticationProvider> authenticationProviders = new HashMap<>();
+  private final ParseCurrentUserController controller;
+
+  public ParseAuthenticationManager(ParseCurrentUserController controller) {
+    this.controller = controller;
+  }
+
+  public void register(ParseAuthenticationProvider provider) {
+    final String authType = provider.getAuthType();
```
nlutsenko: Maybe add an exception around `authType` should not be `nil`?

Parse/src/main/java/com/parse/ParseAuthenticationManager.java
```diff
@@ -0,0 +1,77 @@
...
+        if (user != null) {
+          user.synchronizeAuthData(authType);
+        }
+        return null;
+      }
+    });
+  }
+
+  /* package for tests */ void unregister(ParseAuthenticationProvider provider) {
+
```
nlutsenko: Looks like this is missing implementation.

Parse/src/main/java/com/parse/ParseUser.java
```diff
@@ -1265,13 +1252,9 @@ public ParseUser then(Task<Void> task) throws Exception {
...
-  private void logOutWith(String authType) {
-    synchronized (mutex) {
-      ParseAuthenticationProvider provider = authenticationProviders.get(authType);
-      if (provider != null && isLinked(authType)) {
-        provider.deauthenticate();
-      }
-    }
+  /* package */ Task<Void> linkWithAsync(
+      String authType, Map<String, String> authData) {
+    return linkWithAsync(authType, authData, getSessionToken());
```
nlutsenko: Maybe enforce `authType` and `authData` to be non-nil?

Parse/src/main/java/com/parse/ParseUser.java
```diff
@@ -1255,8 +1240,10 @@ public ParseUser then(Task<Void> task) throws Exception {
     }
   }
 
-  /* package */ Task<Void> linkWithAsync(
-      final String authType, final Map<String, String> authData, final String sessionToken) {
+  private Task<Void> linkWithAsync(
+      final String authType,
+      final Map<String, String> authData,
+      final String sessionToken) {
     return taskQueue.enqueue(new Continuation<Void, Task<Void>>() {
```
nlutsenko: Maybe enforce authType and authData to be non-nil?
Also, not sure why we need this method, since we don't seem to be using it much often.

nlutsenko:
Just few nits, nothing big, love the approach.

nlutsenko:
LGTM, Awesome!

Issue #76, opened 2015-08-27
ParseObjectCoder.encode is not thread-safe


hallucinogen:
`ParseObjectCoder.encode` signature accepts `ParseObject.State`, `ParseOperationSet` and `ParseEncoder`. While `ParseObject.State` and `ParseEncoder` is immutable and thread-safe, `ParseOperationSet` can be mutated anytime and is not guarded by any mutex. We need to either guard the `ParseOperationSet` with mutex or make a copy of this `ParseOperationSet` somewhere and pass the copy to `ParseObjectCoder.encode`.

hallucinogen:
This is an example of how `ParseObjectCoder.encode` is not thread-safe: https://github.com/ParsePlatform/Parse-SDK-Android/blob/master/Parse/src/main/java/com/parse/NetworkUserController.java#L41

lukas1994:
We could make a deep-copy of the operation set in ```encode()``` but that might not be very efficient. What do you think?

hallucinogen:
Deep-copy won't solve the problem I mentioned in #78.

lukas1994:
from https://github.com/ParsePlatform/Parse-SDK-Android/pull/78 :
I checked all places where ```ParseOperationSet``` is used and couldn't find any concurrency issues.
The scenario you described above can't happen because if ```save()``` gets called a new ```ParseOperationSet``` gets enqueued (to ```operationSetQueue```) and all methods mutate the top of the queue.

hallucinogen:
Sounds good. Closing this one

PR #78, opened 2015-08-28
Made ParseOperationSet thread-safe.

lukas1994:
See: https://github.com/ParsePlatform/Parse-SDK-Android/issues/76

### DIFF
--- a/Parse/src/main/java/com/parse/ParseOperationSet.java
+++ b/Parse/src/main/java/com/parse/ParseOperationSet.java
@@ -11,16 +11,16 @@ package com.parse;
 import org.json.JSONException;
 import org.json.JSONObject;
 
-import java.util.HashMap;
 import java.util.Iterator;
 import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * A set of field-level operations that can be performed on an object, corresponding to one command.
  * For example, all of the data for a single call to save() will be packaged here. It is assumed
  * that the ParseObject that owns the operations handles thread-safety.
  */
-/** package */ class ParseOperationSet extends HashMap<String, ParseFieldOperation> {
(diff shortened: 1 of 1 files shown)


hallucinogen:
Not sure this is the correct solution to #76. `ParseOperationSet` docs mentioned that `ParseOperationSet` is not thread-safe so we can enforce thread-safety (and thus, atomicity) from its usage while keeping `ParseOperationSet` data structure fast and simple.

Imagine this scenario
1. Make an object: `obj.put("corgi", "awesome")`. It will create `ParseSetOperation { "corgi" => "awesome" }`
2. `obj.saveInBackground()`. It will pass the `operationSet` to `NetworkObjectController`.
3. Somehow (this is the one that I'm not sure of) mutate the `operationSet` from another thread: `ParseSetOperation { "corgi" => "cute" }`
4. Save finish.

The expected result is to have an object saved with `obj.get("corgi")` equals to `"awesome"` but in this case it will return `"cute"`. This PR will make sure that `ParseOperationSet` will never be in invalid state (i.e. in the middle of mutation), which is great. But it doesn't answer the possible race condition problem.

So the question is more: is the race condition possible to begin with? If it isn't, then we don't have to change anything. If it's, then that `ParseOperationSet` **usage** is what we need to fix, not the `ParseOperationSet` itself.

lukas1994:
OK I'll check all places where ```ParseOperationSet``` is used.

lukas1994:
I checked all places where ```ParseOperationSet``` is used and couldn't find any concurrency issues.
The scenario you described above can't happen because if ```save()``` gets called a new ```ParseOperationSet``` gets enqueued (to ```operationSetQueue```) and all methods mutate the top of the queue.

lukas1994:
It doesn't make sense to make the ```ParseOperationSet``` thread-safe, the synchronization will just decrease efficiency.

PR #82, opened 2015-08-28
Add ParseHttpResponse state builder constructor

wangmengyan95:
1. Add ParseHttpResponse state builder constructor and unit test
2. Replace all reasonPhase with reasonPhrase


hallucinogen:
surewhynot

PR #89, opened 2015-08-31
Move Method enum to ParseHttpRequest

wangmengyan95:
In order to public `ParseHttpRequest`, we need to public `Method` enum for http method.
One way is move `Method` enum from `ParseRequest` to `ParseHttpRequest`, but this involves lots of code changes.
The other way is to add `setMethodStr()` and `getMethodStr()` for `ParseHttpRequest`, this does not involve so many code changes, but make `ParseHttpRequest` looks ugly.
Any thoughts @grantland @hallucinogen ?

### DIFF
--- a/Parse/src/main/java/com/parse/ParseHttpRequest.java
+++ b/Parse/src/main/java/com/parse/ParseHttpRequest.java
@@ -10,12 +10,59 @@ package com.parse;
 
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.Iterator;
 import java.util.Map;
 
 /** package */ class ParseHttpRequest {
+
+  public enum Method {
+    GET, POST, PUT, DELETE;
+
+    public static Method fromString(String string) {
+      Method method;
+      switch (string) {
+        case "GET":
+          method = GET;
+          break;
+        case "POST":
+          method = POST;
+          break;
+        case "PUT":
+          method = PUT;
+          break;
+        case "DELETE":
+          method = DELETE;
+          break;
+        default:
+          throw new IllegalArgumentException("Invalid http method: <" + string + ">");
+      }
+      return method;
+    }
+
+    @Override
+    public String toString() {
+      String string;
+      switch (this) {
+        case GET:
+          string = "GET";
+          break;
+        case POST:
+          string = "POST";
+          break;
+        case PUT:
+          string = "PUT";
+          break;
+        case DELETE:
+          string = "DELETE";
+          break;
+        default:
+          throw new IllegalArgumentException("Invalid http method: <" + this+ ">");
+      }
+      return string;
+    }
+  }
+
   private final String url;
-  private final ParseRequest.Method method;
+  private final ParseHttpRequest.Method method;
   private final Map<String, String> headers;
   private final ParseHttpBody body;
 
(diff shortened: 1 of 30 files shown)


Parse/src/main/java/com/parse/ParseHttpRequest.java
```diff
@@ -14,8 +14,56 @@
...
+        case "POST":
+          method = POST;
+          break;
+        case "PUT":
+          method = PUT;
+          break;
+        case "DELETE":
+          method = DELETE;
+          break;
+        default:
```
grantland: Is the correct default `null` or `throw`?

Parse/src/main/java/com/parse/ParseHttpRequest.java
```diff
@@ -14,8 +14,56 @@
...
+        case POST:
+          string = "POST";
+          break;
+        case PUT:
+          string = "PUT";
+          break;
+        case DELETE:
+          string = "DELETE";
+          break;
+        default:
```
grantland: Here i'd assume we'd either call `super.toString()` or throw `IllegalStateException` since it cannot happen.

wangmengyan95: Discuss offline, will throw an exception here.

Parse/src/main/java/com/parse/ParseHttpRequest.java
wangmengyan95: Discuss offline, will throw an exception here.

Parse/src/main/java/com/parse/ParseHttpRequest.java
```diff
@@ -10,12 +10,59 @@
...
 import java.util.HashMap;
-import java.util.Iterator;
 import java.util.Map;
 
 /** package */ class ParseHttpRequest {
+
+  public enum Method {
+    GET, POST, PUT, DELETE;
+
+    public static Method fromString(String string) {
```
grantland: weren't you going to hide this? `/* package */`

wangmengyan95: yep, but I think maybe some developers may need this so I make it public.

Parse/src/main/java/com/parse/ParseHttpRequest.java
```diff
@@ -10,12 +10,59 @@
...
+          method = POST;
+          break;
+        case "PUT":
+          method = PUT;
+          break;
+        case "DELETE":
+          method = DELETE;
+          break;
+        default:
+          throw new IllegalArgumentException("Invalid http method string " + string);
```
grantland: nit: `Invalid http method: <string>`

Parse/src/main/java/com/parse/ParseHttpRequest.java
```diff
@@ -10,12 +10,59 @@
...
+          string = "POST";
+          break;
+        case PUT:
+          string = "PUT";
+          break;
+        case DELETE:
+          string = "DELETE";
+          break;
+        default:
+          throw new IllegalArgumentException("Invalid http method enum " + this);
```
grantland: nit: `Invalid http method: <this>`

grantland:
Small nits, but LGTM. Feel free to squash and merge when done.

PR #94, opened 2015-09-01
Add OS Version to ParseInstallation object

JcMinarro:
Add OS Version to ParseIntallation object

### DIFF
--- a/Parse/src/main/java/com/parse/ParseInstallation.java
+++ b/Parse/src/main/java/com/parse/ParseInstallation.java
@@ -11,6 +11,7 @@ package com.parse;
 import android.content.Context;
 import android.content.pm.PackageInfo;
 import android.content.pm.PackageManager;
+import android.os.Build;
 import android.text.TextUtils;
 
 import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


JcMinarro:
OK,

I have singed your Contributor License Agreement.

grantland:
Is there any specific reason you'd want to add this? 

We try to limit the default properties to those that are can be consistent across all the platforms that we support and are valuable to all developers or are required to support features offered by Parse. In addition, adding new default properties requires checking to make sure that they are not already in use by any developers or else it'd be a breaking change for them and they'd lose their data.

You are also welcome to add this property in your own codebase by using `object.put(key, value)`/`object.get(key)` without needing to merge this upstream to the SDK.

JcMinarro:
It is a really useful data about the device of the user. Some time developers need to know which OS version their users are using to send different kind of pushes, For Example. That property is on other platform too, iOS for example.

hallucinogen:
Unfortunately:
1. There're at least 20 apps using `osVersion` in `ParseInstallation` and we are not sure how are they currently populating the value.
2. This will be pretty hard to maintain on .NET SDK. OS Version will not work on WinRT and getting OS Version on Unity can only be reliably done on Unity 5.1+. Unless we're OK with this field to occasionally returns `null` on some platforms....

JcMinarro:
:+1:  I will put it on my app then.

PR #99, opened 2015-09-02
Clean ParseHttpRequset and ParseHttpResposne

wangmengyan95:
Clean `ParseHttpRequest` and `ParseHttpResponse` and add comment
For `Builder`
1. All constructor are public 
2. All fields are private
3. Add `addHeader()`, `addHeaders()` and `setHeaders()` helper methods
For `ParseHttpRequest` and `ParseHttpResponse`
1. All constructor are private
2. All fields are final and private
3. Add `newBuilder()` helper method
4. Header map is unmodifiable

### DIFF
--- a/Parse/src/main/java/com/parse/ParseHttpRequest.java
+++ b/Parse/src/main/java/com/parse/ParseHttpRequest.java
@@ -12,8 +12,15 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 
+/**
+ * The http request we send to parse server. Instances of this class are not immutable. The
+ * request body may be consumed only once. The other fields are immutable.
+ */
 /** package */ class ParseHttpRequest {
 
+  /**
+   * The {@link ParseHttpRequest} method type.
+   */
   public enum Method {
     GET, POST, PUT, DELETE;
 
@@ -61,43 +68,14 @@ import java.util.Map;
     }
   }
 
-  private final String url;
-  private final ParseHttpRequest.Method method;
-  private final Map<String, String> headers;
-  private final ParseHttpBody body;
-
-  protected ParseHttpRequest(Builder builder) {
-    this.url = builder.url;
-    this.method = builder.method;
-    this.headers = Collections.unmodifiableMap(new HashMap<>(builder.headers));
-    this.body = builder.body;
-  }
-
-  public String getUrl() {
-    return url;
-  }
-
-  public ParseHttpRequest.Method getMethod() {
-    return method;
-  }
-
-  public Map<String, String> getAllHeaders() {
-    return headers;
-  }
-
-  public String getHeader(String name) {
-    return headers.get(name);
-  }
-
-  public ParseHttpBody getBody() {
-    return body;
-  }
-
+  /**
+   * Builder of {@link ParseHttpRequest}.
+   */
   public static class Builder {
-    protected String url;
-    protected ParseHttpRequest.Method method;
-    protected Map<String, String> headers;
-    protected ParseHttpBody body;
+    private String url;
+    private Method method;
+    private Map<String, String> headers;
+    private ParseHttpBody body;
 
     public Builder() {
       this.headers = new HashMap<>();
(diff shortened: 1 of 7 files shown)


Parse/src/main/java/com/parse/ParseHttpRequest.java
```diff
@@ -12,8 +12,15 @@
...
 import java.util.Map;
 
+/**
+ * The http request we send to parse server. Instances of this class are not immutable. The
+ * request body may be consumed only once. The other fields are immutable.
+ */
 /** package */ class ParseHttpRequest {
 
+  /**
+   * The {@link ParseHttpRequest} method type.
```
grantland: If the reference is the class you're in, we generally use `{@code blah}`

Parse/src/main/java/com/parse/ParseHttpRequest.java
```diff
@@ -139,4 +122,40 @@ public ParseHttpRequest build() {
...
+  private ParseHttpRequest(Builder builder) {
+    this.url = builder.url;
+    this.method = builder.method;
+    this.headers = Collections.unmodifiableMap(new HashMap<>(builder.headers));
+    this.body = builder.body;
+  }
+
+  public Builder newBuilder() {
+    return new Builder(this);
+  }
```
grantland: Any reason for this method? Seems redundant with `new Builder(request)`

Parse/src/main/java/com/parse/ParseHttpRequest.java
```diff
@@ -12,8 +12,15 @@
 import java.util.HashMap;
 import java.util.Map;
 
+/**
+ * The http request we send to parse server. Instances of this class are not immutable. The
+ * request body may be consumed only once. The other fields are immutable.
+ */
```
grantland: How is this class mutable?

Parse/src/main/java/com/parse/ParseHttpResponse.java
```diff
@@ -14,11 +14,14 @@
 import java.util.Map;
 
 /**
- * The base class of a http response we receive from parse server. It can be implemented by
- * different http library such as Apache http, Android URLConnection, Square OKHttp and so on.
+ * The http response we receive from parse server. Instances of this class are not immutable. The
+ * response body may be consumed only once. The other fields are immutable.
```
grantland: how is this class mutable?

Parse/src/main/java/com/parse/ParseHttpResponse.java
```diff
@@ -116,6 +130,14 @@ public ParseHttpResponse build() {
...
+   * Generates a new {@link com.parse.ParseHttpResponse.Builder} based on an
+   * {@link ParseHttpResponse}, the {@link com.parse.ParseHttpResponse.Builder}'s values are come
+   * from the {@link ParseHttpResponse}.
+   *
+   * @return A new {@link com.parse.ParseHttpResponse.Builder} whose values are come from the
+   * {@link ParseHttpResponse}.
+   */
   public Builder newBuilder() {
     return new Builder(this);
   }
```
grantland: Same RE `new Builder(response)`

Parse/src/main/java/com/parse/ParseHttpRequest.java
wangmengyan95: kk

Parse/src/main/java/com/parse/ParseHttpRequest.java
wangmengyan95: er.. I check OKHttp, they provide this helper method so I add it.

Parse/src/main/java/com/parse/ParseHttpRequest.java
wangmengyan95: The body of the `ParseHttpRequest` is mutable, they can be change after we build the `ParseHttpRequest`.

Parse/src/main/java/com/parse/ParseHttpResponse.java
wangmengyan95: Same with `ParseHttpRequest`, body is mutable.

Parse/src/main/java/com/parse/ParseHttpRequest.java
grantland: we can always add this later, so let's remove it for now

Parse/src/main/java/com/parse/ParseHttpRequest.java
grantland: :+1:

Parse/src/main/java/com/parse/ParseHttpResponse.java
grantland: :+1:

PR #106, opened 2015-09-04
Split ParseNetworkInterceptor Module

wangmengyan95:
Still can not solve the wired robolectric Manifest bug, right now since this test does not heavily rely on the android framework methods, so I manually import json library and remove robolectric dependency. I will try to make a sample project and file the bug to robolectric. Even we can avoid robolectric this time, when we want to modularization in the future, we have to find a solution. @grantland


grantland:
It looks like we're missing documentation for various methods in `ParseLogInterceptor.Logger`

PR #107, opened 2015-09-04
Use estimatedData for data availability

grantland:
No need for another `Map` with autoboxed `Boolean` values.


wangmengyan95:
LGTM

PR #113, opened 2015-09-08
Add new ParseFile download api

wangmengyan95:
Support fetch a ParseFile as File pointer or InputStream.
File related new apis:
1. `public File getFile() throws ParseException`
2. `public Task<File> getFileInBackground(final ProgressCallback progressCallback)`
3. `public Task<File> getFileInBackground()`
4. `public void getFileInBackground(GetFileCallback fileCallback, final ProgressCallback progressCallback)`
5. `public void getFileInBackground(GetFileCallback fileCallback)`

DataStream related new apis:
1. `public File getDataStream() throws ParseException`
2. `public Task<File> getDataStreamInBackground(final ProgressCallback progressCallback)`
3. `public Task<File> getDataStreamInBackground()`
4. `public void getDataStreamInBackground(GetDataStreamCallback fileCallback, final ProgressCallback progressCallback)`
5. `public void getDataStreamInBackground(GetDataStreamCallback fileCallback)`

### DIFF
--- a/Parse/src/main/java/com/parse/GetFileCallback.java
+++ b/Parse/src/main/java/com/parse/GetFileCallback.java
@@ -8,33 +8,34 @@
  */
 package com.parse;
 
+import java.io.File;
+
 /**
- * A {@code GetDataCallback} is used to run code after a {@link ParseFile} fetches its data on a
- * background thread.
+ * A {@code GetFileCallback} is used to run code after a {@link ParseFile} fetches its data on
+ * a background thread.
  * <p/>
- * The easiest way to use a {@code GetDataCallback} is through an anonymous inner class. Override
- * the {@code done} function to specify what the callback should do after the fetch is complete.
- * The {@code done} function will be run in the UI thread, while the fetch happens in a
+ * The easiest way to use a {@code GetFileCallback} is through an anonymous inner class.
+ * Override the {@code done} function to specify what the callback should do after the fetch is 
+ * complete. The {@code done} function will be run in the UI thread, while the fetch happens in a
  * background thread. This ensures that the UI does not freeze while the fetch happens.
  * <p/>
  * <pre>
- * file.getDataInBackground(new GetDataCallback() {
- *   public void done(byte[] data, ParseException e) {
+ * file.getFileInBackground(new GetFileCallback() {
+ *   public void done(File file, ParseException e) {
  *     // ...
  *   }
  * });
  * </pre>
  */
-public interface GetDataCallback extends ParseCallback2<byte[], ParseException> {
+public interface GetFileCallback extends ParseCallback2<File, ParseException> {
   /**
    * Override this function with the code you want to run after the fetch is complete.
    *
-   * @param data
+   * @param file
    *          The data that was retrieved, or {@code null} if it did not succeed.
    * @param e
    *          The exception raised by the fetch, or {@code null} if it succeeded.
    */
   @Override
-  public void done(byte[] data, ParseException e);
+  public void done(File file, ParseException e);
 }
-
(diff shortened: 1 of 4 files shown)


Parse/src/main/java/com/parse/GetFileCallback.java
```diff
@@ -0,0 +1,41 @@
...
+   * Override this function with the code you want to run after the fetch is complete.
+   *
+   * @param file
+   *          The data that was retrieved, or {@code null} if it did not succeed.
+   * @param e
+   *          The exception raised by the fetch, or {@code null} if it succeeded.
+   */
+  @Override
+  public void done(File file, ParseException e);
+}
```
grantland: nit: newline

Parse/src/main/java/com/parse/ParseFileUtils.java
```diff
@@ -519,6 +519,11 @@ public static String readFileToString(File file, String encoding) throws IOExcep
     return readFileToString(file, Charset.forName(encoding));
   }
 
+  public static void writeStringToFile(File file, String string)
+      throws IOException {
+    writeByteArrayToFile(file, string.getBytes());
+  }
+
```
grantland: I don't think we should add this since it was `@Deprecated` in `org.apache.commons.io.FileUtils`

grantland:
Some nits and a thought Q

Do you think the `getDataStream` APIs are worthwhile when we have `getFile` APIs?

I think we should update the `get(Data|File|DataStream)` APIs to match iOS's since it's a bit more descriptive:

#### getData:
> Synchronously gets the data from cache if available or fetches its contents from the network.

#### getDataStream:
>  This method is like getData but avoids ever holding the entire PFFile contents in memory at once.

wangmengyan95:
For `getDataStream`, this method is not like iOS which does not wait download to cache file finish but open a stream from the cache file directly. In Android, it just wait download finish and read the data from cache file for user. So I do not hold strong opinions for either sides.

Parse/src/main/java/com/parse/ParseFile.java
```diff
@@ -361,78 +363,129 @@ public void saveInBackground(SaveCallback callback) {
...
+   *          A {@code GetDataCallback} that is called when the get completes.
+   */
+  public void getDataInBackground(GetDataCallback dataCallback) {
+    ParseTaskUtils.callbackOnMainThreadAsync(getDataInBackground(), dataCallback);
+  }
+
+  /**
+   * Synchronously gets the file pointer from cache if available or fetches its content from the
+   * network. You probably want to use {@link #getFileInBackground()} instead unless you're already
+   * in a background thread.
```
grantland: We should add the following:

```
<strong>Note: </strong> The {@link File} location may change without notice and should not be stored to be accessed later.
```

/cc @nlutsenko @richardjrossiii

Parse/src/main/java/com/parse/ParseFile.java
```diff
@@ -361,78 +363,129 @@ public void saveInBackground(SaveCallback callback) {
...
+   */
+  public File getFile() throws ParseException {
+    return ParseTaskUtils.wait(getFileInBackground());
+  }
+
+  /**
+   * Asynchronously gets the file pointer from cache if available or fetches its content from the
+   * network. The {@code ProgressCallback} will be called periodically with progress updates.
```
grantland: Same

Parse/src/main/java/com/parse/ParseFile.java
nlutsenko: :+1:

wangmengyan95: :+1:

grantland:
LGTM, feel free to squash and merge!

PR #115, opened 2015-09-09
Add new ParseFile constructors and update save method

wangmengyan95:
Support creating a ParseFile with File.
New apis:
1. `public ParseFile(String name, File file, String contentType)`
2. `public ParseFile(File file)`
3. `public ParseFile(String name, File file)`
4. `public ParseFile(File file, String contentType)`

Update `saveAsync()` to support upload with a File pointer.

### DIFF
--- a/Parse/src/main/java/com/parse/ParseFile.java
+++ b/Parse/src/main/java/com/parse/ParseFile.java
@@ -134,10 +134,39 @@ public class ParseFile {
    * successfully synced with the server.
    */
   /* package for tests */ byte[] data;
+  /* package for tests */ File file;
 
   /* package for tests */ final TaskQueue taskQueue = new TaskQueue();
   private Set<Task<?>.TaskCompletionSource> currentTasks = Collections.synchronizedSet(
       new HashSet<Task<?>.TaskCompletionSource>());
+
+  /**
+   * Creates a new file from a file pointer.
+   *
+   * @param file
+   *          The file.
+   */
+  public ParseFile(File file) {
+    this(file, null);
+  }
+
+  /**
+   * Creates a new file from a file pointer, and content type. Content type will be used instead of
+   * auto-detection by file extension.
+   *
+   * @param file
+   *          The file.
+   * @param contentType
+   *          The file's content type.
+   */
+  public ParseFile(File file, String contentType) {
+    this(new State.Builder().name(file.getName()).mimeType(contentType).build());
+    if (file.length() > MAX_FILE_SIZE) {
+      throw new IllegalArgumentException(String.format("ParseFile must be less than %d bytes",
+          MAX_FILE_SIZE));
+    }
+    this.file = file;
+  }
 
   /**
    * Creates a new file from a byte array, file name, and content type. Content type will be used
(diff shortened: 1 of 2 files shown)


Parse/src/main/java/com/parse/ParseFile.java
```diff
@@ -134,12 +134,74 @@ public String url() {
...
+   * @param name
+   *          The file's name, ideally with extension. The file name must begin with an alphanumeric
+   *          character, and consist of alphanumeric characters, periods, spaces, underscores, or
+   *          dashes.
+   * @param file
+   *          The file.
+   * @param contentType
+   *          The file's content type.
+   */
+  public ParseFile(String name, File file, String contentType) {
```
grantland: Is there any real reason we need to pass in `name` since we can get it from `File#getName()`?

/cc @nlutsenko since it looks like we have this constructor on iOS.

Parse/src/main/java/com/parse/ParseFile.java
```diff
@@ -134,12 +134,74 @@ public String url() {
...
+    this(new State.Builder().name(name).mimeType(contentType).build());
+    if (file.length() > MAX_FILE_SIZE) {
+      throw new IllegalArgumentException(String.format("ParseFile must be less than %d bytes",
+          MAX_FILE_SIZE));
+    }
```
grantland: I thought we don't need this limitation with these new APIs?

Parse/src/main/java/com/parse/ParseFile.java
```diff
@@ -134,12 +134,74 @@ public String url() {
...
+  }
+
+  /**
+   * Creates a new file from a file pointer.
+   *
+   * @param file
+   *          The file.
+   */
+  public ParseFile(File file) {
+    this(null, file, null);
```
grantland: why not use `File#getName()` for the name?

Parse/src/main/java/com/parse/ParseFile.java
```diff
@@ -134,12 +134,74 @@ public String url() {
...
+   * Creates a new file from a file pointer, and content type. Content type will be used instead of
+   * auto-detection by file extension.
+   *
+   * @param file
+   *          The file.
+   * @param contentType
+   *          The file's content type.
+   */
+  public ParseFile(File file, String contentType) {
+    this(null, file, contentType);
```
grantland: same RE `File#getName()`

Parse/src/main/java/com/parse/ParseFile.java
```diff
@@ -273,12 +335,21 @@ public void save() throws ParseException {
...
+                data,
+                sessionToken,
+                progressCallbackOnMainThread(uploadProgressCallback),
+                cancellationToken) :
+            getFileController().saveAsync(
+                state,
+                file,
+                sessionToken,
+                progressCallbackOnMainThread(uploadProgressCallback),
+                cancellationToken);
```
grantland: nit: Can we just use a `if`/`else` here instead? This ternary operator is weird to look at

Parse/src/main/java/com/parse/ParseFile.java
```diff
@@ -298,6 +369,10 @@ public void save() throws ParseException {
    * @return A Task that will be resolved when the save completes.
    */
   public Task<Void> saveInBackground(final ProgressCallback uploadProgressCallback) {
+    if (data != null && file != null) {
+      throw new IllegalArgumentException("File and data can not be set at the same time");
+    }
+
```
grantland: Is this actually possible? Under what circumstances?

grantland:
Passing to @nlutsenko for API discussion

Parse/src/main/java/com/parse/ParseFile.java
nlutsenko: We don't have this constructor on iOS.
Simply `fileWithContentsAtPath:`, since you are right - we can get the file name by using the last path component.

grantland: :+1:

grantland:
Let's keep the constructors to just the following:
* `new File(file, contentType)`
* `new File(file)`

We can determine `name` from `File#getName()`, and if you really want a different file name you can move the file first.

Parse/src/main/java/com/parse/ParseFile.java
grantland: We also don't need the `File` pointer anymore...

Parse/src/main/java/com/parse/ParseFile.java
wangmengyan95: kk

Parse/src/main/java/com/parse/ParseFile.java
wangmengyan95: k'k

Parse/src/main/java/com/parse/ParseFile.java
wangmengyan95: It is impossible. I remembered you recommended me to add this checking in `ParseRESTFileCommand` constructor so I also add it here. Do we still need this?

wangmengyan95:
Will revise the constructor.

Parse/src/main/java/com/parse/ParseFile.java
grantland: delete?

grantland:
LGTM, one nit

PR #119, opened 2015-09-10
Create cache dir if it does not exist before we use tempFile

wangmengyan95:
Fix a bug in ParseAWSRequest, if users never download/upload any `ParseFile`, the cache dir does not exist. In this situation, if users download a `ParseFile`, in `ParseAWSRequset`, `new FileOutputStream(tempFile)` will throw an exception since parent dir does not exist. By using `ParseFileUtils`, we create parent dir if it does not exist which fixes this problem.


grantland:
Wouldn't it be better to just update the previous tests to use a non-existant folder?

grantland:
LGTM

grantland:
Looks like there are merge conflicts, so you'll need to rebase and squash before merging.

PR #120, opened 2015-09-10
Manually delete cacheFile before we write to it

wangmengyan95:


grantland:
What use cases does this fix? It seems like if we're not able to write to the file, we wouldn't be able to delete it either.

wangmengyan95:
If the cacheFile exist, when we call `ParseFileUtils.moveFile(tempFile, cacheFile)`, an IOException will be thrown. We delete it manually to make sure it does not exist before we move tempFile.

grantland:
Ah I see. LGTM!

PR #125, opened 2015-09-10
Publicize Third Party Authentication methods

grantland:
Publicize + JavaDoc. Required for open sourcing ParseFacebookUtils and ParseTwitterUtils.

### DIFF
--- a/Parse/src/main/java/com/parse/ParseAuthenticationProvider.java
+++ b/Parse/src/main/java/com/parse/ParseAuthenticationProvider.java
@@ -13,41 +13,34 @@ import java.util.Map;
 import bolts.Task;
 
 /**
- * Provides a general interface for delegation of the authentication process.
+ * Provides a general interface for delegation of third party authentication.
  */
-/** package */ interface ParseAuthenticationProvider {
+public interface ParseAuthenticationProvider {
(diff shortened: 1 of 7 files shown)


Parse/src/main/java/com/parse/ParseAuthenticationProvider.java
```diff
@@ -13,36 +13,29 @@
...
-   */
-  Task<Map<String, String>> authenticateAsync();
-
-  /**
-   * Deauthenticates (logs out) the user associated with this provider. This call may block.
-   *
-   * @return A {@link Task} that resolves when deauthentication is complete.
+   * @return A {@link Task} that resolves when deauthentication completes.
    */
   Task<Void> deauthenticateAsync();
```
hallucinogen: This should be `deauthenticateInBackground`?

grantland: good catch!

grantland:
@hallucinogen updated

hallucinogen:
Looking good! @nlutsenko should double-check tho since this is a public API.

nlutsenko:
LGTM

PR #126, opened 2015-09-11
Public Interceptor related classes

wangmengyan95:


grantland:
Don't we need to publicize `ParseNetworkInterceptor.Chain` as well?

Should we make `ParseHttpRequest`, `ParseHttpResponse`, their `Builder`s, and `Method`  `final` or is there a reason they'd be subclassed?

wangmengyan95:
`ParseNetworkInterceptor.Chain` is public since `ParseNetworkInterceptor` is public. The compiler recommends to remove `public`.

wangmengyan95:
`Method ` is enum, can not be final.

grantland:
Still missing JavaDoc on getters/setters as well as interface methods.

We might as well move these public APIs to `com.parse.http` since they're all public and don't need private access to anything.

Parse/src/main/java/com/parse/http/ParseHttpBody.java
```diff
@@ -39,6 +40,14 @@
    */
   public abstract void writeTo(OutputStream out) throws IOException;
 
+  /**
+   * Creates an {@code ParseHttpBody} with given contentType and contentLength.
```
grantland: `{@code }` contentType and contentLength

Parse/src/main/java/com/parse/http/ParseHttpResponse.java
```diff
@@ -0,0 +1,248 @@
...
+   * be read once and can't be reset.
+   *
+   * @return The content of the {@code ParseHttpResponse}'s body.
+   */
+  public InputStream getContent() {
+    return content;
+  }
+
+  /**
+   * Returns the size of the {@code ParseHttpResponse}'s body. -1 if the size of the
```
grantland: `{@code -1}`

Parse/src/main/java/com/parse/http/ParseHttpRequest.java
```diff
@@ -16,14 +16,22 @@
...
 
   /**
    * The {@code ParseHttpRequest} method type.
    */
   public enum Method {
+
     GET, POST, PUT, DELETE;
 
+    /**
+     * Creates a {@code Method} from the given string. Valid stings are GET, POST, PUTa ahd DELETE.
```
grantland: s/stings/strings/g
s/PUTa ahd/PUT, and/g

`{@code }` GET, POST, PUT, DELETE

Parse/src/main/java/com/parse/http/ParseHttpRequest.java
```diff
@@ -88,36 +110,85 @@ public Builder(ParseHttpRequest request) {
...
+     *          The headers that need to be added.
+     * @return This {@code Builder}.
+     */
     public Builder addHeaders(Map<String, String> headers) {
       this.headers.putAll(headers);
       return this;
     }
 
+    /**
+     * Sets headers of this {@code Builder}. All existing headers will be deleted.
```
grantland: s/deleted/cleared/g

grantland:
Some small nits, but LGTM:

* `{@code <blah>}` things like `-1`, `Content-Type` (`ParseHttpBody`)

PR #133, opened 2015-09-14
Change ParseHttpBody fields access

wangmengyan95:
1. Change ParseHttpBody fields to private
2. Change ParseByteHttpBody fields to package protected


grantland:
LGTM

PR #136, opened 2015-09-15
Clean up Push warnings

grantland:


Parse/src/main/java/com/parse/PushService.java
```diff
@@ -77,12 +76,6 @@
  * <p/>
  * <pre>
  * &lt;service android:name="com.parse.PushService" /&gt;
- * &lt;receiver android:name="com.parse.ParseBroadcastReceiver"&gt;
```
hallucinogen: also remove `{@link ParseBroadcastReceiver}` from line 74.

hallucinogen:
Nits, otherwise LGTM.

PR #141, opened 2015-09-17
Fix NPE in ParseKeyValueCache due to cache dir not exist

wangmengyan95:
1. Fix `ParseKeyValueCache.saveToKeyValueCache` `cacheDir.listFiles().length` NPE.
2. Refactor `ParseKeyValueCache.saveToKeyValueCache`
3. Add unit tests

### DIFF
--- a/Parse/src/test/java/com/parse/ParseKeyValueCacheTest.java
+++ b/Parse/src/test/java/com/parse/ParseKeyValueCacheTest.java
@@ -14,20 +14,29 @@ import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
 
+import java.io.File;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.Callable;
 
 import bolts.Task;
 
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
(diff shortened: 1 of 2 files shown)


wangmengyan95:
The additional checking I add `files == null`(assume mkdir fails) is hard to reach in unit test, so probably that is why codecov test not pass.

Parse/src/test/java/com/parse/ParseKeyValueCacheTest.java
```diff
@@ -59,4 +65,33 @@ public Void call() throws Exception {
     }
     ParseTaskUtils.wait(Task.whenAll(tasks));
   }
+
+  @Test
+  public void testSaveToKeyValueCacheWithoutCacheDir() throws Exception {
+    // Delete the cache folder(Simulate users clear the app cache)
+    File keyValueCacheDir = new File(temporaryFolder.getRoot(), "ParseKeyValueCache");
```
grantland: nit: It would be better to hold a reference to the `File` we use in `initialize()` in `setUp` and use that here.

grantland:
LGTM, small nit

Also, it looks like `ParseFileController` might also be affected by a similar problem.

wangmengyan95:
BTW, @grantland `testMultipleAsynchronousWrites` seems really wried, there is no assert in the test case, is it intended?

Issue #162, opened 2015-10-01
Can't set the notification color in Lollipop


mars3142:
It is not possible to set the notification color in Android Lollipop and above.

mars3142:
I'll create a pull request.

grantland:
This can be accomplished by extending `ParsePushBroadcastReceiver` and overriding `#getNotification()`.

mars3142:
But the override for the getNotification is too much boilerplate code. This is not a possible solution for everyone. Why can't you include my pull request, because it's much easier with my solution!?

grantland:
We try to balance adding useful helpers with keeping a clean and simple API. Making it easier for developers to solve complex problems is always great, but only if it solves common problems and the savings outweigh the added complexity of the additional APIs.

There are many ways to customize notifications, but the required minimum are `smallIcon`, `title`, and `text`. We supply `title` and `text` from the push payload and `smallIcon` is defined by overriding `getSmallIcon()` or defining meta-data in `AndroidManifest.xml`.

Outside of the bare minimum, there are various other attributes such as sound, lights, color, action, vibrate, sort key, etc. While it's possible to include various helper methods and meta-data attributes to allow customization, it doesn't make development significantly easier compared to the amount of extra API duplication that already exists in AOSP.

At the point when you're customizing something beyond the required minimum it isn't really boilerplate code, but app specific code.

PR #169, opened 2015-10-02
setColor for Notifications on Lollipop+

mars3142:
It's the pull request for the issue #162


mars3142:
The two changed classes hadn't any tests previously. Do you need tests for this?

grantland:
Thanks for the PR, however, this seems to be something that can easily be accomplished without modifying the SDK by extending `ParsePushBroadcastReceiver#getNotification()` on a per-app basis which is recommended in our guide.

robertdolca:
This feature would be very useful. Overwriting the whole #getNotification() only for changing the background color seems like a bad practice.

mars3142:
Thanks, @robertdolca - This was my intention.

Issue #37, opened 2015-08-20
StrictMode: Initialise off the Main Thread.


chrisjenx:
When calling the `initialize` method, I would hope that any IO would be done off the UI thread.
Seen as the library is build on bolts, It would be safe assumption to make sure that any touching of resources should be handled by a task.

It would also be prudent to queue up subsequent tasks until the `initialize` is fully completed.

``` bash
5660-5660/com.owlr.tv.debug D/StrictMode﹕ StrictMode policy violation; ~duration=31 ms: android.os.StrictMode$StrictModeDiskReadViolation: policy=2591 violation=2
            at android.os.StrictMode$AndroidBlockGuardPolicy.onReadFromDisk(StrictMode.java:1137)
            at libcore.io.BlockGuardOs.access(BlockGuardOs.java:67)
            at java.io.File.doAccess(File.java:283)
            at java.io.File.exists(File.java:363)
            at android.app.ContextImpl.getDir(ContextImpl.java:2229)
            at android.content.ContextWrapper.getDir(ContextWrapper.java:257)
            at android.net.SSLSessionCache.<init>(SSLSessionCache.java:97)
            at com.parse.ParseRequest.initialize(ParseRequest.java:137)
            at com.parse.Parse.initialize(Parse.java:184)
            at com.owlr.io.ParseAPIWrapper.init(ParseAPIWrapper.java:81)
```

The prior strict mode violation occurs when calling:

``` java
Parse.enableLocalDatastore(application);
Parse.initialize(application, metaDataManager.getParseAppId(),
        metaDataManager.getParseClientKey());
```

Either that or its made a bit clearer that IO operations happen on `initialize`.

YaoPersonal:
If we merge #180, we no longer have read violation causing from RESTclient. But we still have other read and write violation in other place.

For read violation:
>android.content.ContextWrapper.getCacheDir(ContextWrapper.java:232)
>com.parse.ParseKeyValueCache.initialize(ParseKeyValueCache.java:56)
>com.parse.Parse.initialize(Parse.java:207)

For write violation:
>java.io.OutputStream.write(OutputStream.java:82)
>com.parse.Parse.checkCacheApplicationId(Parse.java:388)
>com.parse.Parse.initialize(Parse.java:211)

alexblack:
+1

Tom-Fox1:
Closing this as no one has reported this issue recently, and we don't have the resources to investigate these historic reports.

[...]

PR #180, opened 2015-10-07
[android] Remove ParseRequest defaultClient

YaoPersonal:
We do not need defaultClient as a global state in class ParseRequest anymore.

### DIFF
--- a/Parse/src/main/java/com/parse/Parse.java
+++ b/Parse/src/main/java/com/parse/Parse.java
@@ -414,14 +414,14 @@ public class Parse {
           || (!isLocalDatastoreEnabled && eventuallyQueue instanceof ParsePinningEventuallyQueue)) {
         checkContext();
         eventuallyQueue = isLocalDatastoreEnabled
-          ? new ParsePinningEventuallyQueue(context)
-          : new ParseCommandCache(context);
+          ? new ParsePinningEventuallyQueue(context, ParsePlugins.get().restClient())
+          : new ParseCommandCache(context, ParsePlugins.get().restClient());
 
         // We still need to clear out the old command cache even if we're using Pinning in case
         // anything is left over when the user upgraded. Checking number of pending and then
         // initializing should be enough.
         if (isLocalDatastoreEnabled && ParseCommandCache.getPendingCount() > 0) {
-          new ParseCommandCache(context);
+          new ParseCommandCache(context, ParsePlugins.get().restClient());
         }
       }
       return eventuallyQueue;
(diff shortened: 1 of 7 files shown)


Parse/src/main/java/com/parse/ParseRESTCommand.java
```diff
@@ -249,6 +249,10 @@ protected ParseHttpBody newBody(ProgressCallback uploadProgressCallback) {
     }
   }
 
+  public Task<JSONObject> executeAsync() {
+    return executeAsync(ParsePlugins.get().restClient());
+  }
+
```
grantland: This defeats the purpose of injection as it makes it difficult to unit test execution paths that use this method.

In reality, we should be passing the client through to anything that was using `ParseRequest#executeAsync()` and `ParseRequest#executeAsync(ProgressCallback, ProgressCallback, CancellationToken)`.

Parse/src/main/java/com/parse/Parse.java
```diff
@@ -415,14 +414,14 @@ static void checkCacheApplicationId() {
...
-          : new ParseCommandCache(context);
+          ? new ParsePinningEventuallyQueue(context, ParsePlugins.get().restClient())
+          : new ParseCommandCache(context, ParsePlugins.get().restClient());
 
         // We still need to clear out the old command cache even if we're using Pinning in case
         // anything is left over when the user upgraded. Checking number of pending and then
         // initializing should be enough.
         if (isLocalDatastoreEnabled && ParseCommandCache.getPendingCount() > 0) {
-          new ParseCommandCache(context);
+          new ParseCommandCache(context, ParsePlugins.get().restClient());
```
grantland: nit: let's grab `restClient` once and reuse the same variable

Parse/src/main/java/com/parse/ParseEventuallyQueue.java
```diff
@@ -30,6 +30,16 @@
    */
   private TestHelper testHelper;
 
+  private ParseHttpClient httpClient;
+
+  protected void setHttpClient(ParseHttpClient client) {
+    httpClient = client;
+  }
```
grantland: is there any reason why we're using a getter and setter? why not use a constructor so we can make `httpClient` final?

Parse/src/main/java/com/parse/ParseObject.java
```diff
@@ -1660,11 +1660,13 @@ public String then(Task<Void> task) throws Exception {
 
   // Currently only used by ParsePinningEventuallyQueue for saveEventually due to the limitation in
   // ParseCommandCache that it can only return JSONObject result.
-  /* package */ Task<JSONObject> saveAsync(final ParseOperationSet operationSet, String sessionToken)
-      throws ParseException {
+  /* package */ Task<JSONObject> saveAsync(
+      final ParseOperationSet operationSet,
+      String sessionToken,
+      ParseHttpClient client) throws ParseException {
```
grantland: nit: let's make `client` the first param

grantland:
Could you also see if this resolves #37?

grantland:
LGTM, mind squashing?

grantland:
Squashed in #214

PR #195, opened 2015-10-16
Remove gradle.properties

wangmengyan95:
Remove `gradle.properties` and move version to `build.gradle`.


Parse/build.gradle
```diff
@@ -24,7 +24,7 @@ android {
         minSdkVersion rootProject.ext.minSdkVersion
         targetSdkVersion rootProject.ext.targetSdkVersion
         versionCode 1
-        versionName VERSION_NAME
+        versionName version
```
grantland: does this work or do we need to use `project.version`?

wangmengyan95: It works, but let's use project.version.

grantland:
LGTM!

PR #214, opened 2015-10-22
Remove ParseRequest defaultClient

grantland:
We do not need defaultClient as a global state in class ParseRequest anymore.


wangmengyan95:
LGTM!

PR #215, opened 2015-10-22
Remove GET_ACCOUNTS requirement in Parse

### LINKED ISSUE #129: Why do Parse Push Notification still need "Get_Account" permission to work?
GCM is no longer require "android.permission.GET_ACCOUNTS" to work, but why parse still need that permission.

grantland:
GET_ACCOUNTS is only required for GCM on devices running lower than 4.0.4.
This will allow developers to leave out this permission if they are targetting
4.0.4+.

This also removes some trailing whitespace.

Resolves #129

### DIFF
--- a/Parse/src/main/java/com/parse/ManifestInfo.java
+++ b/Parse/src/main/java/com/parse/ManifestInfo.java
@@ -473,7 +473,6 @@ import java.util.List;
       "android.permission.INTERNET",
       "android.permission.ACCESS_NETWORK_STATE",
       "android.permission.WAKE_LOCK",
-      "android.permission.GET_ACCOUNTS",
       "com.google.android.c2dm.permission.RECEIVE",
       context.getPackageName() + ".permission.C2D_MESSAGE"
     };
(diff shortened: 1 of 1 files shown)


Parse/src/main/java/com/parse/ManifestInfo.java
```diff
@@ -549,27 +548,27 @@ private static String getGcmManifestMessage() {
...
-        "<uses-permission android:name=\"android.permission.ACCESS_NETWORK_STATE\" />\n" + 
-        "<uses-permission android:name=\"android.permission.VIBRATE\" />\n" + 
-        "<uses-permission android:name=\"android.permission.WAKE_LOCK\" />\n" + 
-        "<uses-permission android:name=\"android.permission.GET_ACCOUNTS\" />\n" + 
-        "<uses-permission android:name=\"com.google.android.c2dm.permission.RECEIVE\" />\n" + 
+        "<uses-permission android:name=\"android.permission.INTERNET\" />\n" +
+        "<uses-permission android:name=\"android.permission.ACCESS_NETWORK_STATE\" />\n" +
+        "<uses-permission android:name=\"android.permission.VIBRATE\" />\n" +
+        "<uses-permission android:name=\"android.permission.WAKE_LOCK\" />\n" +
+        "<uses-permission android:name=\"android.permission.GET_ACCOUNTS\" />\n" +
```
bnham: seems like we should remove this from the comment

bnham:
seems safe to merge. i'll leave the decision over whether you want to include the permission in the comment up to you.

grantland:
I think it's safer to keep the permission there, as it could be a reason that Parse isn't working.

PR #234, opened 2015-11-05
Update proguard-rules.pro for android-23

### LINKED ISSUE #231: Issue with android M compilesdkversion 23
I am getting the following gradle error after updating to latest sdk version 

>Warning:com.parse.ParseApacheHttpClient: can't find referenced method 'org.apache.http.conn.ssl.SSLSocketFactory getHttpSocketFactory(int,android.net.SSLSessionCache)' >in library class android.net.SSLCertificateSocketFactory

I am using "useLibrary" option in the gradle file to include legacy apache
android {
useLibrary 'org.apache.http.legacy'
}

grantland:
android-23 removed some APIs that our SDK uses internally. Luckily,
we only conditionally use these APIs on older devices so this is
a non-issue for our SDK running on android-23.

This change adds instructions to ignore warnings from proguard stating
that these APIs do not exist when using `compileSdkVersion = 23` since
we already have checks in place.

Test Plan:

Change ParseStarterProject's `build.gradle`:

```
android {
    compileSdkVersion 23
    useLibrary `org.apache.http.legacy`

    // ...

    buildTypes {
        release {
            minifyEnabled true

            // ...
        }
    }
}
```

Resolves #231

### DIFF
--- a/ParseStarterProject/proguard-rules.pro
+++ b/ParseStarterProject/proguard-rules.pro
@@ -23,5 +23,7 @@
 # Required for Parse
 -keepattributes *Annotation*
 -keepattributes Signature
+-dontwarn android.net.SSLCertificateSocketFactory
+-dontwarn android.app.Notification
 -dontwarn com.squareup.**
 -dontwarn okio.**


grantland:
`ParseApacheHttpClient` which uses missing `android.net.SSLCertificateSocketFactory` only gets used on < android-19: https://github.com/ParsePlatform/Parse-SDK-Android/blob/master/Parse/src/main/java/com/parse/ParseHttpClient.java#L51

`NotificationCompatImplBase` which uses missing `android.app.Notification.setLatestEventInfo` only gets used on < android-14: https://github.com/ParsePlatform/Parse-SDK-Android/blob/master/Parse/src/main/java/com/parse/NotificationCompat.java#L103

wangmengyan95:
LGTM

PR #245, opened 2015-11-11
Decouple object subclassing.

richardjrossiii:
This fully decouples object subclassing from the ParseObject class, bringing it up to parity with our other native platforms.

### DIFF
--- a/Parse/src/main/java/com/parse/ParseObject.java
+++ b/Parse/src/main/java/com/parse/ParseObject.java
@@ -12,8 +12,6 @@ import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
 
-import java.lang.reflect.Member;
-import java.lang.reflect.Modifier;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -29,7 +27,6 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.Callable;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.locks.Lock;
 
@@ -72,17 +69,16 @@ public class ParseObject {
   // and not check after a while
   private static final String KEY_IS_DELETING_EVENTUALLY_OLD = "isDeletingEventually";
 
-  private static final Map<Class<? extends ParseObject>, String> classNames =
-      new ConcurrentHashMap<>();
-  private static final Map<String, Class<? extends ParseObject>> objectTypes =
-      new ConcurrentHashMap<>();
-
   private static ParseObjectController getObjectController() {
     return ParseCorePlugins.getInstance().getObjectController();
   }
 
   private static LocalIdManager getLocalIdManager() {
     return ParseCorePlugins.getInstance().getLocalIdManager();
+  }
+
+  private static ParseObjectSubclassingController getSubclassingController() {
+    return ParseCorePlugins.getInstance().getSubclassingController();
   }
 
   /** package */ static class State {
@@ -360,21 +356,13 @@ public class ParseObject {
           "You must specify a Parse class name when creating a new ParseObject.");
     }
     if (AUTO_CLASS_NAME.equals(theClassName)) {
-      theClassName = getClassName(this.getClass());
+      theClassName = getSubclassingController().getClassName(getClass());
     }
 
     // If this is supposed to be created by a factory but wasn't, throw an exception.
-    if (this.getClass().equals(ParseObject.class) && objectTypes.containsKey(theClassName)
-        && !objectTypes.get(theClassName).isInstance(this)) {
+    if (!getSubclassingController().isSubclassValid(theClassName, getClass())) {
       throw new IllegalArgumentException(
           "You must create this type of ParseObject using ParseObject.create() or the proper subclass.");
-    }
-
-    // If this is an unregistered subclass, throw an exception.
-    if (!this.getClass().equals(ParseObject.class)
-        && !this.getClass().equals(objectTypes.get(theClassName))) {
-      throw new IllegalArgumentException(
-          "You must register this ParseObject subclass before instantiating it.");
     }
 
     operationSetQueue = new LinkedList<>();
@@ -410,17 +398,7 @@ public class ParseObject {
    * @return A new {@code ParseObject} for the given class name.
    */
   public static ParseObject create(String className) {
-    if (objectTypes.containsKey(className)) {
-      try {
-        return objectTypes.get(className).newInstance();
-      } catch (Exception e) {
-        if (e instanceof RuntimeException) {
-          throw (RuntimeException) e;
-        }
-        throw new RuntimeException("Failed to create instance of subclass.", e);
-      }
-    }
-    return new ParseObject(className);
+    return getSubclassingController().newInstance(className);
   }
 
   /**
@@ -434,7 +412,7 @@ public class ParseObject {
    */
   @SuppressWarnings("unchecked")
   public static <T extends ParseObject> T create(Class<T> subclass) {
-    return (T) create(getClassName(subclass));
+    return (T) create(getSubclassingController().getClassName(subclass));
   }
 
   /**
@@ -497,13 +475,7 @@ public class ParseObject {
    */
   @SuppressWarnings({"unused", "unchecked"})
   public static <T extends ParseObject> T createWithoutData(Class<T> subclass, String objectId) {
-    return (T) createWithoutData(getClassName(subclass), objectId);
-  }
-
-  private static boolean isAccessible(Member m) {
-    return Modifier.isPublic(m.getModifiers())
-        || (m.getDeclaringClass().getPackage().getName().equals("com.parse")
-        && !Modifier.isPrivate(m.getModifiers()) && !Modifier.isProtected(m.getModifiers()));
+    return (T) createWithoutData(getSubclassingController().getClassName(subclass), objectId);
   }
 
   /**
@@ -515,41 +487,11 @@ public class ParseObject {
    *          The subclass type to register.
    */
   public static void registerSubclass(Class<? extends ParseObject> subclass) {
-    String className = getClassName(subclass);
-    if (className == null) {
-      throw new IllegalArgumentException("No ParseClassName annotation provided on " + subclass);
-    }
-    if (subclass.getDeclaredConstructors().length > 0) {
-      try {
-        if (!isAccessible(subclass.getDeclaredConstructor())) {
-          throw new IllegalArgumentException("Default constructor for " + subclass
-              + " is not accessible.");
-        }
-      } catch (NoSuchMethodException e) {
-        throw new IllegalArgumentException("No default constructor provided for " + subclass);
-      }
-    }
-    Class<? extends ParseObject> oldValue = objectTypes.get(className);
-    if (oldValue != null && subclass.isAssignableFrom(oldValue)) {
-      // The old class was already more descendant than the new subclass type. No-op.
-      return;
-    }
-    objectTypes.put(className, subclass);
-    if (oldValue != null && !subclass.equals(oldValue)) {
-      if (className.equals(getClassName(ParseUser.class))) {
-        ParseUser.getCurrentUserController().clearFromMemory();
-      } else if (className.equals(getClassName(ParseInstallation.class))) {
-        ParseInstallation.getCurrentInstallationController().clearFromMemory();
-      }
-    }
+    getSubclassingController().registerSubclass(subclass);
   }
 
   /* package for tests */ static void unregisterSubclass(Class<? extends ParseObject> subclass) {
-    unregisterSubclass(getClassName(subclass));
-  }
-
-  /* package for tests */ static void unregisterSubclass(String className) {
-    objectTypes.remove(className);
+    getSubclassingController().unregisterSubclass(subclass);
   }
 
   /**
@@ -3514,26 +3456,6 @@ public class ParseObject {
     synchronized (mutex) {
       saveEvent.unsubscribe(callback);
     }
-  }
-
-  /**
-   * Gets the class name based on the {@link ParseClassName} annotation associated with a class.
-   *
-   * @param clazz
-   *          The class to inspect.
-   * @return The name of the Parse class, if one is provided. Otherwise, {@code null}.
-   */
-  static String getClassName(Class<? extends ParseObject> clazz) {
-    String name = classNames.get(clazz);
-    if (name == null) {
-      ParseClassName info = clazz.getAnnotation(ParseClassName.class);
-      if (info == null) {
-        return null;
-      }
-      name = info.value();
-      classNames.put(clazz, name);
-    }
-    return name;
   }
 
   /**
--- a/Parse/src/main/java/com/parse/ParseObjectSubclassingController.java
+++ b/Parse/src/main/java/com/parse/ParseObjectSubclassingController.java
@@ -1,0 +1,139 @@
+/*
+ * Copyright (c) 2015-present, Parse, LLC.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+package com.parse;
+
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Modifier;
+import java.util.HashMap;
+import java.util.Map;
+
+/* package */ class ParseObjectSubclassingController {
+  private static class ObjectSubclassInfo {
+    public final Class<? extends ParseObject> type;
+    public final Constructor<? extends ParseObject> constructor;
+
+    public ObjectSubclassInfo(Class<? extends ParseObject> clazz) throws NoSuchMethodException, IllegalAccessException {
+      type = clazz;
+      constructor = getConstructor(clazz);
+    }
+
+    private static Constructor<? extends ParseObject> getConstructor(Class<? extends  ParseObject> clazz) throws NoSuchMethodException, IllegalAccessException {
+      Constructor<? extends ParseObject> constructor = clazz.getDeclaredConstructor();
+      if (constructor == null) {
+        throw new NoSuchMethodException();
+      }
+      int modifiers = constructor.getModifiers();
+      if (Modifier.isPublic(modifiers) || (clazz.getPackage().getName().equals("com.parse") &&
+          !(Modifier.isProtected(modifiers) || Modifier.isPrivate(modifiers)))) {
+        return constructor;
+      }
+      throw new IllegalAccessException();
+    }
+  }
+
+  private final Object mutex = new Object();
+  private final Map<String, ObjectSubclassInfo> registeredSubclasses = new HashMap<>();
+
+  /* package */ String getClassName(Class<? extends ParseObject> clazz) {
+    ParseClassName info = clazz.getAnnotation(ParseClassName.class);
+    if (info == null) {
+      throw new IllegalArgumentException("No ParseClassName annotation provided on " + clazz);
+    }
+    return info.value();
+  }
+
+  /* package */ boolean isSubclassValid(String className, Class<? extends ParseObject> clazz) {
+    ObjectSubclassInfo info = null;
+
+    synchronized (mutex) {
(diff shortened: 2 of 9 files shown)


Parse/src/main/java/com/parse/ParseObject.java
```diff
@@ -516,41 +491,11 @@ private static boolean isAccessible(Member m) {
...
-    }
-    Class<? extends ParseObject> oldValue = objectTypes.get(className);
-    if (oldValue != null && subclass.isAssignableFrom(oldValue)) {
-      // The old class was already more descendant than the new subclass type. No-op.
-      return;
-    }
-    objectTypes.put(className, subclass);
-    if (oldValue != null && !subclass.equals(oldValue)) {
-      if (className.equals(getClassName(ParseUser.class))) {
-        ParseUser.getCurrentUserController().clearFromMemory();
```
richardjrossiii: Similar to the state we have on .NET, we lose this logic during this decouple. Thoughts on a good cross-platform and extensible way to have a hook when a certain class is unregistered?

Parse/src/main/java/com/parse/ParseObject.java
```diff
@@ -361,23 +360,15 @@ public ParseObject(String theClassName) {
...
           "You must create this type of ParseObject using ParseObject.create() or the proper subclass.");
     }
 
-    // If this is an unregistered subclass, throw an exception.
-    if (!this.getClass().equals(ParseObject.class)
-        && !this.getClass().equals(objectTypes.get(theClassName))) {
-      throw new IllegalArgumentException(
-          "You must register this ParseObject subclass before instantiating it.");
-    }
-
```
grantland: did we lose this case?

Parse/src/main/java/com/parse/ParseObject.java
```diff
@@ -411,17 +402,7 @@ public ParseObject(String theClassName) {
...
-        return objectTypes.get(className).newInstance();
-      } catch (Exception e) {
-        if (e instanceof RuntimeException) {
-          throw (RuntimeException) e;
-        }
-        throw new RuntimeException("Failed to create instance of subclass.", e);
-      }
-    }
-    return new ParseObject(className);
+    return getSubclassingController().instantiate(className);
```
grantland: not sure how I feel about the name `instantiate(className)`. maybe `newInstance(className)` or `create(className)`?

Parse/src/main/java/com/parse/ParseObject.java
```diff
@@ -516,41 +491,11 @@ private static boolean isAccessible(Member m) {
    *          The subclass type to register.
    */
   public static void registerSubclass(Class<? extends ParseObject> subclass) {
-    String className = getClassName(subclass);
-    if (className == null) {
-      throw new IllegalArgumentException("No ParseClassName annotation provided on " + subclass);
-    }
```
grantland: did we lose this error condition?

Parse/src/main/java/com/parse/ParseObject.java
grantland: Could we keep this for now with a TODO to move to a hook?

The only clean way (in API terms) would be to add listeners/callbacks to `ParseObjectSubclassingController` for unregister that `ParseUser` can subscribe to, but that involves a bunch of class allocations...

Parse/src/main/java/com/parse/ParseObjectSubclassingController.java
```diff
@@ -0,0 +1,133 @@
...
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Modifier;
+import java.util.HashMap;
+import java.util.Map;
+
+/* package */ class ParseObjectSubclassingController {
+  private static class ObjectSubclassInfo {
+    public final Class<? extends ParseObject> type;
+    public final String className;
+    public final Constructor<? extends ParseObject> constructor;
```
grantland: is there any performance gain from caching the `constructor`? we didn't before and utilized `clazz#newInstance()`, but now we have to keep reference to a `Constructor`.

Parse/src/main/java/com/parse/ParseObjectSubclassingController.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    private static Constructor<? extends ParseObject> getConstructor(Class<? extends  ParseObject> clazz) throws NoSuchMethodException {
+      Constructor<? extends ParseObject> constructor = clazz.getDeclaredConstructor();
+      int modifiers = constructor.getModifiers();
+      if (Modifier.isPublic(modifiers) || (clazz.getPackage().getName().equals("com.parse") &&
+          !(Modifier.isProtected(modifiers) || Modifier.isPrivate(modifiers)))) {
+        return constructor;
+      }
+      throw new NoSuchMethodException();
```
grantland: do we want to keep the case of the default constructor not being available vs not existing separate? it seems we lost granularity when throwing errors for these two cases

Parse/src/main/java/com/parse/ParseObjectSubclassingController.java
```diff
@@ -0,0 +1,133 @@
...
+    private static String getClassName(Class<? extends ParseObject> clazz) {
+      ParseClassName info = clazz.getAnnotation(ParseClassName.class);
+      if (info == null) {
+        return null;
+      }
+      return  info.value();
+    }
+  }
+
+  private Object mutex;
```
grantland: `final`

Parse/src/main/java/com/parse/ParseObjectSubclassingController.java
```diff
@@ -0,0 +1,133 @@
...
+  private Map<String, ObjectSubclassInfo> registeredSubclasses;
```
grantland: `final`

Parse/src/main/java/com/parse/ParseObjectSubclassingController.java
```diff
@@ -0,0 +1,133 @@
...
+
+  public ParseObjectSubclassingController() {
+    mutex = new Object();
+    registeredSubclasses = new HashMap<>();
```
grantland: might be cleaner to just move instantiation inline with declaration since these are `final`

Parse/src/main/java/com/parse/ParseObjectSubclassingController.java
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  String getClassName(Class<? extends ParseObject> clazz) {
```
grantland: nit: `/* package */`

Parse/src/main/java/com/parse/ParseObjectSubclassingController.java
```diff
@@ -0,0 +1,133 @@
...
+    return ObjectSubclassInfo.getClassName(clazz);
+  }
+
+  boolean isTypeValid(String className, Class<? extends ParseObject> clazz) {
```
grantland: nit: `/* package */`

Parse/src/main/java/com/parse/ParseObjectSubclassingController.java
```diff
@@ -0,0 +1,133 @@
...
+    synchronized (mutex) {
+      info = registeredSubclasses.get(className);
+    }
+
+    return info == null
+      ? clazz == ParseObject.class
+      : info.type == clazz;
+  }
+
+  void registerSubclass(Class<? extends ParseObject> clazz) {
```
grantland: nit: `/* package */`

Parse/src/main/java/com/parse/ParseObjectSubclassingController.java
```diff
@@ -0,0 +1,133 @@
...
+    if (!ParseObject.class.isAssignableFrom(clazz)) {
+      throw new IllegalArgumentException("Cannot register a type that is not a subclass of ParseObject");
+    }
```
grantland: Could we ever hit this condition? We have strong typing already in the method signature that `clazz` has to `extend ParseObject`

Parse/src/main/java/com/parse/ParseObjectSubclassingController.java
```diff
@@ -0,0 +1,133 @@
...
+
+    String className = ObjectSubclassInfo.getClassName(clazz);
```
grantland: we should be consistent with calling into `ObjectSubclassInfo.getClassName(clazz)` vs `ParseObjectSubclassingController.getClassName(clazz)`

Parse/src/main/java/com/parse/ParseObjectSubclassingController.java
```diff
@@ -0,0 +1,133 @@
...
+
+    synchronized (mutex) {
+      ObjectSubclassInfo previousInfo = registeredSubclasses.get(className);
+      if (previousInfo != null) {
+        if (clazz.isAssignableFrom(previousInfo.type)) {
+          // Previous subclass is more specific or equal to the current type, do nothing.
+          return;
```
grantland: any reason why we have return here but not in the next condition?

Parse/src/main/java/com/parse/ParseObjectSubclassingController.java
```diff
@@ -0,0 +1,133 @@
...
+        } else if (previousInfo.type.isAssignableFrom(clazz)) {
+          // Previous subclass is parent of new child subclass, fallthrough and actually
+          // register this class.
+                    /* Do nothing */
```
grantland: nit: weird indentation

Parse/src/main/java/com/parse/ParseObjectSubclassingController.java
```diff
@@ -0,0 +1,133 @@
...
+        registeredSubclasses.put(className, new ObjectSubclassInfo(clazz));
+      } catch (NoSuchMethodException ex) {
+        throw new IllegalArgumentException(
+          "Cannot register a type that does not implement the default constructor!"
+        );
+      }
+    }
+  }
+
+  void unregisterSubclass(Class<? extends ParseObject> clazz) {
```
grantland: nit: `/* package */`

Parse/src/main/java/com/parse/ParseObjectSubclassingController.java
```diff
@@ -0,0 +1,133 @@
...
+    String className = getClassName(clazz);
+
+    synchronized (mutex) {
+      registeredSubclasses.remove(className);
+    }
+  }
+
+  ParseObject instantiate(String className) {
```
grantland: nit: `/* package */`

Parse/src/test/java/com/parse/ParseRoleTest.java
```diff
@@ -155,7 +157,7 @@ public void testPutFailureWithInvalidNameValueSet() {
   public void testGetQuery() {
     ParseQuery query = ParseRole.getQuery();
```
grantland: can we add `<?>` here?

Parse/src/test/java/com/parse/SubclassTest.java
```diff
@@ -85,8 +90,8 @@ public void setUp() throws Exception {
   @After
   public void tearDown() throws Exception {
     ParseObject.unregisterParseSubclasses();
-    ParseObject.unregisterSubclass("Person");
-    ParseObject.unregisterSubclass("ClassWithDirtyingConstructor");
+    ParseObject.unregisterSubclass(Person.class);
+    ParseObject.unregisterSubclass(ClassWithDirtyingConstructor.class);
```
grantland: any reason why we're changing this?

Parse/src/test/java/com/parse/SubclassTest.java
```diff
@@ -102,7 +107,7 @@ public void testUnregisteredConstruction() throws Exception {
     try {
       new UnregisteredClass();
     } finally {
-      ParseObject.unregisterSubclass("UnregisteredClass");
+      ParseObject.unregisterSubclass(UnregisteredClass.class);
```
grantland: any reason why we're changing this?

Parse/src/test/java/com/parse/SubclassTest.java
```diff
@@ -132,15 +137,23 @@ public void testRegisteringSubclassesUsesMostDescendantSubclass() throws Excepti
       // throw an exception in unit test environment.
       ParseCurrentUserController controller = mock(ParseCurrentUserController.class);
       ParseCorePlugins.getInstance().registerCurrentUserController(controller);
-      assertEquals(ParseUser.class, ParseObject.create("_User").getClass());
+      assertEquals(ParseUser.class, ParseObject.create("_User" ).getClass());
```
grantland: nit: no space after `"_User"`

Parse/src/test/java/com/parse/SubclassTest.java
```diff
@@ -132,15 +137,23 @@ public void testRegisteringSubclassesUsesMostDescendantSubclass() throws Excepti
...
       ParseObject.registerSubclass(MyUser.class);
-      assertEquals(MyUser.class, ParseObject.create("_User").getClass());
+      assertEquals(MyUser.class, ParseObject.create("_User" ).getClass());
```
grantland: nit: no space after `"_User"`

Parse/src/test/java/com/parse/SubclassTest.java
```diff
@@ -132,15 +137,23 @@ public void testRegisteringSubclassesUsesMostDescendantSubclass() throws Excepti
...
-      ParseObject.registerSubclass(MyUser2.class);
-      assertEquals(MyUser2.class, ParseObject.create("_User").getClass());
+
+      // This is expected to fail as MyUser2 and MyUser are not directly related.
+      try {
+        ParseObject.registerSubclass(MyUser2.class);
+        fail();
+      } catch (IllegalArgumentException ex) {
+        /* expected */
+      }
```
grantland: Can you reply as to why we're making this change?

Parse/src/test/java/com/parse/SubclassTest.java
```diff
@@ -132,15 +137,23 @@ public void testRegisteringSubclassesUsesMostDescendantSubclass() throws Excepti
...
+
+      assertEquals(MyUser.class, ParseObject.create("_User").getClass());
     } finally {
-      ParseObject.unregisterSubclass("_User");
+      ParseObject.unregisterSubclass(ParseUser.class);
```
grantland: any reason why we're changing this?

Parse/src/main/java/com/parse/ParseObjectSubclassingController.java
```diff
@@ -0,0 +1,133 @@
...
+
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Modifier;
+import java.util.HashMap;
+import java.util.Map;
+
+/* package */ class ParseObjectSubclassingController {
+  private static class ObjectSubclassInfo {
+    public final Class<? extends ParseObject> type;
+    public final String className;
```
grantland: Is this used anywhere?

Parse/src/main/java/com/parse/ParseObjectSubclassingController.java
```diff
@@ -0,0 +1,133 @@
...
+ */
+package com.parse;
+
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Modifier;
+import java.util.HashMap;
+import java.util.Map;
+
+/* package */ class ParseObjectSubclassingController {
+  private static class ObjectSubclassInfo {
```
grantland: Do we actually need this abstraction? we only seem to use `constructor`

richardjrossiii: This keeps it in parity with .NET and iOS, even though for Java it's mostly useless, I do agree.

All other platforms currently require metadata for different things (.NET for property event handlers, iOS for property getter/setter automation). We've briefly discussed in the past making ParseObject subclasses observable, which would then need more metadata here.

Parse/src/main/java/com/parse/ParseObjectSubclassingController.java
richardjrossiii: No, not currently. I can remove it.

Parse/src/main/java/com/parse/ParseObjectSubclassingController.java
richardjrossiii: I'm not 100% sure about performance. I did some rudimentary searching, but found no conclusive results.

Mostly just caching the constructor for parity with .NET. There's a minute difference as to what kind of exceptions are thrown by `Constructor.newInstance()` vs `Class.newInstance()` as well...

Parse/src/main/java/com/parse/ParseObjectSubclassingController.java
richardjrossiii: Wilco.

Parse/src/main/java/com/parse/ParseObjectSubclassingController.java
richardjrossiii: Theoretically yes, especially if we go the route of runtime searching for automatic registration. It's mostly just a sanity check, though.

Parse/src/main/java/com/parse/ParseObjectSubclassingController.java
richardjrossiii: This is intentional fallthrough, when the old registration is a parent of the new registration:

Hierarchy: ParseObject -> A ->  B -> C

If we have 'A' registered, and now we register 'B', we should fall through inthis case and perform the operation.

Parse/src/test/java/com/parse/SubclassTest.java
richardjrossiii: The 'unregisterSubclass(String)` overload was removed.

Parse/src/test/java/com/parse/SubclassTest.java
richardjrossiii: This is the inheritance problem we discussed yesterday, and the exception that existed on all other platforms exception Java.

Parse/src/main/java/com/parse/ParseObject.java
richardjrossiii: That specific message? Yes. That error case is still handled by `isTypeValid()`.

grantland:
Passing back due to failing CI

richardjrossiii:
Tests should pass again.

Parse/src/main/java/com/parse/ParseObjectSubclassingController.java
grantland: Could we re-add this abstraction when we need it? Otherwise, this abstraction would look super over-architected to anyone trying to wrap their head around it.

Parse/src/main/java/com/parse/ParseObject.java
grantland: Moved to `ParseObjectSubclassingController#getClassName(clazz)`

Parse/src/main/java/com/parse/ParseObject.java
grantland: ping

Parse/src/main/java/com/parse/ParseObjectSubclassingController.java
grantland: ping

Parse/src/test/java/com/parse/ParseRoleTest.java
grantland: ping

Parse/src/test/java/com/parse/SubclassTest.java
grantland: https://github.com/ParsePlatform/Parse-SDK-Android/pull/245/files#r44582587

Parse/src/test/java/com/parse/SubclassTest.java
grantland: https://github.com/ParsePlatform/Parse-SDK-Android/pull/245/files#r44582587

Parse/src/main/java/com/parse/ParseObjectSubclassingController.java
richardjrossiii: Ok, sounds good to me. Will update.

Parse/src/main/java/com/parse/ParseObject.java
richardjrossiii: Yep, already implemented in fact: https://github.com/ParsePlatform/Parse-SDK-Android/pull/245/files#diff-d1fc53b1cc8d9fefba8e9a00706881e1R104

richardjrossiii:
Ignore facebook bot. it's drunk.

grantland:
LGTM!

PR #249, opened 2015-11-13
Add more log for push

wangmengyan95:
1. Add log when we recive deviceToken from GCM.
2. Add log when developers try to use PPNS but miss something.
3. Add log to recommend developers to use GCM

### DIFF
--- a/Parse/src/main/java/com/parse/GcmRegistrar.java
+++ b/Parse/src/main/java/com/parse/GcmRegistrar.java
@@ -182,6 +182,8 @@ import bolts.Task;
     String registrationId = intent.getStringExtra(REGISTRATION_ID_EXTRA);
 
     if (registrationId != null && registrationId.length() > 0) {
+      PLog.v(TAG, "Received deviceToken <" + registrationId + "> from GCM.");
+
       ParseInstallation installation = ParseInstallation.getCurrentInstallation();
       // Compare the new deviceToken with the old deviceToken, we only update the
       // deviceToken if the new one is different from the old one. This does not follow google
--- a/Parse/src/main/java/com/parse/ParsePushBroadcastReceiver.java
+++ b/Parse/src/main/java/com/parse/ParsePushBroadcastReceiver.java
@@ -141,6 +141,7 @@ public class ParsePushBroadcastReceiver extends BroadcastReceiver {
       PLog.e(TAG, "Can not get push data from intent.");
       return;
     }
+    PLog.v(TAG, "Received push data: " + pushDataStr);
 
     JSONObject pushData = null;
     try {


wangmengyan95:
@hallucinogen it would be great if you can also tay a look.

Parse/src/main/java/com/parse/ManifestInfo.java
```diff
@@ -260,6 +261,40 @@ public static PushType getPushType() {
               PLog.e(TAG, "Cannot use GCM for push because the app manifest is missing some " +
                   "required declarations. Please " + getGcmManifestMessage());
             }
+          } else if (hasAnyPPNSSpecificDeclaration) {
```
hallucinogen: I think we clean this up on purpose. We used to have this flag. @grantland

Parse/src/main/java/com/parse/ManifestInfo.java
```diff
@@ -459,6 +492,16 @@ private static boolean hasAnyGcmSpecificDeclaration() {
     return false;
   }
 
+  private static boolean hasAnyPPNSSpecificDeclaration() {
+    Context context = getContext();
+    if (hasRequestedPermissions(context, "android.permission.RECEIVE_BOOT_COMPLETED") ||
```
hallucinogen: I don't think we want this.

Parse/src/main/java/com/parse/ManifestInfo.java
```diff
@@ -459,6 +492,16 @@ private static boolean hasAnyGcmSpecificDeclaration() {
...
+        getReceiverInfo(ParseBroadcastReceiver.class) != null) {
```
hallucinogen: And this is incorrect. Even with GCM, we'll still use this.

Parse/src/main/java/com/parse/GcmRegistrar.java
```diff
@@ -182,6 +182,8 @@ public Void then(Task<String> task) {
     String registrationId = intent.getStringExtra(REGISTRATION_ID_EXTRA);
 
     if (registrationId != null && registrationId.length() > 0) {
+      PLog.v(TAG, "Get deviceToken " + registrationId + " from GCM.");
```
hallucinogen: I think the only part of this PR I agree is this line.

Parse/src/main/java/com/parse/ParsePushBroadcastReceiver.java
```diff
@@ -138,7 +138,9 @@ public void onReceive(Context context, Intent intent) {
   protected void onPushReceive(Context context, Intent intent) {
     JSONObject pushData = null;
     try {
-      pushData = new JSONObject(intent.getStringExtra(KEY_PUSH_DATA));
+      String pushDataStr = intent.getStringExtra(KEY_PUSH_DATA);
+      PLog.v(TAG, pushDataStr);
```
wangmengyan95: Not sure whether this introduces too much overhead for production, but for debugging purpose, I think this is necessary since there is no reliable for us to tell whether a device has received a push notification or not.

hallucinogen: Should be fine. We also did something similar on .NET.

hallucinogen:
yay!

Parse/src/main/java/com/parse/ParsePushBroadcastReceiver.java
grantland: `"Received push data: " + pushDataStr`

Parse/src/main/java/com/parse/GcmRegistrar.java
grantland: `"Received deviceToken <" + registrationId + "> from GCM."`

grantland:
LGTM

Issue #260, opened 2015-11-20
PushBroadcastReceiver: java.lang.NullPointerException


AraBisTiC:
Push is working for mostly without issues, however just got this crash from one of my users, any ideas?

```java
Fatal Exception: java.lang.RuntimeException: Unable to start receiver dev.moj.pad.util.PushBroadcastReceiver: java.lang.NullPointerException
       at android.app.ActivityThread.handleReceiver(ActivityThread.java:2237)
       at android.app.ActivityThread.access$1500(ActivityThread.java:131)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1272)
       at android.os.Handler.dispatchMessage(Handler.java:99)
       at android.os.Looper.loop(Looper.java:137)
       at android.app.ActivityThread.main(ActivityThread.java:4754)
       at java.lang.reflect.Method.invokeNative(Method.java)
       at java.lang.reflect.Method.invoke(Method.java:511)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:786)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:553)
       at dalvik.system.NativeStart.main(NativeStart.java)
Caused by java.lang.NullPointerException
       at org.json.JSONTokener.nextCleanInternal(JSONTokener.java:116)
       at org.json.JSONTokener.nextValue(JSONTokener.java:94)
       at org.json.JSONObject.(JSONObject.java)
       at org.json.JSONObject.(JSONObject.java)
       at com.parse.ParsePushBroadcastReceiver.onPushOpen(ParsePushBroadcastReceiver.java:198)
       at com.parse.ParsePushBroadcastReceiver.onReceive(ParsePushBroadcastReceiver.java:123)
       at android.app.ActivityThread.handleReceiver(ActivityThread.java:2230)
       at android.app.ActivityThread.access$1500(ActivityThread.java:131)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1272)
       at android.os.Handler.dispatchMessage(Handler.java:99)
       at android.os.Looper.loop(Looper.java:137)
       at android.app.ActivityThread.main(ActivityThread.java:4754)
       at java.lang.reflect.Method.invokeNative(Method.java)
       at java.lang.reflect.Method.invoke(Method.java:511)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:786)
[...]

PR #261, opened 2015-11-20
Add null checking for push raw data

wangmengyan95:
#260, we get some exceptions from developers about getting null push raw data. Not sure why this happen but it make sense to add some null checking.


hallucinogen:
nice one

Parse/src/main/java/com/parse/ParsePushBroadcastReceiver.java
```diff
@@ -136,9 +136,15 @@ public void onReceive(Context context, Intent intent) {
    *      An {@code Intent} containing the channel and data of the current push notification.
    */
   protected void onPushReceive(Context context, Intent intent) {
+    String pushDataStr = intent.getStringExtra(KEY_PUSH_DATA);
+    if (pushDataStr == null) {
+      PLog.e(TAG, "Can not get push data from intent.");
```
grantland: s/Can not get/Unable to read/g

PR #284, opened 2015-12-08
Add better, more flexible APIs for SDK initialization.

richardjrossiii:
## This PR includes public API changes

Proposing a new initialization API, to allow for more flexibility and clarity in the future:

````java
Parse.Initialize(new Parse.Configuration.Builder(context)
    .applicationId(...)
    .clientKey(...)
    .build()
);
````

No APIs are being directly deprecated at this time, only additions for the time being.

Equivalent PRs for .NET: https://github.com/ParsePlatform/Parse-SDK-dotNET/pull/79, and iOS: https://github.com/ParsePlatform/Parse-SDK-iOS-OSX/pull/570.

-------

This PR hopes to unify some of these initialization settings int o a single 'configuration' class that can be easily used to intialize parse in a much cleaner way, especially as we add even more initialization options.

### DIFF
--- a/Parse/src/main/java/com/parse/Parse.java
+++ b/Parse/src/main/java/com/parse/Parse.java
@@ -23,6 +23,8 @@ import java.io.IOException;
 import java.io.RandomAccessFile;
 import java.io.UnsupportedEncodingException;
 import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: Parse/src/main/java/com/parse/Parse.java [hunk 1]
```diff
@@ -35,6 +36,86 @@
...
+
+    private Configuration(Builder builder) {
+      this.context = builder.context;
+      this.applicationId = builder.applicationId;
+      this.clientKey = builder.clientKey;
+      this.localDataStoreEnabled = builder.localDataStoreEnabled;
+      this.interceptors = builder.interceptors;
+    }
+
+    public static final class Builder {
```
grantland: nit: class definitions above instance

Parse/src/main/java/com/parse/Parse.java
```diff
@@ -35,6 +36,86 @@
...
+      private Context context;
+      private String applicationId;
+      private String clientKey;
+      private boolean localDataStoreEnabled;
+      private List<ParseNetworkInterceptor> interceptors = new ArrayList<>();
```
grantland: let's convert this to lazy since utilizing interceptors seems like the minority use case vs no interceptors.

Parse/src/main/java/com/parse/Parse.java
```diff
@@ -35,6 +36,86 @@
...
+
+      public Builder(Context context) {
+        this.context = context;
+      }
```
grantland: This will be a pain for us if we ever want to add vanilla Java support, but we can't really get around it and the problem exists within it's parent class already.

Parse/src/main/java/com/parse/Parse.java
```diff
@@ -35,6 +36,86 @@
...
+      return clientKey;
+    }
+
+    public boolean isLocalDataStoreEnabled() {
+      return localDataStoreEnabled;
+    }
+
+    public List<ParseNetworkInterceptor> networkInterceptors() {
+      return interceptors;
+    }
```
grantland: I don't see accessors being useful and we can always add them back if there's a demand

grantland:
documentation? tests?

Parse/src/main/java/com/parse/Parse.java
richardjrossiii: It's it as simple as adding a default constructor, though?

Parse/src/main/java/com/parse/Parse.java
richardjrossiii: Sounds good to me.

richardjrossiii:
Similar to other platforms, documentation forthcoming after we finalize APIs across platforms. I'll add a test or two though.

Parse/src/main/java/com/parse/Parse.java
richardjrossiii: :+1:

Parse/src/main/java/com/parse/Parse.java
grantland: The problem is that `android.content.Context` doesn't exist without the Android APIs, which would mean this method would cause this class to fail to load in a pure Java environment due to `import android.content.Context` that's required.

However, this is already a problem with `com.parse.Parse` anyway, so this isn't really adding any additional problems since we'd need to either create a new class `com.parse.ParseJava` or break `com.parse.Parse` anyway.

richardjrossiii:
Tests still forthcoming, fixed nits.

Parse/src/main/java/com/parse/Parse.java
```diff
@@ -46,6 +110,8 @@
   private static boolean isLocalDatastoreEnabled;
   private static OfflineStore offlineStore;
 
+  private static Configuration configuration;
```
grantland: I don't think we need to hold this, we can just pass in `configuration.interceptors` through to `initializeParseHttpClientsWithParseNetworkInterceptors()`

Parse/src/main/java/com/parse/Parse.java
```diff
@@ -35,6 +36,69 @@
...
+    private final String clientKey;
+    private final boolean localDataStoreEnabled;
+    private final List<ParseNetworkInterceptor> interceptors;
+
+    private Configuration(Builder builder) {
+      this.context = builder.context;
+      this.applicationId = builder.applicationId;
+      this.clientKey = builder.clientKey;
+      this.localDataStoreEnabled = builder.localDataStoreEnabled;
+      this.interceptors = builder.interceptors;
```
grantland: We'd want to use `Collections.unmodifiableList()` here

Parse/src/test/java/com/parse/ParseClientConfigurationTest.java
```diff
@@ -0,0 +1,60 @@
...
+    Parse.Configuration configurationA = builder.build();
+    builder.addNetworkInterceptor(interceptorB);
+    Parse.Configuration configurationB = builder.build();
+
+    assertFalse(configurationA.interceptors.contains(interceptorB));
+    assertTrue(configurationB.interceptors.contains(interceptorB));
+
+    try {
+      configurationA.interceptors.add(interceptorB);
+      fail();
```
grantland: Could you add a message here "`interceptors` should not be mutable"?

grantland:
LGTM, one nit

Parse/src/test/java/com/parse/ParseClientConfigurationTest.java
richardjrossiii: :+1:

richardjrossiii:
Added message, squashed, and good to go! Will merge once travis passes.

PR #297, opened 2015-12-14
Properly initialize `Configuration.interceptors` when using old-style APIs.

### LINKED ISSUE #296: NPE with new Parse initializer
From https://github.com/ParsePlatform/Parse-SDK-Android/commit/7cec3b41e4f96dc43b4d97e4783b03967aed41f0#commitcomment-14948582:

There seems to be a bug here...if I'm  using Stetho the builder will be not be instantitated and this.interceptors appears to be a null object.  I was able to workaround by downgrading to 1.11.0:

```java
Parse.addParseNetworkInterceptor(new ParseStethoInterceptor());
Parse.initialize(context, parseAppId, parseClientKey);
```

```
E/AndroidRuntime( 3595): Caused by: java.lang.NullPointerException: Attempt to invoke interface method 'void java.util.List.clear()' on a null object reference
E/AndroidRuntime( 3595): 	at com.parse.Parse$Configuration$Builder.setNetworkInterceptors(Parse.java:157)
E/AndroidRuntime( 3595): 	at com.parse.Parse$Configuration$Builder.access$600(Parse.java:47)
E/AndroidRuntime( 3595): 	at com.parse.Parse.initialize(Parse.java:339)
```

richardjrossiii:
This happened because `this.interceptors` was null when `clear()` was invoked.

Fixes #296.

### DIFF
--- a/Parse/src/main/java/com/parse/Parse.java
+++ b/Parse/src/main/java/com/parse/Parse.java
@@ -152,9 +152,14 @@ public class Parse {
         return this;
       }
 
-      private Builder setNetworkInterceptors(Collection<ParseNetworkInterceptor> interceptors) {
-        if (interceptors != null) {
+      /* package for tests */ Builder setNetworkInterceptors(Collection<ParseNetworkInterceptor> interceptors) {
+        if (this.interceptors == null) {
+          this.interceptors = new ArrayList<>();
+        } else {
           this.interceptors.clear();
+        }
+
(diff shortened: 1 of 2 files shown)


grantland:
Can we `/* package */` protect this so we can add unit tests?

Also, it seems like we should be clearing or nullifying `this.interceptors` if we pass in `null`, right?

grantland:
LGTM

Issue #282, opened 2015-12-05
Possible data caching issue


alexblack:
I'm running into a strange issue, and I was wondering if you could offer any initial insights/intuitions while I look into it further.  We're using Android Parse SDK 1.11.0.

High level repro steps:
1. Create an invoice on device 1 using our app, this creates an Invoice object and stores it at parse
2. Device 2 picks this up, gets a copy of the object, stores it locally, it looks good
3. On device 1, we now make a change to one of the fields on this object, a JSON field (containing a map or array)
4. Device 2 picks this up, gets a copy of the object, stores it locally, but it doesn't look good.  The ParseObject it got does not match what is shown on the Parse.com dashboard, it is out of date and doesn't reflect the changes made by device 1 at parse.com.

More details:
a. We can repro this when Device 2 is Android and Device 1 is Android, we can also repro this when Device 2 is Android and device 1 is iOS.  We cannot repro this if device 2 is iOS.  I think this tells us its likely a bug in the Android SDK (assuming iOS SDK behaviour is correct)

b. The code used in Step 2 to get the object from parse is like this:

```
 public Task<List<T>> get() {
    final ParseQuery<T> query = ParseQuery.getQuery(cls);
    query.whereEqualTo("account", account);
    query.setLimit(LIMIT);
    return pageAllResults(query);
  }

  private Task<List<T>> pageAllResults(final ParseQuery<T> query, final List<T> allResults, final int offset, final int page) {
    query.setSkip(offset);
    Log.d(TAG, String.format("pageAllResults (Offset=%d, Page=%d)", offset, page));
    return query.findInBackground().onSuccessTask(new Continuation<List<T>, Task<List<T>>>() {
      @Override
      public Task<List<T>> then(Task<List<T>> task) throws Exception {
        List<T> results = task.getResult();

        allResults.addAll(results);

        if (results.size() == LIMIT)
          return pageAllResults(query, allResults, offset + results.size(), page + 1);
        else
          return Task.forResult(allResults);
      }
    });
  }
```

c. We can't repro this issue for primitive fields, like a string.  So, for example, if on Device 1 you change a JSON field and a primitive field, then, when Device 2 gets the object using the query it will have the updated primitive field value, but not the updated JSON field value.

d. We tried reproing this by making manual changes at parse.com, and we haven't yet been able to repro the problem this way.  Summarizing: if we make the changes using the Android SDK or the iOS SDK then we can repro the problem.  

[...]

PR #303, opened 2015-12-17
Remove caching of auto-converted values

grantland:
We cache auto-convert `JSONObject`/`JSONArray` values from `Map`/`List` which causes confusion when the cached value gets returned instead of fresh values from the server. This PR removes this cache and instead auto-converts `JSONObject`/`JSONArray` values from `Map`/`List` on every call to `getJSONObject(key)` and `getJSONArray(key)`.

### DIFF
--- a/Parse/src/main/java/com/parse/ParseObject.java
+++ b/Parse/src/main/java/com/parse/ParseObject.java
@@ -3139,10 +3139,6 @@ public class ParseObject {
    *          The key to access the value for.
    * @return {@code null} if there is no such key or if it is not a {@link JSONArray}.
    */
-  //TODO (grantland): Do not auto-convert JSONArray and List (v2)
-  // Calling this on an untouched fetched object will mark the object as dirty due to mutable
-  // containers and auto List <-> JSONArray conversion, since arrays are converted to List
-  // by default.
   public JSONArray getJSONArray(String key) {
     synchronized (mutex) {
       checkGetAccess(key);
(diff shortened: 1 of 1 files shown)


grantland:
Resolves #282

Parse/src/main/java/com/parse/ParseEncoder.java
```diff
@@ -29,11 +28,19 @@
...
+        || value instanceof Date
+        || value instanceof List
+        || value instanceof Map
+        || value instanceof byte[]
+        || value == JSONObject.NULL
+        || value instanceof ParseObject
+        || value instanceof ParseACL
+        || value instanceof ParseFile
+        || value instanceof ParseGeoPoint
+        || value instanceof ParseRelation;
```
grantland: Other than re-formatting, I've removed `JSONObject` and `JSONArray`

Parse/src/main/java/com/parse/ParseRESTAnalyticsCommand.java
```diff
@@ -26,43 +26,46 @@
...
-    commandParameters.put("at", NoObjectsEncoder.get().encode(new Date()));
     return new ParseRESTAnalyticsCommand(
-        httpPath, ParseHttpRequest.Method.POST, commandParameters, sessionToken);
+        httpPath, ParseHttpRequest.Method.POST, parameters, sessionToken);
+  }
+
+  public ParseRESTAnalyticsCommand(
+      String httpPath,
+      ParseHttpRequest.Method httpMethod,
+      JSONObject parameters,
```
grantland: Note: `Map<String, ?>` -> `JSONObject` breaks compatibility with ParseCrashreporting

nlutsenko:
Lovely!

grantland:
Split into two PRs. Combining and unnecessary back and forth conversions moved to #305

hallucinogen:
I this is included in #305. LGTM anyway.

PR #304, opened 2015-12-17
Avoid cleaning currentUser authData in setState

### LINKED ISSUE #269: ParseFacebookUtils.isLinked(user) returning false after executing query with 'user' key included
I have Parse object named 'Activity' on back end, which has a pointer to 'User' object.
In my android app, after logging in or linking through ParseFacebookUtils, if I made a parse query on 'Activity' and include key 'user' to that query for fetching users' details, then ParseFacebookUtils.isLinked(ParseUser.getCurrentUser()) starts returning false after that query. However if I kill android app and relaunch it, then ParseFacebookUtils.isLinked(ParseUser.getCurrentUser()) starts returning true again. It seems as for some reason some cached data is getting affected after making query with 'user' key included. This does not happen if query is made without including key 'user'.
This happens if at least one of the fetched 'Activities' has current logged in user set as the 'user' pointer.

wangmengyan95:
Fixes #269

### DIFF
--- a/Parse/src/main/java/com/parse/ParseUser.java
+++ b/Parse/src/main/java/com/parse/ParseUser.java
@@ -503,10 +503,17 @@ public class ParseUser extends ParseObject {
 
   @Override
   /* package */ void setState(ParseObject.State newState) {
-    // Avoid clearing sessionToken when updating the current user's State via ParseQuery result
-    if (isCurrentUser() && getSessionToken() != null
-            && newState.get("sessionToken") == null) {
-      newState = newState.newBuilder().put("sessionToken", getSessionToken()).build();
+    if (isCurrentUser()) {
+      State.Builder newStateBuilder = newState.newBuilder();
+      // Avoid clearing sessionToken when updating the current user's State via ParseQuery result
+      if (getSessionToken() != null && newState.get(KEY_SESSION_TOKEN) == null) {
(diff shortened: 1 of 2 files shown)


Parse/src/main/java/com/parse/ParseUser.java
```diff
@@ -504,9 +504,12 @@ private void restoreAnonymity(Map<String, String> anonymousData) {
...
     // Avoid clearing sessionToken when updating the current user's State via ParseQuery result
-    if (isCurrentUser() && getSessionToken() != null
-            && newState.get("sessionToken") == null) {
-      newState = newState.newBuilder().put("sessionToken", getSessionToken()).build();
+    if (isCurrentUser() && getSessionToken() != null && newState.get(KEY_SESSION_TOKEN) == null) {
+      newState = newState.newBuilder().put(KEY_SESSION_TOKEN, getSessionToken()).build();
+    }
+    // Avoid clearing authData when updating the current user's State via ParseQuery result
+    if (isCurrentUser() && getAuthData() != null && newState.get(KEY_AUTH_DATA) == null) {
+      newState = newState.newBuilder().put(KEY_AUTH_DATA, getAuthData()).build();
```
grantland: Nit: Let's move `isCurrentUser()` check out a level so we don't have to do it twice

grantland:
One nit, otherwise LGTM

PR #305, opened 2015-12-17
Prevent JSONObject/Array + Map/List combinations and unnecessary conversions back and forth

grantland:
We have a bunch of cases where we combine `JSONObject`/`JSONArray` and `Map`/`List` and a bunch of cases where unnecessarily convert `JSONObject` <-> `Map` and `JSONArray` <-> `List` back and forth. This PR simplifies things by removing `JSONObject` and `JSONArray` encoding so that we don't unnecessarily combine or convert back and forth.

* Convert ParseConfig JSON constructor to factory method
* Simplify ParseConfig tests to utilize Maps/Lists
* Do not convert Map -> JSONObject -> Map -> JSONObject for ParseAnalytics
* Do not combine JSONObject/Array with Map/List for ParseRESTObjectBatchCommand
* Do not combine Map & JSONObject for ParsePush
* Do not allow encoding JSONObject/JSONArray

Note: This breaks compatibility with ParseCrashReporting

Depends on #303

### DIFF
--- a/Parse/src/main/java/com/parse/ParseAnalytics.java
+++ b/Parse/src/main/java/com/parse/ParseAnalytics.java
@@ -13,6 +13,8 @@ import android.content.Intent;
 import org.json.JSONException;
 import org.json.JSONObject;
 
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.Map;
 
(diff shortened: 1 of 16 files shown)


hallucinogen:
LGTM. Consistency + Performance ftw

yuzeh:
This change broke backwards compatibility on CloudFunction calls when I was updating my app from 1.10.3 -- the values in the second argument of `ParseCloud.callFunctionInBackground` cannot be JSONObjects or JSONArrays.

What is the blessed way for encoding complex objects going forward? Is there a simple answer (e.g. "use  `java.util.List`s and `java.util.Map`s only", "use `org.json` objects only") to this question?

yuzeh:
@grantland curious if there is a good response to my question above. Thanks!

grantland:
@yuzeh I'd recommend using Java types such as `List` and `Map`.

yuzeh:
Great, thanks!

PR #306, opened 2015-12-17
Remove Parse Gradle plugin from ParseStarterProject

grantland:
No longer relevant without ParseCrashReporting support


hallucinogen:
cleancorgi

PR #311, opened 2015-12-18
Add copying

grantland:
* ParseQuery
* ParseACL
* ParseGeoPoint

### DIFF
--- a/Parse/src/main/java/com/parse/ParseACL.java
+++ b/Parse/src/main/java/com/parse/ParseACL.java
@@ -103,32 +103,41 @@ public class ParseACL {
     return getDefaultACLController().get();
   }
 
+  // State
+  private final Map<String, Permissions> permissionsById = new HashMap<>();
   private boolean shared;
+
   /**
    * A lazy user that hasn't been saved to Parse.
    */
   //TODO (grantland): This should be a list for multiple lazy users with read/write permissions.
   private ParseUser unresolvedUser;
 
-  private Map<String, Permissions> permissionsById;
-
   /**
    * Creates an ACL with no permissions granted.
    */
   public ParseACL() {
-    permissionsById = new HashMap<>();
+    // do nothing
   }
 
-  /* package */ ParseACL copy() {
-    ParseACL copy = new ParseACL();
-    for (String id : permissionsById.keySet()) {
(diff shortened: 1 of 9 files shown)


Parse/src/main/java/com/parse/ParseACL.java
```diff
@@ -103,32 +103,41 @@ public static void setDefaultACL(ParseACL acl, boolean withAccessForCurrentUser)
...
-    ParseACL copy = new ParseACL();
-    for (String id : permissionsById.keySet()) {
-      copy.permissionsById.put(id, new Permissions(permissionsById.get(id)));
+  /**
+   * Creates a copy of {@code acl}.
+   *
+   * @param acl
+   *          The acl to copy.
+   */
+  public ParseACL(ParseACL acl) {
```
wangmengyan95: Since this is a public API, do we need to add null checking here?

Parse/src/test/java/com/parse/ParsePushStateTest.java
```diff
@@ -56,6 +63,38 @@ public void testDefaultsWithData() throws Exception {
...
+    Set channelSetCopy = copy.channelSet();
+    assertNotSame(channelSet, channelSetCopy);
+    assertTrue(channelSetCopy.size() == 2 && channelSetCopy.contains("one"));
+    JSONObject dataCopy = copy.data();
+    assertNotSame(data, dataCopy);
+    assertEquals("bar", dataCopy.get("foo"));
+    assertTrue(copy.pushToAndroid());
+    assertFalse(copy.pushToIOS());
+    ParseQuery.State<ParseInstallation> queryStateCopy = copy.queryState();
+    assertNotSame(queryState, queryStateCopy);
```
wangmengyan95: :+1:

wangmengyan95:
one question, others LGTM!

Parse/src/main/java/com/parse/ParseACL.java
grantland: It implicitly NPE on the next line, which I think is fine since Java APIs work similarly: https://android.googlesource.com/platform/libcore/+/refs/heads/master/luni/src/main/java/java/util/HashMap.java#194

PR #312, opened 2015-12-19
Fix ParseUser#setState(state) unnecessarily creating authData

wangmengyan95:
Fixing a bug that was introduced in #304 that caused `authData` to be created on `ParseUser#setState(state)` with a state that did not contain `authData` in the first place.

### DIFF
--- a/Parse/src/test/java/com/parse/ParseUserTest.java
+++ b/Parse/src/test/java/com/parse/ParseUserTest.java
@@ -1428,7 +1428,36 @@ public class ParseUserTest {
     // Make sure we keep the authData
     assertEquals(1, user.getAuthData().size());
     assertEquals(authData, user.getAuthData().get(authType));
-    // Make sure old state is deleted
+    // Make sure old state is replaced
+    assertFalse(user.has("oldKey"));
+    // Make sure new state is set
+    assertEquals("testAgain", user.getObjectId());
+    assertEquals("valueAgain", user.get("key"));
+  }
+
+  @Test
+  public void testSetStateDoesNotAddNonExistentAuthData() throws Exception {
+    // Set user initial state
+    ParseUser.State userState = new ParseUser.State.Builder()
+        .objectId("test")
+        .put("oldKey", "oldValue")
+        .put("key", "value")
+        .build();
+    ParseUser user = ParseObject.from(userState);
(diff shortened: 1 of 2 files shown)


Parse/src/test/java/com/parse/ParseUserTest.java
```diff
@@ -1435,6 +1435,35 @@ public void testSetCurrentUserStateWithoutAuthData() throws Exception {
     assertEquals("valueAgain", user.get("key"));
   }
 
+  @Test
+  public void testSetNoAuthDataCurrentUserStateWithoutAuthData() throws Exception {
```
grantland: `testSetStateDoesNotAddNonExistentAuthData`

Parse/src/test/java/com/parse/ParseUserTest.java
```diff
@@ -1435,6 +1435,35 @@ public void testSetCurrentUserStateWithoutAuthData() throws Exception {
...
+    user.setIsCurrentUser(true);
+    // Build new state
+    ParseUser.State newUserState = new ParseUser.State.Builder()
+        .objectId("testAgain")
+        .put("key", "valueAgain")
+        .build();
+
+    user.setState(newUserState);
+
+    // Make sure we do not add none existed authData
```
grantland: `Make sure we do not add authData when it did not exist before`

Parse/src/test/java/com/parse/ParseUserTest.java
```diff
@@ -1435,6 +1435,35 @@ public void testSetCurrentUserStateWithoutAuthData() throws Exception {
...
+    assertFalse(user.keySet().contains("authData"));
+    assertEquals(1, user.keySet().size());
+    assertEquals(0, user.getAuthData().size());
+    // Make sure old state is deleted
```
grantland: `Make sure old state was replaced`

grantland:
Some nits and can we rename the PR and commit to say 

```
Fix ParseUser#setState(state) unnecessarily creating authData

Fixing a bug that was introduced in #304 that caused `authData` to be created on `ParseUser#setState(state)` with a state that did not contain `authData` in the first place.
```

so that we have a bit more context as to what this PR is about?

grantland:
LGTM

PR #317, opened 2015-12-22
Simplify authenticating w/ third party & automatic users

grantland:
Authenticating with third party auth providers while automatic
users were enabled and no current user existed would cause an
automatic user to be created just to be resolved.

Instead, don't create an automatic user but just authenticate
normally.


nlutsenko:
Beautiful!

PR #330, opened 2016-01-04
Add configurable serverURL to Parse.Configuration.

richardjrossiii:
Moves `server` from ParseObject to ParseRestCommand, and allows it to be
changed via `Builder.server("myUrl")`.

### DIFF
--- a/Parse/src/main/java/com/parse/Parse.java
+++ b/Parse/src/main/java/com/parse/Parse.java
@@ -22,6 +22,8 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.RandomAccessFile;
 import java.io.UnsupportedEncodingException;
+import java.net.MalformedURLException;
+import java.net.URL;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -48,6 +50,7 @@ public class Parse {
       private Context context;
       private String applicationId;
       private String clientKey;
+      private String server = "https://api.parse.com/1";
       private boolean localDataStoreEnabled;
       private List<ParseNetworkInterceptor> interceptors;
 
@@ -129,6 +132,20 @@ public class Parse {
       }
 
       /**
+       * Set the server URL to be used by Parse.
+       *
+       * This method is only required if you intend to use a different API server than the one at
+       * api.parse.com.
+       *
+       * @param server The server URL to set.
+       * @return The same builder, for easy chaining.
+       */
+      public Builder server(String server) {
+        this.server = server;
+        return this;
+      }
+
+      /**
        * Add a {@link ParseNetworkInterceptor}.
        *
        * @param interceptor The interceptor to add.
@@ -182,6 +199,7 @@ public class Parse {
     /* package for tests */ final Context context;
     /* package for tests */ final String applicationId;
     /* package for tests */ final String clientKey;
+    /* package for tests */ final String server;
     /* package for tests */ final boolean localDataStoreEnabled;
     /* package for tests */ final List<ParseNetworkInterceptor> interceptors;
 
@@ -189,6 +207,7 @@ public class Parse {
       this.context = builder.context;
       this.applicationId = builder.applicationId;
       this.clientKey = builder.clientKey;
+      this.server = builder.server;
       this.localDataStoreEnabled = builder.localDataStoreEnabled;
       this.interceptors = builder.interceptors != null ?
         Collections.unmodifiableList(new ArrayList<>(builder.interceptors)) :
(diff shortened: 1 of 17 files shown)


Parse/src/main/java/com/parse/ParseRESTCommand.java
```diff
@@ -44,6 +44,8 @@
   private static final String HEADER_MASTER_KEY = "X-Parse-Master-Key";
   private static final String PARAMETER_METHOD_OVERRIDE = "_method";
 
+  /* package */ static String server = null;
```
nlutsenko: Looks like it's unused?

Parse/src/main/java/com/parse/ParseRESTObjectBatchCommand.java
```diff
@@ -8,6 +8,8 @@
  */
 package com.parse;
 
+import android.net.Uri;
```
grantland: let's use `java.net.URL` instead

Parse/src/main/java/com/parse/ParseRESTCommand.java
```diff
@@ -189,7 +191,7 @@ public static ParseRESTCommand fromJSONObject(JSONObject jsonObject) {
   private static String createUrl(String httpPath) {
     // We send all parameters for GET/HEAD/DELETE requests in a post body,
     // so no need to worry about query parameters here.
-    return String.format("%s/1/%s", ParseObject.server, httpPath);
+    return String.format("%s/%s", server, httpPath);
```
nlutsenko: Can we use url building here, to avoid problems with `/` at the end of the server.

richardjrossiii:
Updated PR. Take another look :smile:

Parse/src/main/java/com/parse/ParseRESTObjectBatchCommand.java
```diff
@@ -63,7 +63,7 @@
       for (ParseRESTObjectCommand command : commands) {
         JSONObject requestParameters = new JSONObject();
         requestParameters.put("method", command.method.toString());
-        requestParameters.put("path", String.format("/1/%s", command.httpPath));
+        requestParameters.put("path", String.format("%s/%s", server.getPath(), command.httpPath));
```
richardjrossiii: Trailing slash issue would still appear in batch encoding, I don't have a great solution for that however, ATM.

nlutsenko: Is there an ability to ammend path componenets to path when building a URI?

grantland: How is this prevented in iOS? Just different internal URL logic?

Parse/src/test/java/com/parse/NetworkObjectControllerTest.java
```diff
@@ -36,6 +40,20 @@
 @Config(constants = BuildConfig.class, sdk = 21)
 public class NetworkObjectControllerTest {
 
+  @Before
+  public void setUp() {
```
grantland: I think you can just add `throws MalformedURLException` to the method signature instead.

This should also enforce that the test fails if there's an issue.

Parse/src/main/java/com/parse/ParseRESTObjectBatchCommand.java
richardjrossiii: @grantland yep, simply via different url logic.

We could solve this by doing `new URL(server, command.httpPath).getPath()`, though that potentially creates a lot of temporary objects in the interim.

grantland: I don't feel strongly either way

grantland: We are doing this for ParseRESTCommand, though

Parse/src/main/java/com/parse/Parse.java
```diff
@@ -129,6 +132,20 @@ public Builder clientKey(String clientKey) {
...
+       * This method is only required if you intend to use a different API server than the one at
+       * api.parse.com.
+       *
+       * @param server The server URL to set.
+       * @return The same builder, for easy chaining.
+       */
+      public Builder server(String server) {
+        this.server = server;
+        return this;
+      }
```
grantland: Mind updating the tests to cover this?

richardjrossiii: Done.

grantland:
LGTM, squash and feel free to merge!

Issue #323, opened 2015-12-28
ParseFile.cancel() question


ppilip:
Hello,

I'm having issue with the `ParseFile.cancel()` method call. The call does not stop file upload. I've checked in debug mode, the method is being called, Task objects are removed from the list, but the file upload still continues in the background.

I'm using Android Studio with latest Parse SDK. My build file has:

`compile 'com.parse:parse-android:1.11.0'`
`compile files('libs/ParseCrashReporting-1.10.2.jar')`
`compile files('libs/ParseFacebookUtilsV4-1.10.3.jar')`
`compile files('libs/ParseTwitterUtils-1.10.3.jar')`

wangmengyan95:
Hi @ppilip, we do not support stop an HTTP request in the middle. This means once our SDK starts an HTTP request to upload a ParseFile, there is no way to cancel it during the process. You can cacnel a ParseFile upload when it is in the task queue.

ppilip:
Hi @wangmengyan95 ok, thanks for clarifying what this method call does.

The text in the JavaDoc for that method should be changed then (or at least made a bit more clear). Perhaps it would be nice to to add something like "but it does not cancel any data downloading/uploading that already has been started" or something.

Current text is:
`Cancels the current network request and callbacks whether it's uploading or fetching data from
 the server.`

wangmengyan95:
Hi @ppilip, thanks for advice. I will update our doc.

PR #339, opened 2016-01-12
Update comment for ParseFile.cancel()

wangmengyan95:
Update comment for `ParseFile.cancel()`. Emphasize the network request will not be canceled. For #323.


grantland:
LGTM

Issue #352, opened 2016-01-18
no /1/ in api.parse.com


AndroidGecko:
I have a very strange issue.
When running a very simple project from scratch with :
```
    Parse.enableLocalDatastore(this);
    Parse.initialize(this);
    ParseObject testObject = new ParseObject("TestObject");
    testObject.put("foo", "bar");
    testObject.saveInBackground();
```

everything works fine
the request is made to 
https://api.parse.com/1/classes/TestObject
and everything works fine

However if I try to integrate my existing app and place same code inside custom application (+keys in manifest) the api call fails with 404. Reason being that it pokes
https://api.parse.com/classes/TestObject

notice lack of /1/ 

If I manually catch above request in proxy and re-execute it with same data, it passes.

Any ideas why the /1/ is not being added?

wangmengyan95:
Hi @AndroidGecko, thanks for your information. We do have a bug in the master branch about the lack of /1 issue you mention. If you are using the snapshot version, then it is fine. But if you are using the formal release verisn, could you tell us your version number and how do you change the manifest, thanks!

wangmengyan95:
#356

AndroidGecko:
Thanks @wangmengyan95 
Got it working with snapshot and
```
Parse.initialize(
        new Parse.Configuration.Builder(this).server("https://api.parse.com/1/").build());
```
also seems to be working on 1.12.0
after running 
```
gradle build --refresh-dependencies
```
since I was using 1.+ earlier

Thanks! 
btw - that is one great sdk!

wangmengyan95:
@AndroidGecko thanks. The fix has been merged into master and it will be shipped in next release.

PR #356, opened 2016-01-19
Fix default parse server url

wangmengyan95:
We need to add additional `/` at the end of our default server URL address. Otherwise, when we use `new URL(server, httpPath)` to build the URL [here](https://github.com/ParsePlatform/Parse-SDK-Android/blob/master/Parse/src/main/java/com/parse/ParseRESTCommand.java#L194), the `/1` will be ignored.
cc @richardjrossiii

### DIFF
--- a/Parse/src/main/java/com/parse/Parse.java
+++ b/Parse/src/main/java/com/parse/Parse.java
@@ -50,7 +50,7 @@ public class Parse {
       private Context context;
       private String applicationId;
       private String clientKey;
-      private String server = "https://api.parse.com/1";
+      private String server = "https://api.parse.com/1/";
       private boolean localDataStoreEnabled;
       private List<ParseNetworkInterceptor> interceptors;
 
--- a/Parse/src/test/java/com/parse/ParseRESTCommandTest.java
+++ b/Parse/src/test/java/com/parse/ParseRESTCommandTest.java
@@ -80,6 +80,17 @@ public class ParseRESTCommandTest {
     ParseRESTCommand.server = null;
   }
 
+
+  @Test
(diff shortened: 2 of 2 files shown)


grantland:
Fixes #352

grantland:
@richardjrossiii Does this comply with the batch issue raised in #330?

richardjrossiii:
@grantland Yes, URLs are constructed identically in batch vs normal creation.

wangmengyan95:
@grantland @richardjrossiii any other opinions about this fix?

grantland:
As long as everything works, I think it's fine. The original concern with trailing slashes was that it would cause a double slash which would mean root:

```
https://api.parse.com/1//blah -> https://api.parse.com/blah
```

Are you sure this scenario won't be an issue?

wangmengyan95:
@grantland, just tested
```
URL root = new URL("https://api.parse.com/");
String path = "/123";
URL url = new URL(server, path);
```
The finial url is `https://api.parse.com/123`, no double slash.

grantland:
Awesome, LGTM

richardjrossiii:
Hey guys, we may want to re-investigate this. I just tested (locally, and [here](http://ideone.com/Pz9wXf)), it definitely converts a double slash into a path for root.

wangmengyan95:
Yep, I do a more through test [here](http://ideone.com/SJedxR).

grantland:
Discussed in person. The original concern comes from the fact that the paths in iOS/.NET are defined with a leading `/`, so passing in a server path with a trailing `/` will yield `//` which changes the URL to root. We don't define paths in Android with leading `/`, so this is not an issue.

PR #361, opened 2016-01-22
Make eventuallyQueue initialization not rely on ParsePlugins

wangmengyan95:
When we initialize Parse, we initialize eventuallyQueue async. Thus, sometimes in our integration test, when we actually initialize eventuallyQueue, the `ParsePlugins` is not available. This will cause a NPE. This change make eventuallyQueue initialization not rely on `ParsePlugins`, we pass a context object to the initialization function instead of getting it from `ParsePlugins`.

### DIFF
--- a/Parse/src/main/java/com/parse/Parse.java
+++ b/Parse/src/main/java/com/parse/Parse.java
@@ -30,6 +30,7 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.concurrent.Callable;
 
 import bolts.Continuation;
 import bolts.Task;
@@ -402,13 +403,14 @@ public class Parse {
     // Make sure the data on disk for Parse is for the current
     // application.
     checkCacheApplicationId();
-    new Thread("Parse.initialize Disk Check & Starting Command Cache") {
+    final Context context = configuration.context;
+    Task.callInBackground(new Callable<Void>() {
       @Override
-      public void run() {
-        // Trigger the command cache to flush its contents.
-        getEventuallyQueue();
+      public Void call() throws Exception {
+        getEventuallyQueue(context);
+        return null;
       }
-    }.start();
+    });
 
     ParseFieldOperations.registerDefaultDecoders();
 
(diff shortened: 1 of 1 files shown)


Parse/src/main/java/com/parse/Parse.java
```diff
@@ -600,6 +601,10 @@ static void checkCacheApplicationId() {
    */
   /* package */ static ParseEventuallyQueue getEventuallyQueue() {
     Context context = ParsePlugins.Android.get().applicationContext();
+    return getEventuallyQueue(context);
+  }
+
+  /* package */ static ParseEventuallyQueue getEventuallyQueue(Context context) {
```
grantland: We can make this `private` for now

Parse/src/main/java/com/parse/Parse.java
```diff
@@ -369,7 +370,7 @@ public static void initialize(Context context, String applicationId, String clie
     );
   }
 
-  public static void initialize(Configuration configuration) {
+  public static void initialize(final Configuration configuration) {
```
grantland: let's just hold a `final` reference of `Context` instead of the whole `configuration`

wangmengyan95: kk

Parse/src/main/java/com/parse/Parse.java
wangmengyan95: :+1:

Parse/src/main/java/com/parse/Parse.java
```diff
@@ -402,13 +403,13 @@ public static void initialize(Configuration configuration) {
     // Make sure the data on disk for Parse is for the current
     // application.
     checkCacheApplicationId();
-    new Thread("Parse.initialize Disk Check & Starting Command Cache") {
+    Task.callInBackground(new Callable<Void>() {
```
grantland: can we `Task.call(() -> { ... }, Task.BACKGROUND_EXECUTOR)` instead?

grantland:
LGTM

PR #372, opened 2016-01-26
Use push_id instead of push_hash

grantland:
`push_id` better uniquely identifies a push, as opposed to `push_hash`
which will determine two different pushes with the same content as the
same.


Parse/src/main/java/com/parse/ParseRESTAnalyticsCommand.java
```diff
@@ -29,12 +29,12 @@
   private static final String PATH = "events/%s";
 
   private static final String KEY_AT = "at";
-  private static final String KEY_PUSH_HASH = "push_hash";
+  private static final String KEY_PUSH_ID = "push_id";
```
bnham: The key is actually `parsePushId`. The reason why we chose that is because that key was already being used to send the push id to the iOS SDK.

PR #414, opened 2016-03-10
Fix #343: Add proguard consumer file

### LINKED ISSUE #343: Add consumer proguard rules
https://github.com/ParsePlatform/Parse-SDK-Android/blob/master/ParseStarterProject/proguard-rules.pro should be included in Parse project as [a consumer proguard file](https://github.com/googlesamples/android-testing-templates/blob/master/AndroidTestingBlueprint/module-android-library/proguard-consumer-rules.pro#L1). A benefit is that projects that depend on Parse wouldn't need to bundle a dedicated proguard file for it. So that would save lots of time and effort. [Latest Play services has done the same thing](https://developers.google.com/android/guides/setup).

Wdyt?

Allsimon:
The Android Gradle plugin allows to define ProGuard rules which get embedded in the AAR. These ProGuard rules are automatically applied when a consumer app sets minifyEnabled to true.

### DIFF
--- a/Parse/build.gradle
+++ b/Parse/build.gradle
@@ -25,6 +25,7 @@ android {
         targetSdkVersion rootProject.ext.targetSdkVersion
         versionCode 1
         versionName project.version
+        consumerProguardFiles 'release-proguard.pro'
     }
 
     lintOptions {
--- a/Parse/release-proguard.pro
+++ b/Parse/release-proguard.pro
@@ -16,3 +16,14 @@
 #   public *;
 #}
 
+# Keep source file names, line numbers, and Parse class/method names for easier debugging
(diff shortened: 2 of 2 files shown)


ParseStarterProject/proguard-rules.pro
```diff
@@ -14,16 +14,4 @@
...
--keepnames class com.parse.** { *; }
-
-# Required for Parse
--keepattributes *Annotation*
--keepattributes Signature
--dontwarn android.net.SSLCertificateSocketFactory
--dontwarn android.app.Notification
--dontwarn com.squareup.**
--dontwarn okio.**
+#}
```
grantland: nit: newline at EOF

grantland:
Thanks for the contribution!

PR #420, opened 2016-03-11
Fix the bug that wrongly use the GCM when GSF is not fully available.

elvishew:
When I am using Parse push notification, I found I always could not receive any push notification if I am using Huawei phones, and finally I found out that the Huawei phone has google service framework, so Parse would choose GCM as push type, but unfortunately, there is no Play Store in the phone, and somehow when Parse try to register the GCM push, it always got GSF_PACKAGE_NOT_AVAILABLE error, so the push notification won't work at all, but if I install Play Store on the phone, the GCM works.
I don't know whether this is a bug of Huawei's OS, or a bug of Parse, I just fix it. If we can not find the GCM register service, we regard the phone as GCM register unavailable.
With this fix, even there is no Play Store installed in Huawei phones, the push can work in a PPNS mode.

### DIFF
--- a/Parse/src/main/java/com/parse/ManifestInfo.java
+++ b/Parse/src/main/java/com/parse/ManifestInfo.java
@@ -198,6 +198,7 @@ import java.util.List;
     synchronized (lock) {
       if (pushType == null) {
         boolean isGooglePlayServicesAvailable = isGooglePlayServicesAvailable();
+        boolean isGcmRegisterServiceAvailable = isGcmRegisterServiceAvailable();
         boolean isPPNSAvailable = PPNSUtil.isPPNSAvailable();
         boolean hasAnyGcmSpecificDeclaration = hasAnyGcmSpecificDeclaration();
         ManifestCheckResult gcmSupportLevel = gcmSupportLevel();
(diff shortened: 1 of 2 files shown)


Parse/src/main/java/com/parse/ManifestInfo.java
```diff
@@ -466,6 +468,13 @@ private static boolean isGooglePlayServicesAvailable() {
     return Build.VERSION.SDK_INT >= 8 && getPackageInfo("com.google.android.gsf") != null;
   }
 
+  private static boolean isGcmRegisterServiceAvailable() {
+    Intent intent = new Intent(GcmRegistrar.REGISTER_ACTION);
+    intent.setPackage("com.google.android.gsf");
+    List<ResolveInfo> services = getContext().getPackageManager().queryIntentServices(intent, 0);
+    return services != null && services.size() > 0;
+  }
```
grantland: let's move this method into `GcmRegistrar`

grantland:
With this change, we shouldn't need `isGooglePlayServicesAvailable()`, right?

grantland:
I'm not too familiar with this code, @bnham would you have any insight into this?

Parse/src/main/java/com/parse/ManifestInfo.java
```diff
@@ -462,10 +461,6 @@ private static boolean hasAnyGcmSpecificDeclaration() {
     return false;
   }
 
-  private static boolean isGooglePlayServicesAvailable() {
-    return Build.VERSION.SDK_INT >= 8 && getPackageInfo("com.google.android.gsf") != null;
```
bnham: Hah, this reminds me that this code was originally written to work on Donut phones.

bnham: Can we simplify this PR by deleting these two lines here and replacing it with the code you wrote for `isGcmRegisterServiceAvailable`?

bnham:
This looks good, but let's try to minimize the amount of code changed. Can you change your diff so that it just deletes the body of `isGooglePlayServicesAvailable` and replaces it with the body of `isGcmRegisterServiceAvailable`? Then you wouldn't need to change all of those other lines of code.

facebook-github-bot-7:
@elvishew updated the pull request.

facebook-github-bot-4:
@elvishew updated the pull request.

rogerhu:
Do we still need this PR?

Jawnnypoo:
This PR is very out of date, if OP wants to redo it, feel free. Going to close this one though

PR #435, opened 2016-04-06
Update for OkHttp3

rogerhu:
Fixes for OkHttp3

### DIFF
--- a/Parse/src/main/java/com/parse/ParseOkHttpClient.java
+++ b/Parse/src/main/java/com/parse/ParseOkHttpClient.java
@@ -8,22 +8,12 @@
  */
 package com.parse;
 
-import android.net.SSLCertificateSocketFactory;
 import android.net.SSLSessionCache;
 
 import com.parse.http.ParseHttpBody;
 import com.parse.http.ParseHttpRequest;
 import com.parse.http.ParseHttpResponse;
 import com.parse.http.ParseNetworkInterceptor;
-import com.squareup.okhttp.Call;
-import com.squareup.okhttp.Headers;
-import com.squareup.okhttp.Interceptor;
-import com.squareup.okhttp.MediaType;
-import com.squareup.okhttp.OkHttpClient;
-import com.squareup.okhttp.Request;
-import com.squareup.okhttp.RequestBody;
-import com.squareup.okhttp.Response;
-import com.squareup.okhttp.ResponseBody;
 
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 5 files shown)


Parse/src/main/java/com/parse/ParseOkHttpClient.java
```diff
@@ -48,17 +47,18 @@
...
 
     // Don't handle redirects. We copy the setting from AndroidHttpClient.
     // For detail, check https://quip.com/Px8jAxnaun2r
-    okHttpClient.setFollowRedirects(false);
+    builder.followRedirects(false);
 
-    okHttpClient.setSslSocketFactory(SSLCertificateSocketFactory.getDefault(
-        socketOperationTimeout, sslSessionCache));
+    // Broken until OkHttp3 can expose trust managers -- https://github.com/square/okhttp/commit/85f74e2004eaf0d4ff339e8644df3d8e716361e5
+    // builder.sslSocketFactory(SSLCertificateSocketFactory.getDefault(socketOperationTimeout, sslSessionCache));
```
rogerhu: is this still necessary given the connect timeout is already setup?

grantland: I'm not entirely sure, but I think this was a relic from our original use of `AndroidHttpClient.newInstance(userAgent, sessionCache)`.

If you are sure that OkHttp client doesn't need a SSL cache, I'd be happy to accept this!

rogerhu: It doesn't need one except helps obviously with saving SSL handshakes when reconnecting.

In order to use it, I think we have to wait for OkHttp v3.3.0 to drop.  See conversation here:

https://github.com/square/okhttp/issues/2323#issuecomment-207231855

rogerhu: I don't have enough experience to know whether SSL caching functionality is a blocker... anyone on the Parse team could provide some insight?

setheclark:
I don't see a change in the ParseHttpClient class to update the OKHTTPCLIENT_PATH field for OkHttp3.  I think this means the classpath lookup to make sure OkHttp3 in available would fail.

rogerhu:
Thanks for checking.  Updated.

facebook-github-bot-6:
@rogerhu updated the pull request.

Parse/src/main/java/com/parse/ParseOkHttpClient.java
grantland: I think it would be safe to use the default, so we can just remove this bit.

grantland:
Mind removing that commented out bit? After that it LGTM!

rogerhu:
Remove commented section

facebook-github-bot-8:
@rogerhu updated the pull request.

grantland:
Merging. Thanks for the contribution!

rogerhu:
Thanks!  Any chance to cut a release?

grantland:
Currently waiting on one PR, but will make a cut next week if it isn't in by then.

PR #436, opened 2016-04-07
For open source Parse with /parse mounts, this fix helps avoid the problem of remembering to add the trailing slash.

rogerhu:


grantland:
@richardjrossiii Could you take a look at this?

Parse/src/test/java/com/parse/ParseClientConfigurationTest.java
```diff
@@ -30,14 +30,14 @@ public void testBuilder() {
     Parse.Configuration.Builder builder = new Parse.Configuration.Builder(null);
     builder.applicationId("foo");
     builder.clientKey("bar");
-    builder.server("some.server");
+    builder.server("http://myserver.com/parse");
```
richardjrossiii: Can we perhaps duplicate this test case, instead of just modifying it in-place? One test for each branch?

richardjrossiii:
I'd prefer we use `String.endsWith()` instead of `charAt()`, even if the performance of the latter is better. Seems more clear to me.

rogerhu:
Fixed...can you take a look again?  I took out the server() on the original test and broke them out to separate tests to check both cases.

richardjrossiii:
Nice! Merging now.

rogerhu:
Thanks!

posativ:
Example traceback for GitHub's search.

```
com.parse.ParseRequest$ParseRequestException: bad json response
	at com.parse.ParseRequest.newTemporaryException(ParseRequest.java:290)
	at com.parse.ParseRESTCommand.onResponseAsync(ParseRESTCommand.java:308)
	at com.parse.ParseRequest$3.then(ParseRequest.java:137)
	at com.parse.ParseRequest$3.then(ParseRequest.java:133)
	at bolts.Task$15.run(Task.java:917)
	at bolts.BoltsExecutors$ImmediateExecutor.execute(BoltsExecutors.java:105)
	at bolts.Task.completeAfterTask(Task.java:908)
	at bolts.Task.continueWithTask(Task.java:715)
	at bolts.Task.continueWithTask(Task.java:726)
	at bolts.Task$13.then(Task.java:818)
	at bolts.Task$13.then(Task.java:806)
	at bolts.Task$15.run(Task.java:917)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587)
	at java.lang.Thread.run(Thread.java:818)
Caused by: org.json.JSONException: Value Cannot of type java.lang.String cannot be converted to JSONObject
	at org.json.JSON.typeMismatch(JSON.java:111)
	at org.json.JSONObject.<init>(JSONObject.java:160)
	at org.json.JSONObject.<init>(JSONObject.java:173)
	at com.parse.ParseRESTCommand.onResponseAsync(ParseRESTCommand.java:298)
	... 13 more
```

rogerhu:
Could you explain what this has to do with this PR?

posativ:
You run into this error when you're not using the master and you cannot find it easily.

rogerhu:
Could you provide a code sample?  How is this PR related to your issue?

posativ:
If you use the latest release of Parse SDK for Android with a custom Parse server hosted on e.g. foo.example.org:4043/foobar and forget to add a slash, you get this nonsense error message.

The PR resolves this.

If it still doesn't make sense, remove my comments.

rogerhu:
Yes we should release a new version right? :)

posativ:
Yup, might save a few hours questionig the self-hosted server ;)

rogerhu:
Can someone make this happen?

PR #464, opened 2016-05-26
Don't override default SSL Socket Factory

egk35:
This SSL Socket Factory is causing a javax.net.ssl.SSLPeerUnverifiedException exception when integrating an app with both Parse and Account Kit (https://developers.facebook.com/products/account-kit). Is there a reason why we're using the SSLCertificateSocketFactory? I think the default one should be fine.


facebook-github-bot-3:
Thank you for your pull request and welcome to our community. We require contributors to sign our Contributor License Agreement, and we don't seem to have you on file. In order for us to review and merge your code, please sign up at https://code.facebook.com/cla - and if you have received this in error or have any questions, please drop us a line at cla@fb.com. Thanks!

grantland:
I don't believe there's an explicit reason why we have this, so I'll merge it in for now.

PR #465, opened 2016-05-26
Don't override default SSL Socket Factory

egk35:
This SSL Socket Factory is causing a javax.net.ssl.SSLPeerUnverifiedException exception when integrating an app with both Parse and Account Kit (https://developers.facebook.com/products/account-kit). Is there a reason why we're using the SSLCertificateSocketFactory? I think the default one should be fine.


grantland:
I don't believe there's an explicit reason why we have this, so I'll merge it in for now.

PR #468, opened 2016-05-27
Update for OkHttp3

rogerhu:
Changes to https://github.com/square/okhttp/commit/bab8943f2b5c8bc44a64067ada556aa62fcec3a7 has caused network interceptors to return headers all in lower case because a change in which the toMultimap() function.   This change appears to have been added to avoid duplicate headers being sent (https://github.com/square/okhttp/blob/bab8943f2b5c8bc44a64067ada556aa62fcec3a7/okhttp-tests/src/test/java/okhttp3/internal/http/HeadersTest.java#L298-L306).

The downside is that Parse/OkHttp network interceptors exhibit this behavior, but normal interceptors do not.   We call the toMultimap() routine that causes headers to be converted to lowercase on downstream requests (see https://github.com/ParsePlatform/Parse-SDK-Android/blob/master/Parse/src/main/java/com/parse/ParseOkHttpClient.java#L187-L189)

I updated the tests to rely on lowercase headers to avoid the test falures.  I think the reasoning behind needing toMultiMap() is in case someone needs to rewrite the headers and you want to make sure you are overwriting the right one.

### DIFF
--- a/Parse/build.gradle
+++ b/Parse/build.gradle
@@ -42,13 +42,13 @@ android {
 dependencies {
     compile 'com.parse.bolts:bolts-tasks:1.4.0'
 
-    provided 'com.squareup.okhttp3:okhttp:3.2.0'
+    provided 'com.squareup.okhttp3:okhttp:3.3.1'
     provided 'com.facebook.stetho:stetho:1.3.0'
 
     testCompile 'org.robolectric:robolectric:3.0'
     testCompile 'org.skyscreamer:jsonassert:1.2.3'
     testCompile 'org.mockito:mockito-core:1.10.19'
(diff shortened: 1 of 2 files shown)


facebook-github-bot-3:
By analyzing the blame information on this pull request, we identified @grantland, @Allsimon and @wangmengyan95 to be potential reviewers.

grantland:
Looks like there are test failures

facebook-github-bot-5:
@rogerhu updated the pull request.

facebook-github-bot-0:
@rogerhu updated the pull request.

facebook-github-bot-9:
@rogerhu updated the pull request.

grantland:
Looks great, thanks for the fix!

Issue #433, opened 2016-04-01
Got "com.parse.ParseRequest$ParseRequestException" when trying to update object.


Xavius:
Hi, all!
Every time I tried to update object, got such error without any description message. On the server side at the same time I got this trace:

> Uncaught internal server error. { [MongoError: exception: Mod on _id not allowed]
>   name: 'MongoError',
>   message: 'exception: Mod on _id not allowed',
>   errmsg: 'exception: Mod on _id not allowed',
>   code: 10148,
>   ok: 0 } MongoError: exception: Mod on _id not allowed
>     at Function.MongoError.create (/usr/local/lib/node_modules/parse-server/node_modules/mongodb/node_modules/mongodb-core/lib/error.js:31:11)
>     at commandCallback (/usr/local/lib/node_modules/parse-server/node_modules/mongodb/node_modules/mongodb-core/lib/topologies/server.js:1136:66)
>     at Callbacks.emit (/usr/local/lib/node_modules/parse-server/node_modules/mongodb/node_modules/mongodb-core/lib/topologies/server.js:116:3)
>     at null.messageHandler (/usr/local/lib/node_modules/parse-server/node_modules/mongodb/node_modules/mongodb-core/lib/topologies/server.js:291:23)
>     at Socket.<anonymous> (/usr/local/lib/node_modules/parse-server/node_modules/mongodb/node_modules/mongodb-core/lib/connection/connection.js:285:22)
>     at emitOne (events.js:90:13)
>     at Socket.emit (events.js:182:7)
>     at readableAddChunk (_stream_readable.js:153:18)
>     at Socket.Readable.push (_stream_readable.js:111:10)
>     at TCP.onread (net.js:529:20)

Has anyone faced a such issue?

Xavius:
Ok, so I've found an issue.
According to documentation "PUT" request should looks like:
`curl -X PUT \
  -H "X-Parse-Application-Id: APP_ID" \
  -H "Content-Type: application/json" \
  -d '{"score":73453}' \
  https://my-parse-server:1337/parse/classes/MyClass/MyObjectId`
but Parse Android SDK send also add "objectId" to request body:
`  -d '{"objectId":"MyObjectId","score":73453}' \`

So, parse-server tries to update object Id, too. But MongoDB prevents this.

shiftinteractive:
Any word on a fix for this?

TMwL:
I'm in the same situation...
Is there any known fix or a workaround for this?

[...]

PR #536, opened 2016-10-23
Do not add the object ID to the JSON representation

alsutton:
The Object ID is added in ParseRESTObjectCommand as part of the URL,
including it in the JSON representation causes the following error
on the server;

exception: Mod on _id not allowed

This fix addresses issue #433


alsutton:
Closing out to get it off my plate.

PR #542, opened 2016-11-08
fix bug in ParseOkHttpClient that can cause errors when saving ParseFile

zhoul-HS:
See #364


zhoul-HS:
Any chance of releasing 1.13.2 with this fix? i'm reluctant to use a whole module just for this :/ i'd rather get it off maven

rogerhu:
+1

StuStirling:
I can't use the latest SDK until this fixed. Any idea when a new maven release will be generated?

PR #548, opened 2016-11-24
Fix bug in ParseOkHttpClient

ly85206559:
ClassCastException happened while delete ParseObject

### DIFF
--- a/Parse/src/main/java/com/parse/ParseOkHttpClient.java
+++ b/Parse/src/main/java/com/parse/ParseOkHttpClient.java
@@ -117,7 +117,7 @@ import okio.Okio;
         okHttpRequestBuilder.get();
         break;
       case DELETE:
-        okHttpRequestBuilder.delete();
+        okHttpRequestBuilder.delete(null);
         break;
       case POST:
       case PUT:


grantland:
Could you provide more context on the bug you're fixing? Possibly a stack trace or an issue with more information?

ly85206559:
: java.lang.ClassCastException: okhttp3.RequestBody$2 cannot be cast to com.parse.ParseOkHttpClient$ParseOkHttpRequestBody
at com.parse.ParseOkHttpClient.getParseHttpRequest(ParseOkHttpClient.java:193) 
at com.parse.ParseOkHttpClient.access$000(ParseOkHttpClient.java:39)
at com.parse.ParseOkHttpClient$1.intercept(ParseOkHttpClient.java:214)

and this is Okhttp source code

    public Builder delete(RequestBody body) {
      return method("DELETE", body);
    }

    public Builder delete() {
      return delete(RequestBody.create(null, new byte[0]));
    }

so if invoke delete(),RequestBody is created by okhttp, so can't cast to ParseOkHttpRequestBody.

rogerhu:
could use a test but good to go

rogerhu:
On second thought not sure what this fix does since ParseFile doesn't expose a delete...how does this help?

rogerhu:
Going to close until we have more data on what it's supposed to fix.

PR #554, opened 2016-12-06
Remove Apache dependency to allow for higher Android target SDK.

Jawnnypoo:
This has a chance of resolving https://github.com/ParsePlatform/Parse-SDK-Android/issues/551 though I cannot confirm as the message from Google is vague and does not point out what the issue is specifically. Reguardless, OkHttp 3.X is tried and true, and it seems like it would be a great replacement since the work is already done for it. 

This is also a more complete PR of the work done in https://github.com/ParsePlatform/Parse-SDK-Android/pull/418 without depending on beta versions of the SDK tools, plugin, etc.

The change in SDK version required a bump to Robolectric which is why a few of the tests changed as well. 

I also added a nice Gradle plugin that allows you to check for the latest dependencies. I attempted to bring all dependencies up to date, but ran into complications with the latest versions of Mockito, so leaving a few of those for another PR.

Since this does drop an HTTP client implementation without deprecation, it may warrant a bump from 1.13.X to 1.14.X, but I leave that up to you. I just think if Google is truly blocking updates and app submissions based on Apache flaws (either coming from this library or otherwise) it is best to just leave it behind since it is deprecated in the SDK anyhow.


Jawnnypoo:
Sorry for the spammy notifications, should be good to go now. Just had some strange Travis issues.

Jawnnypoo:
Not really sure what I can do about the code coverage, this PR mostly just removes code, so I don't think it should be a concern

Jawnnypoo:
Any thoughts on this @grantland ?

Parse/build.gradle
```diff
@@ -39,16 +39,16 @@ android {
...
     compile 'com.parse.bolts:bolts-tasks:1.4.0'
+    compile "com.squareup.okhttp3:okhttp:$okhttpVersion"
+    provided 'com.facebook.stetho:stetho:1.4.1'
 
-    provided 'com.squareup.okhttp3:okhttp:3.3.1'
-    provided 'com.facebook.stetho:stetho:1.3.0'
-
-    testCompile 'org.robolectric:robolectric:3.0'
-    testCompile 'org.skyscreamer:jsonassert:1.2.3'
+    testCompile 'org.robolectric:robolectric:3.1.4'
```
rogerhu: bump to 3.3.1?

rogerhu:
+1 okhttp is now the underlying interface of HttpUrlConnection.  Can we  get this merged in?

rogerhu:
Issue is related to: https://github.com/ParsePlatform/Parse-SDK-Android/issues/231

Jawnnypoo:
Yeah I would love to see this moved forward. Supporting only one HttpClient should reduce the headache of supporting this SDK massively.

Jawnnypoo:
I think @grantland is one of the only devs on this project that has the authority to merge PRs on this repo, but I don't think he is too active on doing so.

Parse/src/main/java/com/parse/NotificationCompat.java
```diff
@@ -56,8 +56,11 @@
   static class NotificationCompatImplBase implements  NotificationCompatImpl {
     @Override
     public Notification build(Builder b) {
-      Notification result = (Notification) b.mNotification;
-      result.setLatestEventInfo(b.mContext, b.mContentTitle, b.mContentText, b.mContentIntent);
+      Notification result = b.mNotification;
+      NotificationCompat.Builder builder = new NotificationCompat.Builder(b.mContext);
+      builder.setContentTitle(b.mContentTitle);
+      builder.setContentText(b.mContentText);
+      builder.setContentIntent(b.mContentIntent);
```
rogerhu: is this going to work without running builder.build() and returning the result?

rogerhu:
The rest of the PR (minus the notification stuff) looks great, thanks for doing this.

Does anyone else have strong reservations/opinions about dropping the Apache client?  Given Android 2.x is 1.0% of the market share (https://developer.android.com/about/dashboards/index.html), it seems like we should do this.

We will bump to v1.14.XX once this happens.

Parse/src/main/java/com/parse/NotificationCompat.java
Jawnnypoo: Good catch, updated

Jawnnypoo:
@rogerhu This doesn't require dropping any support, we can continue to target 9 as the min version, using Okhttp exclusively does not change that number. 👍

Jawnnypoo:
Also, I fixed the Notification issue, so this should be good to go

rogerhu:
Thanks.  I'm not seeing the latest build pushed up to Maven so going to check with Parse guys to figure out how to handle... :)

flovilmart:
@grantland on that point, could you let us on Sonatype or open a Pr for distribute on tags?

rogerhu:
I suspect that the builds are going up on Sonatype fine (https://books.sonatype.com/nexus-book/reference/staging-repositories.html) but need to be released either manually or via [API](http://www.sonatype.org/nexus/2015/03/31/nexus-staging-plugin-automatic-releasepromotion-of-artifacts-to-maven-central-from-gradle/)

rogerhu:
v13.3.3 is released thanks to @flovilmart.  Apparently the release process is not quite automated on Travis yet but is being worked on.

This PR has been merged into v1.14.1-snapshot.  Would appreciate people check and see if there are any lingering issues.

rogerhu:
v1.14.0 is now out which should include these changes.

Jawnnypoo:
Awesome! Thanks!

Issue #122, opened 2015-09-10
ParseObject implements Parcelable


WanClem:
I think ParseObject needs to implement Serializable.In a scenario where a developer is building a long list of ParseObject items that are pulled from the internet and displayed on a list view,if serialized it could help mimick offline content on the occasion where the user has no connectivity.That is,the previously serialized list items could be deserialized and populated on the list view when there is no connectivity on the phone.I tried subclassing ParseObject to get this effect but the Problem is that Subclassing ParseObject is forcing me to use a Zero-Argument Constructor and a predefined className which leaves me with the question "What if I want to provide classNames dynamically without predefining with @ParseClassName annotation as done by ParseObject?." Please help.

grantland:
First off, `Parcelable` is more efficient than `Serializable` for passing objects through `Intent`s, so use that when possible. I'll rename this issue as such.

Secondly, we're planning on converting `ParseObject` to implement `Serializable` at some point, but it's a bit more complicated that you might think:

1. Do we parcel just server data or server data and dirty changes?

    Right now we're planning on passing both server data and dirty changes. Once `object.revert()` lands in #112, you'll be able to clear dirty changes and have access the server data.

2. What do we do if there are multiple operation sets in the object's queue like we're in the middle of saving an object?

    The best option might be to consider the previous operations as failed and roll forward the changes into one operation set. This will mean that the unparceled object might have some dirty changes that will be persisted on the server and additional calls to `save()` will execute them again. Doing so will be fine for idempotent operations, but might cause some issues with `increment(key)` and `decrement(key)`. We can issue a logcat warning when parceling an object with in progress operations to alert them of these possible inconsistencies.

3. What do we do if the object has been deleted?

    This seems like a super rare case and might be fine doing nothing. The only side-affect of doing nothing will mean that the unparceled object will incorrectly return `false` from `isDirty()`.

4. What do we do if the object is currently being deleted?

    Follow (2) and assume the `delete()` has failed and warn in logcat. Same repercussions as (3).

5. What do we do for in progress `saveEventually()` or `deleteEventually()`?

    Follow (2) and assume the eventually calls failed and warn in logcat. Same repercussions as (2) as well.

6. Is it ok that the unparceled object will be the same instance when LDS is enabled?

[...]

PR #557, opened 2016-12-06
Parcelable implementation for ParseFile

johncordeiro:
This is the implementation of Parcelable interface on ParseFile for remote and local files. It's very useful to pass complex objects through activities, specially when you are using Parse with delegated properties in Kotlin.

### DIFF
--- a/Parse/src/main/java/com/parse/ParseFile.java
+++ b/Parse/src/main/java/com/parse/ParseFile.java
@@ -69,7 +69,8 @@ public class ParseFile implements Parcelable {
     };
   }
 
-  /* package */ static class State implements Parcelable {
+  // For test
+  protected final static class State implements Parcelable {
 
     /* package */ static class Builder {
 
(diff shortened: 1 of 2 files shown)


johncordeiro:
Is there anyone maintaining this project?

rogerhu:
+1

rogerhu:
Can you simply extend the class instead of needing to put in on ParseFile?

rogerhu:
Does anyone else have any opinions about the need to do this?

natario1:
I don’t think that parceling a whole file with its byte array would be considered good practice. I also remember Parcel (or Bundle) throwing when its contents are too large.

In my opinion this should work only when the file is backed by an url, and if not (e.g. ParseFile not saved yet) throw an exception. That would be harmless and a good step in the direction of implementing Parcelable in the SDK.

rogerhu:
Yeah there is a Parcelable max size of 1 Mb (https://developer.android.com/reference/android/os/TransactionTooLargeException.html) 

Going to close unless there is a revision to it.

johncordeiro:
Yes, it seems to be reasonable, the code was updated.

rogerhu:
See https://github.com/ParsePlatform/Parse-SDK-Android/issues/122

johncordeiro:
@rogerhu #122 is another step after ParseFile implementation, because ParseFile is not a ParseObject. We could handle them separately.

PR #579, opened 2017-02-01
Attempt to re-save installation if Installation was deleted on the server

hermanliang:
Reference: [Parse-SDK-iOS-OSX/PFInstallation.m#saveAsync](https://github.com/ParsePlatform/Parse-SDK-iOS-OSX/blob/2504c19f03698b4f7d89a22708133a410b17fba9/Parse/PFInstallation.m#L224)

### DIFF
--- a/Parse/src/main/java/com/parse/ParseObject.java
+++ b/Parse/src/main/java/com/parse/ParseObject.java
@@ -2815,6 +2815,15 @@ public class ParseObject {
     }
   }
 
+  /* package */ void markAllFieldDirty() {
+    synchronized (mutex) {
+      estimatedData.clear();
+      for (String key : state.keySet()) {
+        performPut(key, state.get(key));
+      }
+    }
+  }
+
   /**
    * performOperation() is like {@link #put(String, Object)} but instead of just taking a new value,
    * it takes a ParseFieldOperation that modifies the value.
--- a/Parse/src/test/java/com/parse/ParseInstallationTest.java
+++ b/Parse/src/test/java/com/parse/ParseInstallationTest.java
@@ -29,9 +29,11 @@ import java.util.TimeZone;
 import bolts.Task;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 import static org.mockito.Matchers.any;
+import static org.mockito.Matchers.eq;
 import static org.mockito.Mockito.doReturn;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.spy;
@@ -106,6 +108,50 @@ public class ParseInstallationTest {
         assertTrue(e.getMessage().contains("Cannot modify"));
       }
     }
+  }
+
+  @Test
+  public void testSaveAsync() throws Exception {
+    String sessionToken = "sessionToken";
+    Task<Void> toAwait = Task.forResult(null);
+
+    ParseCurrentInstallationController controller =
+        mock(ParseCurrentInstallationController.class);
+    ParseInstallation currentInstallation = new ParseInstallation();
+    when(controller.getAsync())
+        .thenReturn(Task.forResult(currentInstallation));
+    ParseCorePlugins.getInstance()
+        .registerCurrentInstallationController(controller);
+    ParseObject.State state = new ParseObject.State.Builder("_Installation")
+        .put("deviceToken", "deviceToken")
+        .build();
+    ParseInstallation installation = ParseInstallation.getCurrentInstallation();
+    assertNotNull(installation);
+    installation.setState(state);
+
+    ParseObjectController objController = mock(ParseObjectController.class);
+    // mock return task when Installation was deleted on the server
+    Task<ParseObject.State> taskError = Task.forError(new ParseException(ParseException.OBJECT_NOT_FOUND, ""));
+    // mock return task when Installation was re-saved to the server
+    Task<ParseObject.State> task = Task.forResult(null);
+    when(objController.saveAsync(
+        any(ParseObject.State.class),
+        any(ParseOperationSet.class),
+        eq(sessionToken),
+        any(ParseDecoder.class)))
+        .thenReturn(taskError)
+        .thenReturn(task);
+    ParseCorePlugins.getInstance()
+        .registerObjectController(objController);
+
+    installation.saveAsync(sessionToken, toAwait);
+
+    verify(controller, times(1)).getAsync();
+    verify(objController, times(2)).saveAsync(
+        any(ParseObject.State.class),
+        any(ParseOperationSet.class),
+        eq(sessionToken),
+        any(ParseDecoder.class));
   }
 
   @Test
--- a/Parse/src/test/java/com/parse/ParseInstallationTest.java
+++ b/Parse/src/test/java/com/parse/ParseInstallationTest.java
@@ -146,7 +146,7 @@ public class ParseInstallationTest {
 
     installation.saveAsync(sessionToken, toAwait);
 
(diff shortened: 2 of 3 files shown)


rogerhu:
Besides this PR matching iOS, do you see this issue for yourself?

hermanliang:
@rogerhu 
The issue would be happended in the following steps (but doesn't happend on iOS device):
1. App with the following code
```java 
// put some parameters here
...
installation.saveInBackground();
```
2. Launch the app
3. installation added to server successfully
4. Delete the added installation on the server
5. Terminate the app and re-launch again
6. The installation not re-created to the server

**The following is my scenario:**  
I have two environments for the app. One is Sandbox using Database-A (db-A), the other is Production using Database-B (db-B).  
The releasing app might be with Sandbox (for testing/debugging purpose) or Production settings. Both settings are with the same `applicationId`, so I can easily replace the settings and keep the app data intact.  
1. The device install the app with Sandbox settings first
2. Launch the app, the installation data would be added to the db-A
3. Replace the app with Production settings on the same device
4. Launch the app, the installation data would be not added to the db-B unless I clear app data from Settings App.

If the device with the above installation scenario, the push notification won't be received in production environment due to no installation data in db-B.
But the above scenario doesn't happened on iOS devices.

rogerhu:
Great thanks for the explaination.  Is there a way to add a test for this?

hermanliang:
@rogerhu 
I added a test. Thanks!

Parse/src/test/java/com/parse/ParseInstallationTest.java
```diff
@@ -109,6 +111,50 @@ public void testImmutableKeys() {
...
+        eq(sessionToken),
+        any(ParseDecoder.class)))
+        .thenReturn(taskError)
+        .thenReturn(task);
+    ParseCorePlugins.getInstance()
+        .registerObjectController(objController);
+
+    installation.saveAsync(sessionToken, toAwait);
+
+    verify(controller, times(1)).getAsync();
```
rogerhu: minor optimization -- this can be `verify(controller).getAsync();```

```Alias to verify(mock, times(1)) E.g:


   verify(mock).someMethod("some arg");
 
Above is equivalent to:

   verify(mock, times(1)).someMethod("some arg");
 
```

Parse/src/main/java/com/parse/ParseObject.java
```diff
@@ -2815,6 +2815,15 @@ private void rebuildEstimatedData() {
     }
   }
 
+  /* package */ void markAllFieldDirty() {
+    synchronized (mutex) {
+      estimatedData.clear();
```
rogerhu: @natario1 will any changes be necessary for safeKeys() if we merge this in?

rogerhu:
I think code coverage will be higher after you rebase. :)

rogerhu:
rebased / force pushed to your branch

Parse/src/test/java/com/parse/ParseInstallationTest.java
hermanliang: update it by 94da911  
Thanks

rogerhu:
Lgtm thanks

rogerhu:
lgtm

Parse/src/main/java/com/parse/ParseObject.java
natario1: @rogerhu no, it’s OK as it is now.

natario1:
@hermanliang @rogerhu isn’t your method `ParseObject.markAllFieldsDirty` doing the same of `ParseObject.rebuildEstimatedData` ([ref](https://github.com/natario1/Parse-SDK-Android/blob/selectkeys/Parse/src/main/java/com/parse/ParseObject.java#L2872))?

I think you could have used that instead.

hermanliang:
@natario1  I think the two methods call is not the same and also for different purposes. `ParseObject.rebuildEstimatedData` does not force all fields to get dirty because it uses `estimatedData.put(key, value)` to rebuild estimatedData and it does not change the operations, but `performPut(key, value)` will update the operations.

rogerhu:
Can we use markallDirty inside the rebuildEstimatedData function?  Seems like it's repeated?

Also it should be plural markAllFields

hermanliang:
@rogerhu  They're different.
[rebuildEstimatedData](https://github.com/ParsePlatform/Parse-SDK-Android/blob/0ceda1ae9b1668407deff75f5b5972fbaf70975c/Parse/src/main/java/com/parse/ParseObject.java#L2876) vs. [markAllFieldDirty](https://github.com/ParsePlatform/Parse-SDK-Android/blob/0ceda1ae9b1668407deff75f5b5972fbaf70975c/Parse/src/main/java/com/parse/ParseObject.java#L2888)  
1. [`performPut(key, value)`](https://github.com/ParsePlatform/Parse-SDK-Android/blob/0ceda1ae9b1668407deff75f5b5972fbaf70975c/Parse/src/main/java/com/parse/ParseObject.java#L2929-L2951) also updates the operations.  
2. `rebuildEstimatedData` will be called before/after update to the server (ex. [handleSaveResultAsync](https://github.com/ParsePlatform/Parse-SDK-Android/blob/0ceda1ae9b1668407deff75f5b5972fbaf70975c/Parse/src/main/java/com/parse/ParseObject.java#L1415) -> [setState](https://github.com/ParsePlatform/Parse-SDK-Android/blob/0ceda1ae9b1668407deff75f5b5972fbaf70975c/Parse/src/main/java/com/parse/ParseObject.java#L721))

So if use markAllDirty inside the rebuildEstimatedData, it might be increased the transmission size.
1. All fields will be transferred to the server instead of modified (dirty) fields only.
2. After updating to the server, the data is dirty again.

rogerhu:
Gotcha. They are different.

Looking at https://github.com/ParsePlatform/Parse-SDK-iOS-OSX/blob/2504c19f03698b4f7d89a22708133a410b17fba9/Parse/PFInstallation.m#L73-L80 though markAllFields (should be plural) puts the estimatedData.  Yours seems to clear and then write the current state instead of the cached data?

hermanliang:
Yes, the cached installation needs to be rebuilt by new installation after being saved to the server.
To do this, the existing installation needs to be [saved as a new installation](https://github.com/ParsePlatform/Parse-SDK-Android/blob/master/Parse/src/main/java/com/parse/ParseRESTObjectCommand.java#L33-L47).
1. Clear objectID (assign `null`).
2. Mark all fields as needed for updates.

After the re-saving process, cached installation file will be overwritten by new one.

The above operation seems to not allow when the local datastore is enabled due to [objectID is unchangeable](https://github.com/ParsePlatform/Parse-SDK-Android/blob/master/Parse/src/main/java/com/parse/OfflineStore.java#L1459) after it's created. (But I think it should be allowed in this case.)

BTW, thanks for your correction. I created a PR (#602) for renaming the function (markAllFieldDirty -> markAllFieldsDirty).

rogerhu:
great thanks for the explanation.

I guess my question is that markAllFieldsDirty() in the iOS SDK (https://github.com/ParsePlatform/Parse-SDK-iOS-OSX/blob/2504c19f03698b4f7d89a22708133a410b17fba9/Parse/PFInstallation.m#L73-L80) seems to mark all the cached data entries as dirty.

it seems like you want to clear the cache, reset it to the original state, and clear object ID.

I guess my question is why there's a difference?

hermanliang:
I know your question now. 😄  
Yes, they're different and iOS's `markAllFieldsDirty` is right. 
[This line](https://github.com/ParsePlatform/Parse-SDK-Android/blob/1e882557c248b21d427596b8163b6d0b161a0263/Parse/src/main/java/com/parse/ParseObject.java#L2886) should be removed. No needs to clear the `estimatedData` before reset it to the original state.
I'll submit another PR to fix it.
Thanks!

PR #596, opened 2017-03-11
Adding availableKeys to ParseObject.State

### LINKED ISSUE #595: IllegalStateExceptions thrown when using query.selectKeys()
I am inclined to believe that this is a Android SDK bug rather than parse-server.

Steps to reproduce:

1. Create a TestTable like the following, with just to objects:

        key:        foo            bar           else
                    “yes”          111           “unwanted key”
                    “yes”                        “unwanted key”
    
    That is,

        ParseObject o1 = new ParseObject(“TestTable”);
        o1.set(“foo”, “yes”);
        o1.set(“bar”, 111);
        o1.set(“else”, “unwanted key”);
        o1.save();
        ParseObject o2 = new ParseObject(“TestTable”);
        o2.set(“foo”, “yes”);
        o2.set(“else”, “unwanted key”);
        o2.save();

2. No issues arise if you call find() as is:

        ParseQuery q = ParseQuery.getQuery(“TestTable”);
        List<ParseObject> objs = q.find();

    Calling `obj.getInt(“bar”)` will return 111 for the first object, 0 for the second.
    That’s what we are used to.

3. But if we want to exclude the “else” key,

        ParseQuery q = ParseQuery.getQuery(“TestTable”);
        q.selectKeys(Arrays.asList(“foo”, “bar”));
        List<ParseObject> objs = q.find();

    Calling `obj.getInt(“bar”)` returns 111 for the first object indeed, but throws an exception for the first object, the IllegalStateException asking you to fetch the object.

I hope the report is clear: if a selected key exists in the server but is simply undefined, <strike> `obj.has(“key”)` incorrectly returns false</strike> `obj.get(“key”)` incorrectly throws. This makes selectKeys() useless for any table that might have some undefined data among the selected keys. I am surprised this was not brought up yet.

_Note: make sure to wipe out app cache between point 2 or 3, or 3 will work because the object is cached. Or, just test 3. before testing 2._

Thank you @rogerhu for stepping up, I am happy to see this back on track.

### LINKED ISSUE #597: No support for dotted notation in query.selectKeys()
parse-server supports dotted notation in ParseQuery selectKeys.

```
ParseQuery query = ParseQuery.getQuery(“Post");
query.include(“author”);
query.selectKeys(Arrays.asList(“author”, “author.picture”));
// Get only the picture of the authors of these posts.
```

The request is sent to the server and the response is OK, but by default the Android SDK assumes that `include`d objects are complete, returning incorrect results for various APIs and breaking a few things like #595 #596 .

Specifically the incomplete (but assumed to be complete) state will overwrite the current state for that object. So, for example, even if we have full data in memory for the “author” object, it will now have only “picture”, and even worse fetchIfNeeded() will do nothing.

natario1:
- Adds a availableKeys property to ParseObject.state that includes keys that can be safely accessed. Fixes #595 so now there is consistency between complete and non-complete objects.

   While it seems to work on my machine, I have no deep understanding of the library. Nothing should be broken though. 

- Exposing `ParseObject.isDataAvailable(String key)`. Right now there are:

    -  `ParseObject.has(String key)`: alias to containsKey(key)
    - `ParseObject.containsKey(String key):` true if the server sent out data for this key

    Neither of these can be used to know if `ParseObject.get*` methods can be called safely. In some cases `has(key)` returns false but `get(key)` can be called without throwing. In some others `has(key)` returns false and `get(key)` will throw.

    This PR exposes `isDataAvailable(String key)` for partially fetched objects. When false, `get*` methods will throw. So with the PR:

   - `ParseObject.containsKey(String key)`: we have some data for this key from the server.
   - `ParseObject.isDataAvailable()`: the object is complete, all getters can be called.
   - `ParseObject.isDataAvailable(String key)`: the object might not be complete, but you can call get(key) safely.
    
  This makes sense to me.

- Fixes #597 adding support for nested selected keys. If there are some, nested objects will not be considered complete.

### DIFF
--- a/Parse/src/main/java/com/parse/NetworkQueryController.java
+++ b/Parse/src/main/java/com/parse/NetworkQueryController.java
@@ -134,8 +134,10 @@ import bolts.Task;
       if (resultClassName == null) {
         resultClassName = state.className();
       }
+      JSONArray safeKeys = new JSONArray(state.selectedKeys());
       for (int i = 0; i < results.length(); ++i) {
         JSONObject data = results.getJSONObject(i);
+        data.put("__safeKeys", safeKeys);
         T object = ParseObject.fromJSON(data, resultClassName, state.selectedKeys() == null);
         answer.add(object);
 
--- a/Parse/src/main/java/com/parse/NetworkQueryController.java
+++ b/Parse/src/main/java/com/parse/NetworkQueryController.java
@@ -134,11 +134,12 @@ import bolts.Task;
       if (resultClassName == null) {
         resultClassName = state.className();
       }
-      JSONArray safeKeys = new JSONArray(state.selectedKeys());
+      boolean isSubset = state.selectedKeys() != null;
+      JSONArray selectedKeys = isSubset ? new JSONArray(state.selectedKeys()) : null;
       for (int i = 0; i < results.length(); ++i) {
         JSONObject data = results.getJSONObject(i);
-        data.put(ParseObject.KEY_SAFEKEYS, safeKeys);
-        T object = ParseObject.fromJSON(data, resultClassName, state.selectedKeys() == null);
+        if (isSubset) data.put(ParseObject.KEY_SELECTED_KEYS, selectedKeys);
+        T object = ParseObject.fromJSON(data, resultClassName, !isSubset);
         answer.add(object);
 
         /*
--- a/Parse/src/main/java/com/parse/NetworkQueryController.java
+++ b/Parse/src/main/java/com/parse/NetworkQueryController.java
@@ -137,7 +137,7 @@ import bolts.Task;
       JSONArray safeKeys = new JSONArray(state.selectedKeys());
       for (int i = 0; i < results.length(); ++i) {
         JSONObject data = results.getJSONObject(i);
-        data.put("__safeKeys", safeKeys);
+        data.put(ParseObject.KEY_SAFEKEYS, safeKeys);
         T object = ParseObject.fromJSON(data, resultClassName, state.selectedKeys() == null);
         answer.add(object);
 
--- a/Parse/src/main/java/com/parse/NetworkQueryController.java
+++ b/Parse/src/main/java/com/parse/NetworkQueryController.java
@@ -134,12 +134,9 @@ import bolts.Task;
       if (resultClassName == null) {
         resultClassName = state.className();
       }
-      boolean isSubset = state.selectedKeys() != null;
-      JSONArray selectedKeys = isSubset ? new JSONArray(state.selectedKeys()) : null;
       for (int i = 0; i < results.length(); ++i) {
         JSONObject data = results.getJSONObject(i);
-        if (isSubset) data.put(ParseObject.KEY_SELECTED_KEYS, selectedKeys);
-        T object = ParseObject.fromJSON(data, resultClassName, !isSubset);
+        T object = ParseObject.fromJSON(data, resultClassName, ParseDecoder.get(), state.selectedKeys());
         answer.add(object);
 
         /*
--- a/Parse/src/main/java/com/parse/ParseDecoder.java
+++ b/Parse/src/main/java/com/parse/ParseDecoder.java
@@ -122,7 +122,7 @@ import org.json.JSONObject;
     }
 
     if (typeString.equals("Object")) {
-      return ParseObject.fromJSON(jsonObject, null, true, this);
+      return ParseObject.fromJSON(jsonObject, null, !jsonObject.has(ParseObject.KEY_SELECTED_KEYS), this);
     }
 
     if (typeString.equals("Relation")) {
--- a/Parse/src/main/java/com/parse/ParseDecoder.java
+++ b/Parse/src/main/java/com/parse/ParseDecoder.java
@@ -122,7 +122,7 @@ import org.json.JSONObject;
     }
 
     if (typeString.equals("Object")) {
-      return ParseObject.fromJSON(jsonObject, null, !jsonObject.has(ParseObject.KEY_SELECTED_KEYS), this);
+      return ParseObject.fromJSON(jsonObject, null, this);
     }
 
     if (typeString.equals("Relation")) {
--- a/Parse/src/main/java/com/parse/ParseObject.java
+++ b/Parse/src/main/java/com/parse/ParseObject.java
@@ -64,7 +64,7 @@ public class ParseObject {
   */
   private static final String KEY_COMPLETE = "__complete";
   private static final String KEY_OPERATIONS = "__operations";
-  /* package */ static final String KEY_SELECTED_KEYS = "__selectedKeys";
+  private static final String KEY_SELECTED_KEYS = "__selectedKeys";
   /* package */ static final String KEY_IS_DELETING_EVENTUALLY = "__isDeletingEventually";
   // Because Grantland messed up naming this... We'll only try to read from this for backward
   // compat, but I think we can be safe to assume any deleteEventuallys from long ago are obsolete
@@ -603,32 +603,45 @@ public class ParseObject {
    *          The object's data.
    * @param defaultClassName
    *          The className of the object, if none is in the JSON.
-   * @param isComplete
- *          {@code true} if this is all of the data on the server for the object.
+   * @param decoder
+   *          Delegate for knowing how to decode the values in the JSON.
+   * @param selectedKeys
+   *          Set of keys selected when quering for this object. If none, the object is assumed to
+   *          be complete, i.e. this is all the data for the object on the server.
    */
   /* package */ static <T extends ParseObject> T fromJSON(JSONObject json, String defaultClassName,
-                                                          boolean isComplete) {
-    return fromJSON(json, defaultClassName, isComplete, ParseDecoder.get());
+                                                          ParseDecoder decoder,
+                                                          Set<String> selectedKeys) {
+    boolean complete = selectedKeys == null || selectedKeys.isEmpty();
+    if (!complete) {
+      JSONArray keys = new JSONArray(selectedKeys);
+      try {
+        json.put(KEY_SELECTED_KEYS, keys);
+      } catch (JSONException e) {
+        throw new RuntimeException(e);
+      }
+    }
+    return fromJSON(json, defaultClassName, decoder);
   }
 
   /**
    * Creates a new {@code ParseObject} based on data from the Parse server.
    * @param json
-   *          The object's data.
+   *          The object's data. It is assumed to be complete, unless the JSON has the
+   *          {@link #KEY_SELECTED_KEYS} key.
    * @param defaultClassName
    *          The className of the object, if none is in the JSON.
-   * @param isComplete
-   *          {@code true} if this is all of the data on the server for the object.
    * @param decoder
    *          Delegate for knowing how to decode the values in the JSON.
    */
   /* package */ static <T extends ParseObject> T fromJSON(JSONObject json, String defaultClassName,
-                                                          boolean isComplete, ParseDecoder decoder) {
+                                                          ParseDecoder decoder) {
     String className = json.optString(KEY_CLASS_NAME, defaultClassName);
     if (className == null) {
       return null;
     }
     String objectId = json.optString(KEY_OBJECT_ID, null);
+    boolean isComplete = !json.has(KEY_SELECTED_KEYS);
     @SuppressWarnings("unchecked")
     T object = (T) ParseObject.createWithoutData(className, objectId);
     State newState = object.mergeFromServer(object.getState(), json, decoder, isComplete);
@@ -641,7 +654,7 @@ public class ParseObject {
    *
    * Method is used by parse server webhooks implementation to create a
    * new {@code ParseObject} from the incoming json payload. The method is different from
-   * {@link #fromJSON(JSONObject, String, boolean)} ()} in that it calls
+   * {@link #fromJSON(JSONObject, String, ParseDecoder, Set)} ()} in that it calls
    * {@link #build(JSONObject, ParseDecoder)} which populates operation queue
    * rather then the server data from the incoming JSON, as at external server the incoming
    * JSON may not represent the actual server data. Also it handles
--- a/Parse/src/main/java/com/parse/ParseObject.java
+++ b/Parse/src/main/java/com/parse/ParseObject.java
@@ -65,7 +65,7 @@ public class ParseObject {
   private static final String KEY_COMPLETE = "__complete";
   private static final String KEY_OPERATIONS = "__operations";
   // Array of keys selected when querying for the object. Helps decoding nested {@code ParseObject}s
-  // correctly, and helps constructing the {@code State.safeKeys()} set.
+  // correctly, and helps constructing the {@code State.availableKeys()} set.
   private static final String KEY_SELECTED_KEYS = "__selectedKeys";
   /* package */ static final String KEY_IS_DELETING_EVENTUALLY = "__isDeletingEventually";
   // Because Grantland messed up naming this... We'll only try to read from this for backward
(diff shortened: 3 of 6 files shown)


rogerhu:
Thanks.  Do you know how the iOS version handles this issue?

natario1:
@rogerhu I don't know. As far as I know the JS sdk simply does not throw, so you can call get with any key, and if not present, you get undefined.

flovilmart:
What's the behavior on the iOS SDK? Should remove the throwing altogether?

rogerhu:
So iOS Parse has selectKeys() and findObjects() as part of [PFQuery] (https://github.com/ParsePlatform/Parse-SDK-iOS-OSX/blob/7a820b75c6726808475af9ce705053ff0e8cbd11/Parse/PFQuery.m) and [PFQueryState] (https://github.com/ParsePlatform/Parse-SDK-iOS-OSX/blob/7a820b75c6726808475af9ce705053ff0e8cbd11/Parse/Internal/Query/State/PFQueryState.m)

Is selectedKeys() the equivalent of safeKeys? (https://github.com/ParsePlatform/Parse-SDK-iOS-OSX/blob/7a820b75c6726808475af9ce705053ff0e8cbd11/Tests/Unit/QueryStateUnitTests.m)

natario1:
@rogerhu we also have selectedKeys() as part of ParseQuery.State and that’s used to understand if the object should be considered ‘complete’. So if you select some keys all the objects will be incomplete. The issue is that when objects are incomplete and you call get* on a legit key, ParseObject will throw.

The PR adds safeKeys to ParseObject.State , meaning, ‘keys we don’t currently have in our data, but you should be able to call get() on without crashing the app’.

This is dirty and I would be OK with @flovilmart with simply removing the throwing.

rogerhu:
Where do we need to remove IllegalStateException throwing?  I think that makes more sense too to keep consistent with the other SDKs.

natario1:
I have found how iOS handles this issue, [exactly here](https://github.com/ParsePlatform/Parse-SDK-iOS-OSX/blob/master/Parse/PFObject.m#L767).

There’s a `availableKeys` set, that holds actual "keys received from the server" **plus** "keys selected on the query".

This is practically the same to this PR (except that here the `safeKeys` Set means "keys selected on the query but not received from the server"). So this looks OK to me. What do you think?

rogerhu:
Should we call it availableKeys instead then?

rogerhu:
Does it deal with dotted notation as well?

natario1:
Yes, I can refactor to give the same meaning it has in the iOS SDK. It’s a bit risky though, because in that case we must ensure that the `availableKeys` set is always consistent with the `serverData` map.

I don’t fully understand that part (server data is held in two different maps both in `ParseObject` and in `ParseObject.State`...). 

Yes, this implements dotted notation.

rogerhu:
I'm more curious if there's anywhere in the iOS code that deals with dotted notation issues as well...

natario1:
@rogerhu it seems they don't, so they should have the #597 bug.

The relevant part is [here](https://github.com/ParsePlatform/Parse-SDK-iOS-OSX/blob/master/Parse/Internal/PFDecoder.m#L89): every decoded nested object is created as ‘complete’, which is wrong if some keys where selected.

The exact same line for the Android SDK is [here](https://github.com/ParsePlatform/Parse-SDK-Android/pull/596/files#diff-e30b7266f09ee255a64daa415de45d25L125) and this fixes it.

Parse/src/main/java/com/parse/ParseObject.java
```diff
@@ -921,8 +939,34 @@ protected boolean visit(Object object) {
...
+        }
 
         Object value = json.get(key);
+        if (value instanceof JSONObject && json.has(KEY_SELECTED_KEYS)) {
+          // This might be a ParseObject. Pass selected keys to understand if it is complete.
+          JSONArray selectedKeys = json.getJSONArray(KEY_SELECTED_KEYS);
+          JSONArray nestedKeys = new JSONArray();
+          for (int i = 0; i < selectedKeys.length(); i++) {
+            String nestedKey = selectedKeys.getString(i);
+            if (nestedKey.startsWith(key+".")) nestedKeys.put(nestedKey.substring(key.length()+1));
```
rogerhu: What happena if this is multiple nested keys?  A.b.c.d.e?

Parse/src/test/java/com/parse/ParseObjectTest.java
```diff
@@ -237,7 +238,16 @@ public void testGetUnavailable() {
     ParseObject.State state = mock(ParseObject.State.class);
     when(state.className()).thenReturn("TestObject");
     when(state.isComplete()).thenReturn(false);
+    ParseObject object = ParseObject.from(state);
+    object.get("foo");
```
rogerhu: Is this just an assertion that no exception will be raised?

natario1: No, this checks that the exception is raised (see the @Test annotation). Actually I have not changed this method, I have added testGetAvailableIfKeySafe() a few lines below, that tests safeKeys() .

Parse/src/main/java/com/parse/ParseObject.java
natario1: At first (942-954) the current object will be given its own direct keys (without dots, e.g. b). 
Later (958+) we extract the child object subkeys (e.g. b.c.d.e). That is passed to decoder.decode(), which will pass back the JSON object to this method. It will get ‘b’ as a safe key, and will pass ‘c.d.e’... and so on. It goes this way with nested iterations, until there are no nested ParseObjects.

rogerhu: Great thanks for the explanation

rogerhu: Do we have a test for this part?

natario1: I have added two in ParseDecoderTest, I will add another one for the a.b situation.

rogerhu: Thanks also the key + ”.” could be spaced out

Parse/src/main/java/com/parse/ParseObject.java
```diff
@@ -578,33 +599,30 @@ public Void then(Task<Void> task) throws Exception {
...
    * Creates a new {@code ParseObject} based on data from the Parse server.
-   *
    * @param json
    *          The object's data.
    * @param defaultClassName
    *          The className of the object, if none is in the JSON.
    * @param isComplete
    *          {@code true} if this is all of the data on the server for the object.
    * @param decoder
-   *          Delegate for knowing how to decode the values in the JSON.
```
rogerhu: Was this line deleted accidentally?

natario1: Yes, good catch.

Parse/src/main/java/com/parse/NetworkQueryController.java
```diff
@@ -134,9 +134,12 @@ public Integer then(Task<JSONObject> task) throws Exception {
       if (resultClassName == null) {
         resultClassName = state.className();
       }
+      boolean isSubset = state.selectedKeys() != null;
+      JSONArray selectedKeys = isSubset ? new JSONArray(state.selectedKeys()) : null;
       for (int i = 0; i < results.length(); ++i) {
         JSONObject data = results.getJSONObject(i);
-        T object = ParseObject.fromJSON(data, resultClassName, state.selectedKeys() == null);
+        if (isSubset) data.put(ParseObject.KEY_SELECTED_KEYS, selectedKeys);
```
rogerhu: Can you pass in selectedKeys into a method instead that accepts selectedKeys() instead?

https://github.com/ParsePlatform/Parse-SDK-iOS-OSX/blob/master/Parse/PFObject.m#L772-L774

It seems weird that the NetworkQueryController has access to this private key..

Parse/src/main/java/com/parse/ParseDecoder.java
```diff
@@ -122,7 +122,7 @@ public Object decode(Object object) {
     }
 
     if (typeString.equals("Object")) {
-      return ParseObject.fromJSON(jsonObject, null, true, this);
+      return ParseObject.fromJSON(jsonObject, null, !jsonObject.has(ParseObject.KEY_SELECTED_KEYS), this);
```
rogerhu: it seems to me that fromJSON() should figure out whether things are complete instead of here right?

Just trying to get KEY_SELECTED_KEYS out of being a public static variable..

Parse/src/main/java/com/parse/NetworkQueryController.java
natario1: @rogerhu I get your point, I tried working directly with a Set, but for that we would need to change the signatures of ParseObject.fromJSON, ParseObject.mergeFromServer, and even worse ParseDecoder.decode, these are called from anywhere and an extra set of "selected keys" rarely makes sense.
The key is also used by ParseDecoder to get keys from the JSON...

Maybe we could add two static methods to ParseObject, `void insertSelectedKeys(JSONObject data, Set<String>)` , used by NetworkQueryController, and `boolean hasSelectedKeys(JSONObject data)` , used by ParseDecoder. What do you think?

natario1: Or I’ll just add another ParseObject.fromJSON that deals with this.

rogerhu: Yep :)

rogerhu:
Thanks for working on this bug/issue..quite complicated to fix and appreciate the effort.  Most of my feedback are minor issues in general.

Parse/src/main/java/com/parse/ParseObject.java
```diff
@@ -921,8 +940,34 @@ protected boolean visit(Object object) {
...
+          JSONArray nestedKeys = new JSONArray();
+          for (int i = 0; i < selectedKeys.length(); i++) {
+            String nestedKey = selectedKeys.getString(i);
+            if (nestedKey.startsWith(key + ".")) nestedKeys.put(nestedKey.substring(key.length() + 1));
+          }
+          if (nestedKeys.length() > 0) {
+            ((JSONObject) value).put(KEY_SELECTED_KEYS, nestedKeys);
+          }
+        }
         Object decodedObject = decoder.decode(value);
```
natario1: Well we need a fromJSON signature that does not take selectedKeys() .. because ParseDecoder will just have a JSON and is not aware of the selected keys bundled in the JSON (I made KEY_SELECTED_KEYS private as you said), [see](https://github.com/ParsePlatform/Parse-SDK-Android/pull/596/files#diff-e30b7266f09ee255a64daa415de45d25L125).

From this point on, bundling the keys in the JSON allows us to not change all the other signatures, and abstract this internal stuff from ParseDecoder. It is also used in Local data store encoding/decoding, see ParseObject.toRest() / fromREST()

Parse/src/main/java/com/parse/ParseObject.java
```diff
@@ -98,6 +99,7 @@ private static ParseObjectSubclassingController getSubclassingController() {
       private long createdAt = -1;
       private long updatedAt = -1;
       private boolean isComplete;
+      private Set<String> safeKeys = new HashSet<>();
       /* package */ Map<String, Object> serverData = new HashMap<>();
```
natario1: Haha yes. Actually the server does not return these, we have them because of query.selectKeys(), but you got the point. Annoying bug.

Parse/src/main/java/com/parse/ParseObject.java
rogerhu: Gotcha forgot about local store too

rogerhu:
Lgtm thanks

flovilmart:
Can we rename those to availableKeys?

rogerhu:
@natario1 how easy would it be to change the construct to be availableKeys?  The main thing I think is this function here: https://github.com/natario1/Parse-SDK-Android/blob/fcf24e3af94f0321d80d399c418a325e7e3da90f/Parse/src/main/java/com/parse/ParseObject.java#L168-L174

natario1:
Ok guys, reverted that logic and renamed to availableKeys() .

Since this has become something serious, the next design question would be if we should add the set also to `ParseObject` (not only `ParseObject.State`). As far as I understand the `ParseObject.State` is synced with the server, while `ParseObject` members do also trace dirty changes.

So, for example, you have the `ParseObject.State.serverData` map and the `ParseObject.estimatedData` map. The latter is modified by put operations. It appears they are synced together after save operations.

So, now that we have `ParseObject.State.availableKeys`, should we also add `ParseObject.estimatedKeys`?

- Pros: would allow to change [this line](https://github.com/natario1/Parse-SDK-Android/blob/selectkeys/Parse/src/main/java/com/parse/ParseObject.java#L3446) from

        public boolean isDataAvailable(String key) {
          synchronized (mutex) {
            // Fallback to estimatedData to include dirty changes.
           return isDataAvailable() || state.availableKeys().contains(key) || estimatedData.containsKey(key);
          }
        }

   to a much cleaner

        public boolean isDataAvailable(String key) {
          synchronized (mutex) {
           return isDataAvailable() || estimatedKeys.contains(key);
          }
        }

- Pros: consistency with `estimatedData` and with the iOS SDK implementation of isDataAvailable

- Cons: adding a useless Set for every ParseObject (it will differ from `estimatedData.keySet()` only when the object comes from a Query with selected keys...).

- Cons: must be 100% sure that `estimatedKeys` follows operation onto `estimatedData` or we will throw when we shouldn't. Changing the implementation of an important method (`isDataAvailable(key)`) and I don’t think I have enough understanding of the class.

What do you think @rogerhu @flovilmart ? I think it can be part of a future PR ...

natario1:
Just to clarify, the current implementation (with this PR) of `isDataAvailable(key)` **does** follow dirty changes. This works. We just have to check estimated data.

flovilmart:
If it works, don't change the internal yet :)

rogerhu:
Good enough for now. :). Thanks!

natario1:
When are you planning to release next version @rogerhu ?

flovilmart:
Snapshots are released with each merge. That should be good to test no?

rogerhu:
Working also on Bintray access so I can deploy the v14.0.0 version currently on snapshot too. :)

natario1:
Nice, no hurry. By the way, we are pushing 1.14.1 to Sonatype, is that a typo? (should be 1.14.0)

rogerhu:
Yeah my fail.  :)  It's 1.14.1-SNAPSHOT (I'll fix it to 1.14.0)

Parse/src/main/java/com/parse/NetworkQueryController.java
```diff
@@ -136,7 +136,7 @@ public Integer then(Task<JSONObject> task) throws Exception {
       }
       for (int i = 0; i < results.length(); ++i) {
         JSONObject data = results.getJSONObject(i);
-        T object = ParseObject.fromJSON(data, resultClassName, state.selectedKeys() == null);
+        T object = ParseObject.fromJSON(data, resultClassName, ParseDecoder.get(), state.selectedKeys());
```
rogerhu: @natario1 do you remember why we need to pass in ParseDecoder.get() here?

natario1: @rogerhu it’s just an instance of ParseDecoder.

I had to change the signatures of ParseObject.fromJSON and this seemed legit. ParseDecoder was present in the other signature (used [here](https://github.com/ParsePlatform/Parse-SDK-Android/blob/master/Parse/src/main/java/com/parse/ParseDecoder.java#L125) by ParseDecoder itself, passing `this`), so I added it here as well for consistency.

Since ParseDecoder is a singleton we could just drop it from both signatures, but I guess original authors wanted to keep the chance of passing a different decoder. I don’t know.

PR #598, opened 2017-03-13
Set min coverage to 45%

flovilmart:
Wait for codecov to respond OK with coverage status > 45%


flovilmart:
@rogerhu that's what I intended originally with the codecov.yaml. Would be nice to go above that 45% value still.. OR configure so we have large coverage on the core or the SDK, and less on the overall.

rogerhu:
Lgtm thanks

PR #600, opened 2017-03-14
Support for setting up Server Url in the Manifest

hermanliang:
Since `parse.com` has stopped, the Manifest for `Parser Server` setting up is not appropriate now. 
This PR add a support for setting up ServerUrl in the Manifest. Also updates the `ParseStarterProject`.

### DIFF
--- a/Parse/src/main/java/com/parse/Parse.java
+++ b/Parse/src/main/java/com/parse/Parse.java
@@ -51,7 +51,7 @@ public class Parse {
       private Context context;
       private String applicationId;
       private String clientKey;
-      private String server = "https://api.parse.com/1/";
+      private String server = null;
       private boolean localDataStoreEnabled;
       private List<ParseNetworkInterceptor> interceptors;
 
@@ -98,6 +98,7 @@ public class Parse {
           Context applicationContext = context.getApplicationContext();
           Bundle metaData = ManifestInfo.getApplicationMetadata(applicationContext);
           if (metaData != null) {
+            server(metaData.getString(PARSE_SERVER_URL));
             applicationId = metaData.getString(PARSE_APPLICATION_ID);
             clientKey = metaData.getString(PARSE_CLIENT_KEY);
           }
@@ -145,7 +146,7 @@ public class Parse {
 
         // Add an extra trailing slash so that Parse REST commands include
         // the path as part of the server URL (i.e. http://api.myhost.com/parse)
-        if (server.endsWith("/") == false) {
+        if (server != null && !server.endsWith("/")) {
           server = server + "/";
         }
 
@@ -223,6 +224,7 @@ public class Parse {
     }
   }
 
+  private static final String PARSE_SERVER_URL = "com.parse.SERVER_URL";
   private static final String PARSE_APPLICATION_ID = "com.parse.APPLICATION_ID";
   private static final String PARSE_CLIENT_KEY = "com.parse.CLIENT_KEY";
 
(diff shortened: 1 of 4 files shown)


rogerhu:
What Wiki documentation needs to be updated?

rogerhu:
Might also need a test for the new exception check :)

hermanliang:
@rogerhu test added by ea95d23  
The Wiki seems out of date and hosted on [another project](https://github.com/parseplatform/docs/tree/gh-pages/_includes/android).  
I think I can update it later. :)

Parse/src/main/java/com/parse/Parse.java
```diff
@@ -98,6 +101,7 @@ public Builder(Context context) {
           Context applicationContext = context.getApplicationContext();
           Bundle metaData = ManifestInfo.getApplicationMetadata(applicationContext);
           if (metaData != null) {
+            server(metaData.getString(PARSE_SERVER_URL));
```
flovilmart: Shouldn't it be `server = ...`

rogerhu: There is a method in line 148 that checks to make sure there is a trailing slash so I think it's OK..

flovilmart: Oh ok. Didn't see that

Parse/src/main/java/com/parse/Parse.java
```diff
@@ -51,7 +51,7 @@
       private Context context;
       private String applicationId;
       private String clientKey;
-      private String server = "https://api.parse.com/1/";
+      private String server = null;
```
rogerhu: I think it can just be server; --> null is redundant

hermanliang: fixed
Thanks

Parse/src/test/java/com/parse/ParseClientConfigurationTest.java
```diff
@@ -105,6 +120,72 @@ public void testSetNetworkInterceptors() {
...
+    verify(context).getPackageName();
+    verify(packageManager).getApplicationInfo(packageName, PackageManager.GET_META_DATA);
+    verify(metaData).getString(PARSE_SERVER_URL);
+    verify(metaData).getString(PARSE_APPLICATION_ID);
+    verify(metaData).getString(PARSE_CLIENT_KEY);
+  }
+
+  @Test(expected = RuntimeException.class)
+  public void testConfigureFromManifestWithoutServer() throws Exception {
+    Context context = mock(Context.class);
```
rogerhu: Roboelectric should already have a mocked Context class.  Do you need do to all this setup mocking?

Parse/src/test/java/com/parse/ParseClientConfigurationTest.java
```diff
@@ -105,6 +120,72 @@ public void testSetNetworkInterceptors() {
...
+    PackageManager packageManager = mock(PackageManager.class);
```
rogerhu: can you use RuntimeEnvironment.getRobolectricPackageManager()?

Parse/src/test/java/com/parse/ParseClientConfigurationTest.java
hermanliang: change to use Robolectric  
Thanks

Parse/src/test/java/com/parse/ParseClientConfigurationTest.java
hermanliang: RuntimeEnvironment.getRobolectricPackageManager() is [deprecated](http://robolectric.org/migrating/) in `Robolectric 3.3`
using `shadowOf(packageManager)` instead
Thanks

Parse/src/test/java/com/parse/ParseClientConfigurationTest.java
```diff
@@ -22,9 +32,21 @@
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+import static org.robolectric.Shadows.shadowOf;
 
+@RunWith(RobolectricTestRunner.class)
+@Config(manifest = Config.NONE, packageName = "com.parse.example")
```
rogerhu: Why do you need these two lines?  Did tests never run in this class?

rogerhu:
@hermanliang yeah please submit a PR for the wiki changes after this gets merged in

Parse/src/test/java/com/parse/ParseClientConfigurationTest.java
hermanliang: line 39 is needed to setup Robolectric environment
line 40 yeah, configuring `packageName` is redundant (cat just use the default value)
If `manifest` is not configured, the following log is printed during the test.
```
WARNING: No manifest file found at ./AndroidManifest.xml.
Falling back to the Android OS resources only.
To remove this warning, annotate your test class with @Config(manifest=Config.NONE).
```
But not affect the test.
It seems more appropriate using `@Config(constants = BuildConfig.class)` instead. (http://robolectric.org/getting-started/)
Robolectric will attempt to configure using project's manifest.
Thanks!

hermanliang:
@rogerhu  Sure, I will update the Wiki and [API reference pages](https://github.com/ParsePlatform/Parse-SDK-Android/tree/gh-pages/api) as well.  😄

PR #603, opened 2017-03-15
Minor optimization in ParseObject.markAllFieldsDirty()

hermanliang:
Continuation of the PR #579. 
No needs to clear the `estimatedData` before reset it to the original state.


rogerhu:
Thanks!

Issue #605, opened 2017-03-16
Re-save installation is allowed if LDS is enabled 


hermanliang:
Following PR #579  
The installation data will attempt to resave if installation is removed on the server.
But the resaving is not allowed when the local datastore is enabled due to [objectID is unchangeable](https://github.com/ParsePlatform/Parse-SDK-Android/blob/e0f096e850511e4dea7f1b0781186cf32e83e12b/Parse/src/main/java/com/parse/OfflineStore.java#L1459) in offline store.  
But I think [in this case](https://github.com/ParsePlatform/Parse-SDK-Android/pull/579#issuecomment-285926783) should be allowed .  

1. The `setObjectId(id)` seems only [updating the in-memory map](https://github.com/ParsePlatform/Parse-SDK-Android/blob/e0f096e850511e4dea7f1b0781186cf32e83e12b/Parse/src/main/java/com/parse/OfflineStore.java#L1462-L1474) (call sequences: [1](https://github.com/ParsePlatform/Parse-SDK-Android/blob/e0f096e850511e4dea7f1b0781186cf32e83e12b/Parse/src/main/java/com/parse/ParseObject.java#L1241) -> [2](https://github.com/ParsePlatform/Parse-SDK-Android/blob/e0f096e850511e4dea7f1b0781186cf32e83e12b/Parse/src/main/java/com/parse/ParseObject.java#L1268) -> [3](https://github.com/ParsePlatform/Parse-SDK-Android/blob/e0f096e850511e4dea7f1b0781186cf32e83e12b/Parse/src/main/java/com/parse/OfflineStore.java#L1454)). Maybe just unregister old object from the map when `setObjectId(null)` and allowing `ParseInstalltion` type only.
[...]

PR #607, opened 2017-03-20
Re-save installation is allowed if LDS is enabled

hermanliang:
Following PR #579 and resolving Issue #605

### DIFF
--- a/Parse/src/main/java/com/parse/ParseInstallation.java
+++ b/Parse/src/main/java/com/parse/ParseInstallation.java
@@ -149,13 +149,16 @@ public class ParseInstallation extends ParseObject {
       @Override
       public Task<Void> then(Task<Void> task) throws Exception {
         // Retry the fetch as a save operation because this Installation was deleted on the server.
-        if(task.getError() != null
-                && task.getError() instanceof ParseException
-                && ((ParseException) task.getError()).getCode() == ParseException.OBJECT_NOT_FOUND) {
-          synchronized (mutex) {
-            setState(new State.Builder(getState()).objectId(null).build());
-            markAllFieldsDirty();
-            return ParseInstallation.super.saveAsync(sessionToken, toAwait);
+        if (task.getError() != null
+            && task.getError() instanceof ParseException) {
+          int errCode = ((ParseException) task.getError()).getCode();
+          if (errCode == ParseException.OBJECT_NOT_FOUND
+              || (errCode == ParseException.MISSING_REQUIRED_FIELD_ERROR && getObjectId() == null)) {
+            synchronized (mutex) {
+              setState(new State.Builder(getState()).objectId(null).build());
+              markAllFieldsDirty();
+              return ParseInstallation.super.saveAsync(sessionToken, toAwait);
+            }
           }
         }
         return task;
--- a/Parse/src/main/java/com/parse/ParseInstallation.java
+++ b/Parse/src/main/java/com/parse/ParseInstallation.java
@@ -149,9 +149,7 @@ public class ParseInstallation extends ParseObject {
       @Override
       public Task<Void> then(Task<Void> task) throws Exception {
         // Retry the fetch as a save operation because this Installation was deleted on the server.
-        // Do not attempt to resave an object if LDS is enabled, since changing objectId is not allowed.
-        if(!Parse.isLocalDatastoreEnabled()
-                && task.getError() != null
+        if(task.getError() != null
                 && task.getError() instanceof ParseException
                 && ((ParseException) task.getError()).getCode() == ParseException.OBJECT_NOT_FOUND) {
           synchronized (mutex) {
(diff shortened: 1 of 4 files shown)


rogerhu:
Separate question.  have you considered just setting applicationId to be different for dev/prod?  Mixing cached data in general across prod/dev is going to cause you lots of problems in general.

Parse/src/main/java/com/parse/ParseInstallation.java
```diff
@@ -143,9 +143,7 @@ public String getInstallationId() {
       @Override
       public Task<Void> then(Task<Void> task) throws Exception {
         // Retry the fetch as a save operation because this Installation was deleted on the server.
-        // Do not attempt to resave an object if LDS is enabled, since changing objectId is not allowed.
```
rogerhu: So is it fairly easy to null out your objectId() on a failure and then cause multiple install id's per app?

hermanliang:
@rogerhu Thanks for your suggestion.
Yes, you're right, different appId can prevent problems from mixing cached data. But in my case, I need to retrieve the local stored data (such as SharedPreference, sqlite, ...) when dev/prod switching. If using different appId, it would be hard to get these data from another app. 
Another reason is that in this case, Parse only serves as a push server. Just use the Installation to save the push related information. It's easier to handle the problem, just delete the installation on the server and let app to re-create it and update the cached data as well. I believe if including other Classes (table) in this case will be very painful to handle the mixing cached data problem, and using different appId will be considered.

Parse/src/main/java/com/parse/ParseInstallation.java
hermanliang: Only the `ParseException.OBJECT_NOT_FOUND` ([line 148](ParseException.OBJECT_NOT_FOUND)) can trigger the re-creating process. 
If just null out the objectId and save installation will not create another installation data. (It's weird, I'm still figuring out why)  
BTW, I think [`setObjectId()`](https://github.com/ParsePlatform/Parse-SDK-Android/blob/05795898c2cf63e78bedba6c9e20f569294685df/Parse/src/main/java/com/parse/ParseObject.java#L1232) should not be a public method, package private may be more appropriate.

rogerhu: Gotcha line 148 is indeed key.

rogerhu:
I think you really shouldn't be mixing dev/prod.  :)  Even your push infra should silo the two environments as well (separate GCM keys).   

Your PR seems fine for the problem it tries to rectify...given you pointed out the iOS SDK tries to do something similar does it do the same thing?

hermanliang:
@rogerhu The iOS SDK is NOT allowed this operation if [LDS is enabled](https://github.com/ParsePlatform/Parse-SDK-iOS-OSX/blob/2504c19f03698b4f7d89a22708133a410b17fba9/Parse/PFInstallation.m#L227).  
I also figure out why setObjectId() is not permitted when using LDS.
Avoid inappropriate `pin()` into datastore would be the major concern.
For example, if setObject() is allowed when LDS is enabled
```java
ParseObject object = new ParseObject("Test");
object.put("key", "value");
object.save(); // add to server
object.pin(); // add to LDS
object.setObjectId(null); // for replication purpose
object.save(); // add to server
object.pin(); // will REPLACE previous object to LDS
ParseQuery.getQuery("Test").count(); // count: 2 (server)
ParseQuery.getQuery("Test").fromPin().count(); // count: 1 (datastore)
```

In ParseInstallation, following operation will clear/replace the cached installation's objectId (shouldn't be allowed, fixed in PR #611) 
```
// if LDS disabled
installation.setObjectId(null);
installation.save(); // save task will succeed, but cached objectId will be cleared

// If LDS enabled
installation.setObjectId(null);
installation.pin(); // objectId in datastore will be cleared
```
If setObjectId() is not allowed in ParseInstallation (PR #611), I think the above case will not happen. 
And I think there is no other safety issue in this PR.

rogerhu:
Does this change then how the Android version does it to ios the

rogerhu:
Thanks @hermanliang - this is on a backlog, and appreciate you doing the work on both platforms.

Can anyone take a look and provide any feedback?  I think it addresses a weird corner case of mixing dev/prod instances data.  But @hermanliang  has done a ton of work and would love help taking a look.

Parse/src/main/java/com/parse/ParseInstallation.java
natario1: btw this line can be removed, there is `instanceof` below

hermanliang:
Thanks @rogerhu 
Yeah, we always want to save the installation successfully.
Here's what I'm going to update in this PR.
1. I think #611 can be integrated into this PR because they are relevant.
2. There is another edge case stated in https://github.com/parse-community/Parse-SDK-Android/issues/627#issuecomment-292382766. I think we may add this logic in this PR as well.

rogerhu:
rebase

rogerhu:
rebase

hermanliang:
rebased

rogerhu:
This build failed actually.  Can you fix?

https://travis-ci.org/parse-community/Parse-SDK-Android

hermanliang:
@rogerhu fixed by #642 
plz check
Thanks

PR #611, opened 2017-03-23
Not allowing setObjectId() in ParseInstallation

hermanliang:
When I deal with the issue #607, I found another issue about ParseInstallation.
(It's a very special case, generally should not happen.)

```java
installation.setObjectId(null);
// put some parameters here
installation.save();
```
I originally expected another installation data will be created in the server.
But the installation data is `updated` successfully and doesn't return any exceptions.
The objectId of cached installation data is also cleared (null).
It seems parser server will refer to 'installationId' when updating the installation data.
In this case, the installation data is not re-created when it is deleted on the server. (will throw [Code 135: MissingRequiredFieldError](http://docs.parseplatform.org/android/guide/#installation-related-errors) exception)
I think it should be restricted the usage of setObjectId() in ParseInstallation.

### DIFF
--- a/Parse/src/main/java/com/parse/ParseInstallation.java
+++ b/Parse/src/main/java/com/parse/ParseInstallation.java
@@ -95,6 +95,11 @@ public class ParseInstallation extends ParseObject {
   }
 
   @Override
+  public void setObjectId(String newObjectId) {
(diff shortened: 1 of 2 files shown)


rogerhu:
rebase

PR #617, opened 2017-03-30
Add ProGuard section to README

hermanliang:


rogerhu:
I'd just add it here:

https://github.com/ParsePlatform/Parse-SDK-Android/blob/master/Parse/release-proguard.pro

hermanliang:
It seems necessary when installing SDK in gradle

```groovy
android {
    ...
    buildTypes {
        release {
            minifyEnabled true
            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
        }
    }
}

dependencies {
  ...
  compile 'com.parse:parse-android:1.14.1'
}
```

then run
```
./gradlew assembleRelease
```

BUILD FAILED unless I add the rule

rogerhu:
Correct but adding it the proguard file should include it with the lib

https://github.com/codepath/android_guides/wiki/Building-your-own-Android-library#using-with-proguard

hermanliang:
Does it due to SDK release jar not aar?

rogerhu:
We release both.  There is already a ProGuard file though...Are you saying it gets ignored for you if you were to add this line and use jetpack.io to confirm?

Jawnnypoo:
Yeah, this is probably needed now, since we have a hard dependency now on OkHttp

https://github.com/square/okhttp#proguard

rogerhu:
I understand that part the question is whether we can just add it to https://github.com/ParsePlatform/Parse-SDK-Android/blob/master/Parse/release-proguard.pro.

Otherwise we should update the docs to include all the ProGuard lines listed there.

Jawnnypoo:
Ah yeah, my bad. Yes, we can just add it there, should fix the issue.

rogerhu:
I guess the one thing you have to do is use @aar in order to include the ProGuard lines 

```
compile 'com.parse:parse-android:1.14.1@aar'
```

Jawnnypoo:
@rogerhu Interesting, might be worth adding that as the instruction in the README instead then

rogerhu:
Yep, we should point people to the ProGuard config if they need to manually use it.

Do people still use non-JARs?  What if we made the packaging type AAR by default as well?

Jawnnypoo:
I am all for that, I see no reason to have it build a jar by default, as it is an Android library, not a Java library.

rogerhu:
https://github.com/ParsePlatform/Parse-SDK-Android/pull/621

rogerhu:
To be clear we build both jar and aar.  The change just causes the pom XML file to tell Gradle to use the war version.

 Gradle in the past has difficulties with transient dependencies (I.e. okhttp depends on okio) with aar so I hope this is the only change necessary and we don't also need to use transient=true too

hermanliang:
Thanks to @rogerhu @Jawnnypoo 
This issue should be resolved by #621. 
I think we can close this PR.

PR #621, opened 2017-04-03
Switch default packaging to AAR

rogerhu:
@hermanliang @Jawnnypoo


Jawnnypoo:
LGTM. I think there should be no problems with it

PR #624, opened 2017-04-04
Implementing Parcelable in ParseFile

natario1:
ParseFile Parcelable implementation. With respect to #557 this

- does not parcel the whole byte array
- throws if file is not saved
- safely parcels strings that might be null


rogerhu:
rebase

natario1:
@rogerhu this is ready in my opinion

PR #619, opened 2017-04-03
Implementing Parcelable in ParseGeoPoint

natario1:
As per title. It’s the easiest step in bringing Parcelable into the SDK.

PR #625, opened 2017-04-04
Parcelable object

natario1:
This is a working stub that implements `Parcelable` interface into `ParseObject` ( #122 ).

- Implements Parcelable into `ParseACL`
- Implements Parcelable into `ParseRelation`
- Implements Parcelable into `ParseObject`. It parcels both server data and dirty data. Internally this is done by parceling the `ParseObject.State` for server data, and the `ParseOperationSet`s attached to the object. These sets are squashed into a single `ParseOperationSet`, as outlined by grantland in #122 , and the merged set is saved to parcel.
- Creates `ParseParcelEncoder` and `ParseParcelDecoder`, utility classes that help parceling when we host untyped lists (e.g. `Map<String, Object>`). They work by saving a type string into the parcel and checking that when unparceling. Eventually these classes will be extended to work for `ParseQuery` as well.
- Introduces (untested) smooth parceling for `ParseObject` subclasses. They do not need to provide a `CREATOR` static field. Instead, we have two new protected methods in `ParseObject` which should be familiar to any developer, `onSaveInstanceState(Bundle)` and `onRestoreInstanceState(Bundle)`. For example, they are used in `ParseUser` to keep the `isCurrentUser` boolean. We pass a Bundle because that’s much safer than a `Parcel`.
- Should work out of the box for internal subclasses: `ParsePin`, `ParseInstallation`, `ParseSession`, `ParseRole`, `EventuallyPin` and finally `ParseUser`
- Special treatment for `ParseUser` since it has an internal `State` class and own fields (isNew, isCurrentUser).
- Warns the developer (with `Log.w`) if the object about to be parceled has ongoing save tasks / scheduled saveEventually tasks, to make him aware of the implications outlined by grantland in #122 .
- Warns the developer (with `Log.w`) if the object has ongoing delete tasks / scheduled deleteEventually tasks, as above

### Approach

My approach was to never give `Parcelable` to inner classes (e.g. `ParseObject.State`, `ParseACL.Permissions`) even if they need to be parceled. That is because unparceling requires the `CREATOR` field to be publicly accessible, and we don’t want to make `ParseObject.State` public.

Instead, simple methods like writeToParcel or constructor(Parcel) are implemented to inner classes, and are called from the outer public class. Internally this ensures also inheritance when needed (e.g. `ParseUser.State` that extends `ParseObject.State`). We use `Parcelable` only when the developer would benefit from it.

### Consistency

I built this trying to keep consistency with JSON encoding / decoding that is already built in into classes. For example

- `ParseParcelDecoder` / `Encoder` works just like `ParseDecoder` / `Encoder` for JSON
- `ParseOperationSet` has `toRest(ParseEncoder)` and `fromRest(JSONObject, ParseDecoder)`. Added `toParcel(Parcel, ParseParcelEncoder)` and `fromParcel(Parcel, ParseParcelDecoder)`
- `ParseFieldOperation` interface has `encode(ParseEncoder)` and now `encode(Parcel, ParseParcelEncoder)`
- Same for `ParseFieldOperationFactory` for decoding

### Recursiveness

This now avoids `StackOverflowErrors` caused by nested ParseObject in the object tree. It’s pretty common for a `ParseObject` to be self referenced by one of its keys. This can be explicit (e.g. `object.put(“itself”, object)` or more obscure (e.g. `ParseRelation` is parceled when the parent object is parceled, but the relation itself holds a reference to the parent...).

This is solved by using stateful encoders / decoders called `ParseObjectParcelEncoder` and `ParseObjectParcelDecoder`. When the encoder traverses the objects / keys tree,
- it holds a reference to parceled objects
- when asked to parcel an object that was already parceled, it is parceled as a pointer (just class name and id)

Later, when the decoder parses the parcel,
- it holds a reference to unparceled objects
- when it finds a pointer, and an object with that id was already unparceled, it uses the same instance

This is pretty much what `KnownParseObjectDecoder` is already doing for JSON decoding, so there is still consistency with JSON stuff.
The required encoder / decoder instance is passed to every class so it can safely keep trace of actions.

### Ongoing tasks and LDS

By default this follows what was told in #122, that is, if there are ongoing save / save eventually tasks and we parcel, the tasks might succeed or fail but the unparceled object will act as if they failed.
The operation set queue is squashed to a single `ParseOperationSet`, that is parceled and restored, and contains the dirty changes from the last successful save at the moment of parceling. 

If local datastore is enabled **and** we manage to get the object from LDS memory when unparceling, we can have better behavior because the same instance is restored. In this case the original operation set queue is kept safe, and the unparceled object will be internally updated when save / delete tasks end.

What do you think of this approach?
I think there’s not much missing anymore.

### DIFF
--- a/Parse/src/main/java/com/parse/ParseObject.java
+++ b/Parse/src/main/java/com/parse/ParseObject.java
@@ -4234,10 +4234,6 @@ public class ParseObject implements Parcelable {
 
   @Override
   public void writeToParcel(Parcel dest, int flags) {
-    // TODO operationSetQueue?
-    // TODO isDeletingEventually?
-    // TODO warn if it has ongoing tasks
-
     ParseParcelableEncoder encoder = ParseParcelableEncoder.get();
     synchronized (mutex) {
       state.writeToParcel(dest);
--- a/Parse/src/main/java/com/parse/ParseObject.java
+++ b/Parse/src/main/java/com/parse/ParseObject.java
@@ -386,8 +386,10 @@ public class ParseObject implements Parcelable {
   private final ParseMulticastDelegate<ParseObject> saveEvent = new ParseMulticastDelegate<>();
 
   /* package */ boolean isDeleted;
+  /* package */ boolean isDeleting; // Since delete ops are queued, we don't need a counter.
   //TODO (grantland): Derive this off the EventuallyPins as opposed to +/- count.
   /* package */ int isDeletingEventually;
+  private boolean ldsEnabledWhenParceling;
 
   private static final ThreadLocal<String> isCreatingPointerForObjectId =
       new ThreadLocal<String>() {
@@ -2158,6 +2160,7 @@ public class ParseObject implements Parcelable {
     return toAwait.onSuccessTask(new Continuation<Void, Task<Void>>() {
       @Override
       public Task<Void> then(Task<Void> task) throws Exception {
+        isDeleting = true;
         if (state.objectId() == null) {
           return task.cast(); // no reason to call delete since it doesn't exist
         }
@@ -2167,6 +2170,12 @@ public class ParseObject implements Parcelable {
       @Override
       public Task<Void> then(Task<Void> task) throws Exception {
         return handleDeleteResultAsync();
+      }
+    }).continueWith(new Continuation<Void, Void>() {
+      @Override
+      public Void then(Task<Void> task) throws Exception {
+        isDeleting = false;
+        return null;
       }
     });
   }
@@ -4237,7 +4246,26 @@ public class ParseObject implements Parcelable {
 
   /* package */ void writeToParcel(Parcel dest, ParseParcelEncoder encoder) {
     synchronized (mutex) {
-      // Write className and id regardless of state.
+      // Developer warnings.
+      ldsEnabledWhenParceling = Parse.isLocalDatastoreEnabled();
+      boolean saving = hasOutstandingOperations();
+      boolean deleting = isDeleting || isDeletingEventually > 0;
+      if (saving) {
+        Log.w(TAG, "About to parcel a ParseObject while a save / saveEventually operation is " +
+            "going on. If recovered from LDS, the unparceled object will be internally updated when " +
+            "these tasks end. If not, it will act as if these tasks have failed. This means that " +
+            "the subsequent call to save() will update again the same keys, and this is dangerous " +
+            "for certain operations, like increment(). To avoid inconsistencies, wait for operations " +
+            "to end before parceling.");
+      }
+      if (deleting) {
+        Log.w(TAG, "About to parcel a ParseObject while a delete / deleteEventually operation is " +
+            "going on. If recovered from LDS, the unparceled object will be internally updated when " +
+            "these tasks end. If not, it will assume it's not deleted, and might incorrectly " +
+            "return false for isDirty(). To avoid inconsistencies, wait for operations to end " +
+            "before parceling.");
+      }
+      // Write className and id first, regardless of state.
       dest.writeString(getClassName());
       String objectId = getObjectId();
       dest.writeByte(objectId != null ? (byte) 1 : 0);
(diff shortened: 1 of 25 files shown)


Parse/src/main/java/com/parse/ParseObject.java
```diff
@@ -94,6 +98,14 @@ private static ParseObjectSubclassingController getSubclassingController() {
       return new Builder(className);
     }
 
+    /* package */ static State createFromParcel(Parcel source) {
+      String className = source.readString();
+      if ("_User".equals(className)) {
```
rogerhu: do we have this constant defined somewhere?

Parse/src/main/java/com/parse/ParseUser.java
```diff
@@ -1455,6 +1469,24 @@ public static void enableAutomaticUser() {
 
   //endregion
 
+  //region Parcelable
+
+  @Override
+  protected void onSaveInstanceState(Bundle outState) {
+    super.onSaveInstanceState(outState);
+    synchronized (mutex) {
+      outState.putBoolean("_isCurrentUser", isCurrentUser);
```
rogerhu: same here, should be a constant

Parse/src/main/java/com/parse/ParseObject.java
natario1: I think it’s left as _User everywhere, e.g. [a few lines above](https://github.com/ParsePlatform/Parse-SDK-Android/pull/625/files#diff-8e759e84d8f2c4f7fa6bc715e3786808R95).

Parse/src/main/java/com/parse/ParseUser.java
natario1: Yeah, I’ll fix this

rogerhu:
Are you talking about some type of recursive put?   Yeah there should be a check to make sure you're not putting an object against yourself.  I don't think Parcelable changes that fact (or might exacerbate it?)

In general I think the challenge with supporting this functionality is going to be all the inflight operations that occur.  I think the relative mechanics of implementing a parcelable interface is a lot of plumbing work.  We just have to make sure we  have in the docs about best practices with this stuff.

natario1:
Well not a recursive put, but recursive parceling because we parcel all dirty data for all objects, even child objects. The recursion might also be more hidden, e.g.

    user.setLastPost(post);
    post.put(“author”, user);
    // Then parcel post...

I wrote a similar test and it throws `StackOverflowError` when parceling. Any of you know how this is managed when JSON encoding ParseObjects?

Parse/src/main/java/com/parse/ParseObject.java
```diff
@@ -4237,7 +4246,26 @@ public void writeToParcel(Parcel dest, int flags) {
 
   /* package */ void writeToParcel(Parcel dest, ParseParcelEncoder encoder) {
     synchronized (mutex) {
-      // Write className and id regardless of state.
+      // Developer warnings.
+      ldsEnabledWhenParceling = Parse.isLocalDatastoreEnabled();
+      boolean saving = hasOutstandingOperations();
+      boolean deleting = isDeleting || isDeletingEventually > 0;
+      if (saving) {
+        Log.w(TAG, "About to parcel a ParseObject while a save / saveEventually operation is " +
```
rogerhu: Will it be easy to trigger these issues if people are navigating between activities or you have a lot of rotation-type changes?

natario1: I guess depends on developer habits, but generally this should be rare (a configuration change while the object is in the middle of a save).

I agree the message is too long, but I could not make it shorter... Maybe if we merge this and update docs, we can replace with `About to parcel a ParseObject with ongoing tasks. This might be dangerous. Read <link to docs>`?

rogerhu: We can do that later.  Would rather have a really big warning msg when this thing is first being used rather than a one-liner so we get some feedback about whether it happens a lot or not a ton.

natario1:
Now that this has been refined, it would be great to get the opinion of one of the original authors, beyond what was said in #122 .. Maybe we can summon @inlined ?

rogerhu:
@wangmengyan95 @grantland any thoughts?

grantland:
I haven't had time to go into the code in-depth, but the description seems like this has been well thought out. I haven't touched Parse code in about a year now, but the only thing that jumped out to me originally was the single instance use case for LDS which seems to have been covered.

natario1:
Glad to hear that @grantland , thank you for your time.

rogerhu:
rebase =)

natario1:
@rogerhu what do you think? I think we can merge this first, then I will just add a couple lines to #619 and #624

rogerhu:
rebase again

natario1:
@rogerhu done now

Issue #620, opened 2017-04-03
Allow specifying a custom OkHttp client


Jawnnypoo:
Since 1.14+ have a hard dependency on OkHttp, I think it would be great to expose the ability to specify the OkHttpClient to use. We would probably have to have them pass an OkHttpClient.Builder, so that we can add things to it as needed, and override things if necessary, but this would be great because we would then be exposing many things that were not previously exposed, such as HTTP timeouts, the ability to specify things like certificate pinning, and would eliminate the need for the class `ParseNetworkInterceptor`. 

What do you think @rogerhu ? I am happy to make a PR with this, keeping in mind backwards compatibility.

rogerhu:
Yes absolutely.  That has been my thought all along to solve many of those cert-related issues.

We should just make sure to keep the docs updated too: https://github.com/ParsePlatform/docs

rogerhu:
Agree in passing a builder instead of instance as well

Jawnnypoo:
Closed with https://github.com/parse-community/Parse-SDK-Android/pull/628

PR #628, opened 2017-04-06
OkHttp Builder Exposed for REST and AWS

Jawnnypoo:
As discussed in https://github.com/parse-community/Parse-SDK-Android/issues/620 this adds the ability to specify an OkHttpClient.Builder for the REST API as well as the AWS network calls.

To do this, you just pass it to your Parse.Configuration.Builder:
```
Parse.Configuration config = new Parse.Configuration.Builder(context)
            .awsClientBuilder(new OkHttpClient.Builder())
            .clientBuilder(new OkHttpClient.Builder())
            .build();
```

This involves removing `ParseNetworkInterceptor` due to the fact that it would require quite a bit of extra work to support both, and it would be unclear which interceptors are called first, etc. I think it best to just remove it. This will however require the next version to be something like 1.15, and require us to update some docs.

In the end, I think that this is needed since Parse servers can have all kinds of configurations, such as specific pinning, SSL configurations, etc that would be very cumbersome to support without exposing the internal OkHttpClient. 

@rogerhu Let me know what you think.


Parse/build.gradle
```diff
@@ -41,12 +41,17 @@ android {
 
 ext.okhttpVersion = '3.6.0'
 dependencies {
+    compile 'com.android.support:support-annotations:25.3.1'
```
rogerhu: the support version we use should be an ext { } definition likely in the root build.gradle.  this way the starter project that you updated can also just pull from this version

Parse/src/main/java/com/parse/Parse.java
```diff
@@ -158,72 +152,73 @@ public Builder server(String server) {
...
        * pinning.
+       *
        * @return The same builder, for easy chaining.
        */
       public Builder enableLocalDataStore() {
         localDataStoreEnabled = true;
         return this;
       }
 
-      /* package for tests */ Builder setNetworkInterceptors(Collection<ParseNetworkInterceptor> interceptors) {
```
rogerhu: can you confirm Stetho works fine with the new builder?  I'm trying to figure out why we needed a separate parse interceptor.  https://github.com/parse-community/ParseInterceptors-Android

Parse/src/main/java/com/parse/Parse.java
```diff
@@ -643,37 +629,24 @@ private static ParseEventuallyQueue getEventuallyQueue(Context context) {
     }
   }
 
-  static void checkInit() {
```
rogerhu: What's the reason for deleting?  Parse LiveQuery depends on this function to check to see if Parse has been initialized.

Parse/src/main/java/com/parse/ParseFileController.java
```diff
@@ -41,7 +41,7 @@ public ParseFileController(ParseHttpClient restClient, File cachePath) {
   /* package */ ParseHttpClient awsClient() {
     synchronized (lock) {
       if (awsClient == null) {
-        awsClient = ParsePlugins.get().newHttpClient();
+        awsClient = ParsePlugins.get().awsClient();
```
rogerhu: The awsclient is interesting to me... what is the use case for exposing it publicly (and separate from the OkHttpInstance builder)?

Parse/build.gradle
Jawnnypoo: Sounds good 👍

Parse/src/main/java/com/parse/Parse.java
Jawnnypoo: Hmm yeah I'll have to check on that, not a Stetho user myself

Parse/src/main/java/com/parse/Parse.java
Jawnnypoo: Ah, I saw that it was unused, and since it was not public, I figured it was safe to delete. I did not realize that ParseLiveQuery was of the same package name and would be accessing things that were package protected within Parse. I'll add it back, but it would probably be best to annotate or comment on things where this is the case

Parse/src/main/java/com/parse/ParseFileController.java
Jawnnypoo: Currently, the REST API uses a different HttpClient from the AWS client, and I was not really too sure of the reasons. Possibly just so that both could be making requests at the same time. Do you think they should use the same builder? I can make that the case, and they both just create their own client as needed if that sounds better.

natario1: Should we rename to fileClient? Files are not necessarily hosted on AWS anymore.

rogerhu: +1 for fileClient.

Let's use one for now until someone decides they need a separate one.  It was never exposed in the past.  I think the major reason is that there isn't a shared HttpClient because there is a parseClient() that is used for injecting Parse auth headers (see ParsePlugins.java)

natario1: Do we actually need to expose the fileClient builder? It does not need headers, and it just requests public file URLs that shouldn’t require any particular certificate / intercepting behavior.

If we use a single client, all requests to files will send useless (private) headers + any other burden that developer applies to the builder.. right?

Parse/src/main/java/com/parse/Parse.java
Jawnnypoo: Yeah, we are just going to have to remove support for that library and instead point users to using Stetho on their OkHttpClient Builder that they create for Parse http://facebook.github.io/stetho/

Somewhat unfortunate, but I think it is the best thing moving forward

flovilmart: Guys, be really careful when you abrubptly plan to deprecate a feature that is really useful for debugging/ troubleshooting.
Moreover, relying on an additional 3rd party library seems to be a step back from having an easy to use interceptor as this one.

rogerhu: I believe https://github.com/parse-community/ParseInterceptors-Android is redundant.  Before we upgraded the library to rely mostly on the OkHttp networking, this repo was used to provide a more generic interface for the different ways to handle networking.  

Most of the Android world relies on Stetho/HttpLogInterceptor that provides out of the box so I don't see a reason we need to continue maintaining ParseInterceptors (also it is not really maintained)

Jawnnypoo: @flovilmart I get your concern here, but OkHttp is pretty much standard for a library/application that is using HTTP. We are deprecating that interceptors library in favor of just having people use OkHttp interceptors for logging, which allows for more control and selection for interceptors (including logging ones) such as https://github.com/jgilfelt/chuck. Most people will probably be more familiar with setting up an OkHttp interceptor than a Parse specific one. Also, relying on a tried and true third party library is favorable in my opinion over having three different HTTP implementations to maintain/debug.

Parse/src/main/java/com/parse/ParseFileController.java
Jawnnypoo: Yeah, I will go ahead and have the HTTP and File clients be the same builder, just two different client instances to mock the present behavior. The reasoning for having them both be configurable would just be in case we have someone who has two different servers, one for Parse and one that hosts files, and it is possible though unlikely that they could be configured with different HTTPS configurations. But I am all for keeping it limited to the one builder for now.

rogerhu:
1 left:

```
com.parse.ParseCloudTest > testGetCloudCodeController FAILED
    java.lang.IllegalStateException: Another currentUser controller was already registered: com.parse.CachedCurrentUserController@1f197fa9
        at com.parse.ParseCorePlugins.registerCurrentUserController(ParseCorePlugins.java:146)
        at com.parse.ParseTestUtils.setTestParseUser(ParseTestUtils.java:39)
        at com.parse.ParseCloudTest.setUp(ParseCloudTest.java:47)
Gradle Test Executor 1 finished executing tests.

```

Parse/src/main/java/com/parse/Parse.java
flovilmart: Don't forget to update the README, wiki and docs repository then :)

Jawnnypoo:
@rogerhu This should be good now. I went ahead and made tests that were failing due to Plugins reset the plugins before and after test execution so that this hopefully wont be a problem for any more PRs

rogerhu:
Thanks this is awesome!  Going to merge.  

1. Can you draft an update to the Parse docs (https://github.com/parse-community/docs)?
2. We may need to put up an update on Parse network interceptors that it is no longer going to be compatible with the next 1.14.2 release.

Jawnnypoo:
@rogerhu Gotcha, I can get to that hopefully tomorrow. I for sure think this should be a 1.15.0 release though, since we are removing ParseInterceptors and it would be bad to remove a public class that people might depend on in a dot release. What do you think?

rogerhu:
Yep let's do that!

wangmengyan95:
I can add some context here. 

The reason we create Parse Interceptor is because we need to support three http libraries, eg apache, URLConnection and OKHttp. Since the first two libraries' API are so different, to make sure we do not make huge change to the controller level, we create a wrapper class ParseHttpClient and build the Parse Interceptor on top of it. The interceptor idea and mechanism are from OKHttp. If we only use OKHttp to power Parse, we can just use a standard OKHttp Interceptor.

The reason we need to support three different http libraries is because when we open source parse
1. There were still lots of users using < 4.4 os (I remembered more than 30%), which means we have to support the apache library. Based on the current data here https://developer.android.com/about/dashboards/index.html, only 10% users are under 4.4.
2. We are very conservative about adding any external dependency. So we just support OKHttp, but not require it.

I am fine with require OkHttp since it is actually the engine power the URLConnection in AOSP. If you guys need to make some API changes, make sure you clearly doc it and put it in the release notes. 

Thanks for the contribution!

flovilmart:
Thanks @wangmengyan95 for the clarification!

Issue #636, opened 2017-04-18
Unauthorized Errors


Jawnnypoo:
Updated to 1.15.0 and getting `unauthorized` errors on all endpoints. If I add a custom client builder it works fine. So, I think this is something to do with us not adding the proper interceptor if there are none already.

I think I have an idea what's going on, may require a quick 1.15.1 to patch. @rogerhu

PR #637, opened 2017-04-18
Use the right builder

Jawnnypoo:
Fixes https://github.com/parse-community/Parse-SDK-Android/issues/636

### DIFF
--- a/Parse/src/main/java/com/parse/ParsePlugins.java
+++ b/Parse/src/main/java/com/parse/ParsePlugins.java
@@ -121,7 +121,7 @@ class ParsePlugins {
                         return chain.proceed(request);
                     }
                 });
-                restClient = ParseHttpClient.createClient(configuration.clientBuilder);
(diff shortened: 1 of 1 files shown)


Jawnnypoo:
Pulled this in locally using jitpack and I can confirm this fixes it. @rogerhu Please take a look when you get a moment.

rogerhu:
Do we need the same for fileClient?

Jawnnypoo:
fileClient is actually not including the interceptor that the restClient is, which means it is not including the app id, client key etc. This was the case with 1.14.1 https://github.com/parse-community/Parse-SDK-Android/blob/1.14.1/Parse/src/main/java/com/parse/ParseFileController.java I guess file operations do not require those headers?

rogerhu:
Great.  I've released 1.15.1 to Maven (should take a few hours to sync).  In the meantime, I think you either need a custom OkHttpClient or simply use 1.14.1.  Thanks for the quick verification!

Issue #445, opened 2016-04-30
Trying to recall why we didn't want a GCM static dependency


inlined:
I started working on [ParseServer #1369](https://github.com/ParsePlatform/parse-server/issues/1369) and was trying to implement this the same way as we did v3--without any compile-time dependencies on the GMS library.

Between GCM v3 and v4, things have become much more sophisticated. Now there's the [InstanceId](https://developers.google.com/instance-id/) API, which is basically OAuth for the device instead of the user. If you want to send a GCM message to that device, you get the device to issue a "GCM" scope token. If I write this all from scratch, I'm going to have to write compatibility shims for [Messenger](http://developer.android.com/reference/android/os/Messenger.html) on older platforms, manage crypto keys, sign values, etc. There's a very non-zero chance I'll get some of this wrong, so I wanted to revisit our decision to not depend on the GMS libraries.

The two reasons I can recall are portability or file size, but neither of those are really valid. The GMS client libraries are themselves weakly bound to underly GMS service. It's actually totally possible for Fire Phones to implement some of the same services (though the client Android Manifest would have to explicitly trust the package name). The file size issue doesn't seem to be a big deal anymore either since we could require just `com.google.android.gms:play-services-gcm:8.4.0`

Side Note: With the GCM v4 upgrade, we might also want to consider using the device's Instance ID as the ParseInstallation.installationId for better record stability (since installations are already upserted by that field). I don't recall if there's still regexes anywhere that expect that to be a UUID though.

grantland:
IIRC the main reason was not wanting to break backwards compatibility. Developers who aren't using Maven will have to manually add another dependency or else they'll hit compile/runtime failures.

Aside from that, we didn't want to force developers to add unnecessary dependencies if they aren't utilizing GCM. This is especially annoying if this puts a developer over the dex limit and they have to deal with implementing multi-dex or having to wait for proguard stripping on every build.

We had discussed this many times and our ideal implementation was a fully decoupled shim library, similar to how we open sourced `ParseFacebookUtils` and `ParseTwitterUtils`. This would allow us to utilize the GCM library as well as solve the problems outlined above, however it was never much of a priority since it wasn't broken.

I had many conversations about fully decoupling our SDKs with @nlutsenko, @wangmengyan95, @richardjrossiii and @bnham, and maybe they can fill in any gaps that I forgot, but here is what I remember:

[...]

PR #638, opened 2017-04-18
Allow the GcmBroadcastReceiver to be subclassed

mmimeault:
You can only define one Broadcast Receiver to act as a GCM receiver.
In our case, we want to receive pushes from Parse and another provider. So I was going to write a Proxy that wrap both receivers.

But the Parse SDK prevent that by doing a strong validation of the Manifest. (https://github.com/parse-community/Parse-SDK-Android/blob/master/Parse/src/main/java/com/parse/ManifestInfo.java#L446-L448)
The GCM receiver defined need to be explicitly an instance of GcmBroadcastReceiver.
And that class can't be extended as the onReceive method is final.
(If the validations do not pass, the PushType is None and the registration will not happen) (https://github.com/parse-community/Parse-SDK-Android/blob/master/Parse/src/main/java/com/parse/ManifestInfo.java#L212-L214)

The current PR is a proposition.
But there are many solutions to my problem.
- Remove the final
- Remove the strong validation
- other suggestion

### DIFF
--- a/Parse/src/main/java/com/parse/GcmBroadcastReceiver.java
+++ b/Parse/src/main/java/com/parse/GcmBroadcastReceiver.java
@@ -11,31 +11,17 @@ package com.parse;
 import android.content.BroadcastReceiver;
 import android.content.Context;
 import android.content.Intent;
-
-import java.util.HashSet;
-import java.util.Set;
+import android.support.annotation.CallSuper;
 
 /**
  * @exclude
  */
 public class GcmBroadcastReceiver extends BroadcastReceiver {
 
-    private final Set<BroadcastReceiver> subReceivers = new HashSet<>();
-
     @Override
-    public final void onReceive(Context context, Intent intent) {
+    @CallSuper
+    public void onReceive(Context context, Intent intent) {
         ServiceUtils.runWakefulIntentInService(context, intent, PushService.class);
-        for (BroadcastReceiver broadcastReceiver : subReceivers) {
-            broadcastReceiver.onReceive(context, intent);
-        }
-    }
-
-    protected final void addBroadcastReceiver(BroadcastReceiver broadcastReceiver) {
-        subReceivers.add(broadcastReceiver);
-    }
-
-    protected final void removeBroadcastReceiver(BroadcastReceiver broadcastReceiver) {
-        subReceivers.remove(broadcastReceiver);
     }
 
 }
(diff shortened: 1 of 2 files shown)


natario1:
Just curious, how are you going to get the gcm receiver instance to call `add` on it?

mmimeault:
In the constructor of a class extending it. (protected method). You call add for each other receiver.

But I would prefer just remove the `final` on the onReceive signature.
I don't really understand why the SDK add that limitation.

One or the other is good for me.

natario1:
I would also rather remove the `final` and maybe add the `@CallSuper` annotation

Jawnnypoo:
@CallSuper is a part of the support annotations library, which Parse does not currently have as a dependency. But I would be happy if it did!

natario1:
I think we have it.. ? https://github.com/parse-community/Parse-SDK-Android/blob/master/Parse/build.gradle#L47

Jawnnypoo:
Ahh, I stand corrected. +1 for `@CallSuper` then!

natario1:
Not sure when it was introduced, but it would be great to also annotate the rest of the SDK (thinking mostly of @Nullable and @NonNull )

mmimeault:
Anyway... it doesn't work because of that
https://github.com/parse-community/Parse-SDK-Android/blob/master/Parse/src/main/java/com/parse/ManifestInfo.java#L368-L378

The class declared needs to be GcmBroadcastReceiver.class, not only an instanceOf.

Do anybody know why there is so much validation around that GCM receiver?
https://github.com/parse-community/Parse-SDK-Android/blob/master/Parse/src/main/java/com/parse/ManifestInfo.java#L212-L214

We are using mixpanel push system and we want also to use parse for differents campaign.
I would really like to have both working together.

rogerhu:
Speaking of which, does anyone have bandwidth to consider ripping out GCMService and putting them into a separate repo?  Sort of like ParseGCM like we have for Parse Facebook.

This way people can start using Firebase GCM v4 but this library would provide all the necessary setup to make things work with Parse (i.e. InstandID service)

See: https://github.com/parse-community/Parse-SDK-Android/issues/445

Jawnnypoo:
@rogerhu Do you think it needs to be its own repo?  Maybe it would be better to just have it as another module in this repo? (which could be deployed as its own dependency)

rogerhu:
I think the challenge is that when you deploy via Gradle, it will attempt to deploy across all modules unless you explicitly specify it.

If it makes things easier to get started, we can do it until we're ready to release..

rogerhu:
rebase

rogerhu:
one more rebase needed

racheal31ayers:
On Apr 27, 2017 1:04 PM, "Roger Hu" <notifications@github.com> wrote:

rebase

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/parse-community/Parse-SDK-Android/pull/638#issuecomment-297777261>,
or mute the thread

PR #640, opened 2017-04-20
Do not initialize when parse is already initialized

hermanliang:
If Parse is initialized in the application.
An IllegalStateException might be thrown when unit testing with Robolectric framework.

for example
MyApplication.java
```java
public class MyApplication extends Application {
  @Override
  public void onCreate() {
    super.onCreate();
    Parse.initialize(this);
  }
}
```

SomeTest.java
```java
@RunWith(RobolectricTestRunner.class)
@Config(constants = BuildConfig.class)
public class SomeTest {

  @Test
  public void test1() {
    ...
  }

  @Test
  public void test2() {
    ...
  }

}
```

Will get `IllegalStateException: ParsePlugins is already initialized` because of Robolectric setup application twice in the same testing session.

I think it would be better not to throw an exception when init parse more than once, just issue a warning log and do nothing.

It does not seems to be Robolectric's issue.
https://github.com/robolectric/robolectric/issues/595
https://github.com/robolectric/robolectric/issues/1981
https://github.com/robolectric/robolectric/issues/2456

### DIFF
--- a/Parse/src/main/java/com/parse/Parse.java
+++ b/Parse/src/main/java/com/parse/Parse.java
@@ -34,6 +34,8 @@ import okhttp3.OkHttpClient;
  * library.
  */
 public class Parse {
+  private static final String TAG = "com.parse.Parse";
+
   /**
    * Represents an opaque configuration for the {@code Parse} SDK configuration.
    */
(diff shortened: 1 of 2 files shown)


Jawnnypoo:
Are we positive that the error was not being throw intentionally there to prevent some other side effects of having it init twice? 

Side note: if you extend `ResetPluginsParseTest` it will fix this problem too.

hermanliang:
@Jawnnypoo I think the behavior is not changed because this PR is still not allowing parse to init more than once. 
BTW, `ResetPluginsParseTest` is in the SDK unit tests cannot be extended in application test class., but thanks anyway.

natario1:
Since this is an application/developer issue, maybe it’s better to do this in `Parse.initialize` ? Something like `if (isInitialized()) return`.

ParsePlugins is private, I think the exception might come useful when writing internal tests for the SDK (ensures everything is tear down correctly, you know where you’re failing etc).

hermanliang:
Thanks @natario1 
You're right, we should keep the behavior in `ParsePlugin`.
Checking in `Parse.initialized` is a good idea. 😄

PR #645, opened 2017-04-24
Fix encoding nested objects

natario1:
Adds a fundamental line to `ParseObjectParcelEncoder`. Without this there are issues when parceling objects with circular references


natario1:
Going to improve the test for this

PR #649, opened 2017-04-29
Remove some unused or unneeded proguard rules

Jawnnypoo:
Turns out we don't need a few of these as they were for when we were using a deprecated Notification class from API 23, and the SSLCertificateSocketFactory on other implementations of the Http Client.

In addition, I am removing the lines that keep source files and line numbers, as this should be decided on/configured by the end consumer, as they might not actually want those things to be kept in their end APK.


rogerhu:
rebase

PR #650, opened 2017-05-01
Make ParseSQLiteDatabase init safer

natario1:
The whole `ParseSQLiteDatabase` class assumes that `current != null`, however that does not look 100% safe because `current` is initialized asynchronously in the constructor.

This initializes `current` synchronously while keeping the same logic.

### DIFF
--- a/Parse/src/main/java/com/parse/ParseSQLiteDatabase.java
+++ b/Parse/src/main/java/com/parse/ParseSQLiteDatabase.java
@@ -63,15 +63,26 @@ import bolts.TaskCompletionSource;
     //TODO (grantland): if (!writable) -- do we have to serialize everything?
     openFlags = flags;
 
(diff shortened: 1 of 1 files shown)


rogerhu:
Where do you see this issue happening?

natario1:
I don’t know how frequently it happens, might be never. This currently relies on the fact that the `enqueue` part runs synchronously because the immediate executor is used. But when stack is deep the immediate executor can delegate to background threads and that would make this unsafe.

```java
  static Task<ParseSQLiteDatabase> openDatabaseAsync(final SQLiteOpenHelper helper, int flags) {
    final ParseSQLiteDatabase db = new ParseSQLiteDatabase(flags);
    return db.open(helper).continueWithTask(...);
  }
```

open() is called right after the constructor, and open requires `current` variable to be non null.

Anyway, this is low priority and it’s ok if you prefer not to change this class for now.

PR #651, opened 2017-05-02
Add missing User-Agent header.

aontas:
When moving to OKHTTP3, the User-Agent header was not added to the set of headers sent to the interceptor. It previously was there.

This simply re-adds it so that it is the same as it was before.


rogerhu:
Thanks somehow it got deleted in 6cfc4a0d8b57e508d8b3d801922138c9fa4461db

rogerhu:
Are there any other headers too?

aontas:
The HEADER_OS_VERSION also seems to be missing.

PR #653, opened 2017-05-03
Add missing HEADER_OS_VERSION header.

aontas:


aontas:
Added header I missed in #651.

rogerhu:
Thanks.

PR #655, opened 2017-05-04
Revert #648

natario1:
Reverting #648 because we lately discovered that the issue is much deeper and this was wrong.

### DIFF
--- a/Parse/src/main/java/com/parse/ParseRequest.java
+++ b/Parse/src/main/java/com/parse/ParseRequest.java
@@ -62,11 +62,6 @@ import bolts.TaskCompletionSource;
     return executor;
   }
 
-  /**
-   * This executor should be used for any network operation, in all subclasses.
-   * Chaining network operations with tasks consumed by Task.BACKGROUND_EXECUTOR , which is smaller,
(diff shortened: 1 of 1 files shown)


Parse/src/main/java/com/parse/ParseRequest.java
```diff
@@ -62,11 +62,6 @@ private static ThreadPoolExecutor newThreadPoolExecutor(int corePoolSize, int ma
     return executor;
   }
 
-  /**
-   * This executor should be used for any network operation, in all subclasses.
-   * Chaining network operations with tasks consumed by Task.BACKGROUND_EXECUTOR , which is smaller,
-   * will cause performance bottlenecks and possibly lock the whole SDK.
-   */
   protected static final ExecutorService NETWORK_EXECUTOR = newThreadPoolExecutor(
```
hermanliang: should also revert to package private?

rogerhu: It was never changed.

natario1: In the future I think it will make sense to move all custom executors (network executor, db executor) to the `ParseExecutors` class that exists already... but let’s fix this bolts issue first

PR #659, opened 2017-05-08
Bug fixes: add synchronized block to map operation in OfflineStore

hermanliang:


rogerhu:
Where did you see the bug?

hermanliang:
It's my fault. I not noticed this point when dealt with PR #607. It could be a potential problem with concurrent operations if map not synchronized. I think it's not serious because this logic is not often called.

PR #661, opened 2017-05-08
Allow multiple find/count operations on queries. Expose fromNetwork and isRunning

### LINKED ISSUE #6: Publicize `ParseQuery.fromNetwork()` & allow multiple `find()`/`count()`
`ParseQuery.fromNetwork()` would be useful for toggling a query `.fromLocalDatastore()` <-> `.fromNetwork()` to mimic `CACHE_THEN_NETWORK` and `CACHE_ELSE_NETWORK` from query caching.

We already have the private API, so it'd be super easy to enable this.

We'd also need to remove the limitation of mutating an inflight query, which should be fine since we now since `ParseQuery.State` is passed to the controller as an immutable object.

natario1:
This:

- Fixes #6 
- Exposes query.fromNetwork() so users can mimic cache policies when LDS is enabled
- Allows one query to perform multiple find() / count() operations concurrently
- Exposes query.isRunning() (this is a new API, if you think it’s not needed I can remove that)

### DIFF
--- a/Parse/src/main/java/com/parse/ParseQuery.java
+++ b/Parse/src/main/java/com/parse/ParseQuery.java
@@ -890,8 +890,8 @@ public class ParseQuery<T extends ParseObject> {
   private ParseUser user;
 
   private final Object lock = new Object();
-  private boolean isRunning = false;
-  private TaskCompletionSource<Void> cts;
+  private int isRunning = 0;
+  private TaskCompletionSource<Void> cts = new TaskCompletionSource<>();
 
   /**
    * Constructs a query for a {@link ParseObject} subclass type. A default query with no further
@@ -960,33 +960,21 @@ public class ParseQuery<T extends ParseObject> {
     return ParseUser.getCurrentUserAsync();
   }
 
-  private void checkIfRunning() {
-    checkIfRunning(false);
-  }
-
-  private void checkIfRunning(boolean grabLock) {
-    synchronized (lock) {
-      if (isRunning) {
-        throw new RuntimeException(
-            "This query has an outstanding network connection. You have to wait until it's done.");
-      } else if (grabLock) {
-        isRunning = true;
-        cts = Task.create();
-      }
-    }
-  }
-
   /**
    * Cancels the current network request (if one is running).
    */
   //TODO (grantland): Deprecate and replace with CancellationTokens
   public void cancel() {
     synchronized (lock) {
-      if (cts != null) {
-        cts.trySetCancelled();
-        cts = null;
-      }
-      isRunning = false;
+      cts.trySetCancelled();
+      cts = new TaskCompletionSource<>();
+      isRunning = 0;
+    }
+  }
+
+  public boolean isRunning() {
+    synchronized (lock) {
+      return isRunning > 0;
     }
   }
 
@@ -1030,8 +1018,6 @@ public class ParseQuery<T extends ParseObject> {
    * @see ParseQuery#fromPin(String)
    */
   public ParseQuery<T> setCachePolicy(CachePolicy newCachePolicy) {
-    checkIfRunning();
-
     builder.setCachePolicy(newCachePolicy);
     return this;
   }
@@ -1052,9 +1038,7 @@ public class ParseQuery<T extends ParseObject> {
    *
    * @see ParseQuery#setCachePolicy(CachePolicy)
    */
-  /* package */ ParseQuery<T> fromNetwork() {
-    checkIfRunning();
-
+  public ParseQuery<T> fromNetwork() {
     builder.fromNetwork();
     return this;
   }
@@ -1088,7 +1072,6 @@ public class ParseQuery<T extends ParseObject> {
    * @see ParseQuery#setCachePolicy(CachePolicy)
    */
   public ParseQuery<T> fromPin() {
-    checkIfRunning();
     builder.fromPin();
     return this;
   }
@@ -1105,7 +1088,6 @@ public class ParseQuery<T extends ParseObject> {
    * @see ParseQuery#setCachePolicy(CachePolicy)
    */
   public ParseQuery<T> fromPin(String name) {
-    checkIfRunning();
     builder.fromPin(name);
     return this;
   }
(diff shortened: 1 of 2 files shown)


natario1:
Any thoughts about this?

Parse/src/main/java/com/parse/ParseQuery.java
```diff
@@ -1052,9 +1038,7 @@ public CachePolicy getCachePolicy() {
    *
    * @see ParseQuery#setCachePolicy(CachePolicy)
    */
-  /* package */ ParseQuery<T> fromNetwork() {
-    checkIfRunning();
-
+  public ParseQuery<T> fromNetwork() {
```
Jawnnypoo: So, `fromNetwork` is just the opposite of `fromPin`?

Parse/src/main/java/com/parse/ParseQuery.java
```diff
@@ -960,33 +960,21 @@ public ParseQuery(ParseQuery<T> query) {
...
-        isRunning = true;
-        cts = Task.create();
-      }
-    }
-  }
-
   /**
    * Cancels the current network request (if one is running).
    */
   //TODO (grantland): Deprecate and replace with CancellationTokens
```
Jawnnypoo: Should we do what this comment suggests? Deprecate and replace with cancellation tokens?

Parse/src/main/java/com/parse/ParseQuery.java
natario1: Yes! Before this being exposed, if you had a `ParseQuery` and called `query.fromPin()` or `query.fromLocalDatastore()` there was no way to switch back to network from the same query object. This PR allows that and also allows calling `find()` multiple times from the same object.

Parse/src/main/java/com/parse/ParseQuery.java
natario1: I don’t know what that means.. We should overload the signature of find, findInBackground, count & countInBackground to allow a `CancellationToken` ?

Jawnnypoo: I think we would have it where each of those operations would return a cancellation token, which you could then use with `.cancel(token)` which would cancel that specific call. But, I think that goes beyond the scope here and can be addressed in a different PR if desired. 👍

natario1: I get your point, but find / count operations already have a return type, how would we pass back a `CancellationToken` the moment one calls find?

hermanliang: Now the cancellation is handled by the SDK itself. I think this suggestion means using the `CancellationTokenSource`/`CancellationToken` to transfer the cancellation handler to Bolt. But it is a big work, bunch of related files need to be modified. Another PR to do this job is preferred.

https://github.com/BoltsFramework/Bolts-Android#cancelling-tasks

natario1: On top of that , if they wanted to deprecate `query.cancel()`, they probably meant to have the developer get / pass a `CancellationToken` and call `source.cancel()` so the token is notified. See also the comment [here](https://github.com/parse-community/Parse-SDK-Android/blob/master/Parse/src/main/java/com/parse/ParseQuery.java#L1472).

The same comment is left in `ParseFile` e.g. [here](https://github.com/parse-community/Parse-SDK-Android/blob/master/Parse/src/main/java/com/parse/ParseFile.java#L709). We might set up a PR that addresses this in both classes. But the only way I see this possible is overloading methods, like

```java
CancellationTokenSource source = new CancellationTokenSource();
query.findInBackground(callback, /* nullable */ source.getToken());
// At any moment..
source.cancel();
```

It would be useful because `ParseFile` and with this PR `ParseQuery` allow multiple concurrent operations, so passing a token to each lets the developer cancel only one operation. If you think it’s needed I’m going to work on it (separately), with the plan

- overload methods to accept a token
- deprecate `cancel()` and have it fallback to `cancelAll()`
- create `cancelAll()` that cancels all currently running task

So the developer either passes a cancellation token to a single operation or uses `cancelAll` if he doesn’t care.

natario1:
I made this consistent with ParseFile so it will be easy if later we want to implement CancellationTokens.

Parse/src/main/java/com/parse/ParseQuery.java
hermanliang: I think it should be implemented internally first, and keep the developer's original behavior. Replace the `Task<Void> cancellationToken` in `ParseQueryController ` with  `CancellationToken cancellationToken`.

https://github.com/parse-community/Parse-SDK-Android/blob/master/Parse/src/main/java/com/parse/ParseQueryController.java

```java
interface ParseQueryController {
  <T extends ParseObject> Task<List<T>> findAsync(ParseQuery.State<T> state, ParseUser user,
      CancellationToken cancellationToken);
  <T extends ParseObject> Task<Integer> countAsync(ParseQuery.State<T> state, ParseUser user,
      CancellationToken cancellationToken);
  <T extends ParseObject> Task<T> getFirstAsync(ParseQuery.State<T> state, ParseUser user,
      CancellationToken cancellationToken);
}
```

Then re-implement all the relevant files and pass cancellationToken to Bolt Task methods (ex. [onSuccessTask](https://github.com/BoltsFramework/Bolts-Android/blob/master/bolts-tasks/src/main/java/bolts/Task.java#L837-L840)).

natario1: OK, I’ll do that just internally for queries and files then. 👍 going to open a separate PR though, do you think this can be merged?

hermanliang:
Does ParseQuery need a method to reset constraints  (ex.` query.reset()`)?
Because some query constraints are nonreversible (ex. `whereEqualTo(key, value)`).
It probably helpful when multiple find queries using different constraints.

natario1:
That should clear all constraints (order, where etc.) just like `new ParseQuery()` ? It would be great to have it per-key, `reset(key)` to undo a key constraint (maybe `revert` like parse object) but I’m not sure it’s possible. I’ll take a look.

hermanliang:
Yeah, I just think about the concept of "query group". Developers can use a single `ParseQuery` instance to do multiple query tasks with different constraints concurrently. Then, they can cancel all tasks using `query.cancel()` at specific app life cycle (ex. `onPause`).

PR #666, opened 2017-05-16
Set the required internet/network permissions in the SDK

hermanliang:
State `INTERNET` and `ACCESS_NETWORK_STATE` `use-permission` in the SDK level.
Therefore, no longer need to be declared in App AndroidManifest.xml.


natario1:
Makes sense to me!

Jawnnypoo:
Sounds good! People usually add these right away to a project anyways, so should be nice for those that don't 👍

rogerhu:
With AAR as the default format this will also help. :)

PR #667, opened 2017-05-17
Exclude extra files

rogerhu:
We can kill the JAR-related config section later

The original exclude pattern...not sure if we need to reintroduce it except for Manifest.

```
         exclude '**/R.class'
         exclude '**/R\$*.class'
         exclude '**/Manifest.class'
         exclude '**/Manifest\$*.class'
         exclude '**/BuildConfig.class'
     }
```
@hermanliang

### DIFF
--- a/Parse/build.gradle
+++ b/Parse/build.gradle
@@ -35,6 +35,10 @@ android {
         consumerProguardFiles 'release-proguard.pro'
     }
 
+    packagingOptions {
+        exclude '**/BuildConfig.class'
+    }
+
     lintOptions {
         abortOnError false
     }


hermanliang:
Nice, I think it can solve the conflict BuildConfig.class with LiveQuery's aar.
https://github.com/parse-community/ParseLiveQuery-Android/issues/60

rogerhu:
Hmm didn't help...BuildConfig is still in there.

rogerhu:
Actually that makes sense...we have a BuildConfig dependency in the code that I added to avoid having to update this constant all the time (https://github.com/parse-community/Parse-SDK-Android/blob/master/Parse/src/main/java/com/parse/ParseObject.java#L56)

May have to remove it so we don't need it.

Jawnnypoo:
Isn't it not supposed to be allowed in a Java/Android project to have two library modules that share the same package name though? I think I remember this causing problems for me in the past. Maybe we need to change ParseLiveQuery to have a different package name from this SDK?

PR #672, opened 2017-05-23
Turn these classes to public

rogerhu:
Just tried to change Parse Live Query package name but there actually classes that need to be made public in order to make this possible.  For now, revert to not using BuildConfig in the Parse SDK Android.

```
/Users/rhu/projects/ParseLiveQuery-Android/ParseLiveQuery/src/main/java/com/parse/livequery/Subscription.java:18: error: State is not public in ParseQuery; cannot be accessed from outside package
    private final ParseQuery.State<T> state;
                            ^
/Users/rhu/projects/ParseLiveQuery-Android/ParseLiveQuery/src/main/java/com/parse/livequery/Subscription.java:71: error: State is not public in ParseQuery; cannot be accessed from outside package
    /* package */ ParseQuery.State<T> getQueryState() {
                            ^
/Users/rhu/projects/ParseLiveQuery-Android/ParseLiveQuery/src/main/java/com/parse/livequery/ParseLiveQueryClientImpl.java:7: error: ParseDecoder is not public in com.parse; cannot be accessed from outside package
import com.parse.ParseDecoder;
                ^
/Users/rhu/projects/ParseLiveQuery-Android/ParseLiveQuery/src/main/java/com/parse/livequery/ParseLiveQueryClientImpl.java:9: error: ParsePlugins is not public in com.parse; cannot be accessed from outside package
import com.parse.ParsePlugins;
                ^
/Users/rhu/projects/ParseLiveQuery-Android/ParseLiveQuery/src/main/java/com/parse/livequery/ParseLiveQueryClientImpl.java:11: error: ParseRESTCommand is not public in com.parse; cannot be accessed from outside package
import com.parse.ParseRESTCommand;
                ^
/Users/rhu/projects/ParseLiveQuery-Android/ParseLiveQuery/src/main/java/com/parse/livequery/SubscribeClientOperation.java:5: error: PointerEncoder is not public in com.parse; cannot be accessed from outside package
import com.parse.PointerEncoder;
                ^
/Users/rhu/projects/ParseLiveQuery-Android/ParseLiveQuery/src/main/java/com/parse/livequery/Su`bscribeClientOperation.java:14: error: State is not public in ParseQuery; cannot be accessed from outside package
    private final ParseQuery.State<T> state;
                            ^
/Users/rhu/projects/ParseLiveQuery-Android/ParseLiveQuery/src/main/java/com/parse/livequery/SubscribeClientOperation.java:17: error: State is not public in ParseQuery; cannot be accessed from outside package
    /* package */ SubscribeClientOperation(int requestId, ParseQuery.State<T> state, String sessionToken) {
                                                                    ^
``

### DIFF
--- a/Parse/src/main/java/com/parse/ParseQuery.java
+++ b/Parse/src/main/java/com/parse/ParseQuery.java
@@ -297,7 +297,7 @@ public class ParseQuery<T extends ParseObject> {
     }
   }
 
-  /* package */ static class State<T extends ParseObject> {
+  /* package */ public static class State<T extends ParseObject> {
 
     /* package */ static class Builder<T extends ParseObject> {
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ before_install:
     - pip install --user codecov
 
 script:
-  - ./gradlew clean testDebugUnitTest jacocoTestReport --info
+  - ./gradlew clean testDebugUnitTest jacocoTestReport
 
 after_success:
   - ./gradlew coveralls
--- a/Parse/src/main/java/com/parse/ParseDecoder.java
+++ b/Parse/src/main/java/com/parse/ParseDecoder.java
@@ -26,7 +26,7 @@ import org.json.JSONObject;
  *
  * @see com.parse.ParseEncoder
  */
-/** package */ class ParseDecoder {
+/** package */ public class ParseDecoder {
 
   // This class isn't really a Singleton, but since it has no state, it's more efficient to get the
   // default instance.
--- a/Parse/src/main/java/com/parse/ParsePlugins.java
+++ b/Parse/src/main/java/com/parse/ParsePlugins.java
@@ -21,7 +21,7 @@ import okhttp3.OkHttpClient;
 import okhttp3.Request;
 import okhttp3.Response;
 
-class ParsePlugins {
+public class ParsePlugins {
 
     private static final String INSTALLATION_ID_LOCATION = "installationId";
 
--- a/Parse/src/main/java/com/parse/ParseRESTCommand.java
+++ b/Parse/src/main/java/com/parse/ParseRESTCommand.java
@@ -31,7 +31,7 @@ import bolts.Task;
 /**
  * A helper object to send requests to the server.
  */
-/** package */ class ParseRESTCommand extends ParseRequest<JSONObject> {
+/** package */ public class ParseRESTCommand extends ParseRequest<JSONObject> {
 
   /* package */ static final String HEADER_APPLICATION_ID = "X-Parse-Application-Id";
   /* package */ static final String HEADER_CLIENT_KEY = "X-Parse-Client-Key";
--- a/Parse/src/main/java/com/parse/PointerEncoder.java
+++ b/Parse/src/main/java/com/parse/PointerEncoder.java
@@ -14,7 +14,7 @@ import org.json.JSONObject;
  * Encodes {@link ParseObjects} as pointers. If the object does not have an objectId, throws an
  * exception.
  */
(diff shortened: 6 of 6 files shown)

### CODE REVIEW

**Comment by Jawnnypoo** [comment]
Does BuildConfig not get created if you don't reference it? I am thinking it still might get generated anyways.

**Comment by rogerhu** [comment]
Yeah I just noticed BuildConfig is still there. 

I think this is related to going to AAR only packages.

**Comment by rogerhu** [comment]
Changed the classes needed for ParseLiveQuery to be public...thoguhts on this?

**Comment by Jawnnypoo** [comment]
Sounds good, I think it would be good to document the classes to let others know they are not truly meant for public use, so that we can change them without having to worry about compatibility.

Something like `This class is public in order to allow for other Parse SDKs to work. Please note that the API of this class can change with any release` or something like that

**Comment by natario1** [comment]
Is there any other option? Some build magic to solve the `BuildConfig` issue? Just my opinion, but I would rather move the whole live query to this repo at this point, if it really needs these internals.

**Comment by rogerhu** [comment]
I don't see a way in the Gradle plugin to rename BuildConfig to avoid conflicts.   That would be my first choice...

**Comment by Jawnnypoo** [comment]
I don't think its a great idea to pull in the live query stuff into this SDK since many might not use it and would just be confused by it. 

I think there are many reasons besides duplicated `BuildConfig` as to why they should not share a package name. I think that is honestly the best solution, thought it would involve a major release in Live Query since it would break compatibility. 

You can try setting an applicationId in the build.gradle to a different package name and see if that generates a different BuildConfig in Live Query. But in the end, my vote goes toward once and for all just changing the package name of Live Query to prevent other future issues.

**Comment by natario1** [comment]
Yeah, well my comment was more like “This looks so bad to me that I would rather...”. But I don’t see any solution either. I also agree that the package should be changed, or live query pulled here.

**Comment by hermanliang** [comment]
Just submit a PR in ParseLiveQuery
https://github.com/parse-community/ParseLiveQuery-Android/pull/64

**Comment by Jawnnypoo** [comment]
Ah that's a good point, I think that would work great

**Comment by natario1** [comment]
So for the future, should we start working on allowing LiveQuery to use new, meaningful public APIs and later change the file structure as well? By this I mean create new public APIs here for what LQ needs. I am not confident with LQ, but looking at the first comment in this PR it requires:

- `ParsePlugins` to get application Id and client key: we can expose these in `Parse` class
- `ParseRestCommand` to get server url: expose in `Parse` class
- `ParseDecoder` + `PointerEncoder`: that’s JSON encoding/decoding. Depending on what they are used for (ParseObjects or what?) we can offer static methods in this repo that might be useful for any user
- `ParseQuery.State`: I guess that it must be made public. It is accessed with [`query.getBuilder().build()`](https://github.com/parse-community/ParseLiveQuery-Android/blob/da38bee581646c2293b37beb666a11af2854030c/ParseLiveQuery/src/main/java/com/parse/Subscription.java#L20) which looks dirty, so we could add something like `public ParseQuery.State freeze() { return builder.build(); }` to get a immutable state from a ParseQuery. If we make this state `Parcelable`, it might have a sense for the developer as well.

Things will be probably worse because the first comment here only lists classes to be made public, not methods.

**Comment by flovilmart** [comment]
We have pretty much the same issue with the iOS, most notably for serialization of queries and objects.

File: Parse/src/main/java/com/parse/ParseQuery.java [hunk 1]
```diff
@@ -297,7 +297,7 @@ private static void throwIfLDSEnabled(boolean enabled) {
     }
   }
 
-  /* package */ static class State<T extends ParseObject> {
+  /* package */ public static class State<T extends ParseObject> {
```
Jawnnypoo: @rogerhu Can we remove the `/* package */ lines since they would be irrelevant since the class is public?

Issue #279, opened 2015-12-03
LDS Performance Improvements


grantland:
This is a master issue for all things related to local datastore performance improvements.

- [x] Make LDS stop checkpointing and making copies of objects for mutable containers (#108)
- [ ] Make LDS query matching more efficient by not loading every object into memory for a specified pin.
- [ ] Make LDS use `ParseObjectState` and `ParseUserState` instead of `ParseObject` and `ParseUser`, to reduce the amount of locking required for every LDS operation.
- [ ] Improve LDS `ParseRelation` tracking.
- [ ] Update LDS database schema to allow for query matching in the database layer itself.

For anyone reaching this issue as a result of currently bad LDS performance here's our current recommendations:

- Keep your object schemas as small as possible - the faster it is to parse them from JSON, the better.
- Relations are currently inefficient. If you can avoid them, do so.
- Minimize usage of any values of the following classes:
   - `Collection` and its subclasses
   - `Map` and its subclasses
   - `ParseACL`
   - `ParseGeoPoint`
- Whenever possible, use smaller pins. Because LDS loads all objects in a single pin before evaluating a query, the smaller the pin is, the better.

See https://github.com/ParsePlatform/Parse-SDK-iOS-OSX/issues/32

Allsimon:
Is there any documentation concerning which version is the fastest ?
@FreudGit uploaded an interesting spreadsheet in ParsePlatform/Parse-SDK-iOS-OSX#32 and I'm wondering if performance improvement are shared between iOS/Android

swapnilgt:
@grantland 
After our application is scaling a bit, we are hitting the performance ceiling on the LDS. I have started seeing some activity lately on the Parse-Android-SDK (almost after an year). Are there any plans to resolve the performance issues with LDS in android. Anyways, I have the following questions regarding the performance tips mentioned above:

* When you say "Keep your object schemas as small as possible - the faster it is to parse them from JSON, the better.", what is the scale you are talking about?
* When you say avoid using "`Collection` and its subclasses" and "`Map` and its subclasses", you mean not to use these in the extended custom `ParseObjects` at java layer?
* "Whenever possible, use smaller pins", do you recommend to have less number of entries in a particular class that I have pinned or do you recommend that I can have thousands of records pinned for a particular class but pin then sets of, let's say, 100 records?

PR #673, opened 2017-05-25
New query condition to match all strings that starts with some other given strings

eduardbosch:
Hi to all! 🎉 

This is my first contribution to Parse, and I hope it could be useful for the community.

### The problem
I need a way to perform efficient searches in my database. I have researched, but I haven't found any way to perform efficient partial searches. The only way to perform efficient searches with Parse I think is by doing full text match or start with string matching.

I've read an article about performing keyword partial case insensitive. This solution is based on searching strings starting with another given strings which is really efficient with MongoDB. http://jam.sg/blog/efficient-partial-keyword-searches/

### The solution
My solution was to split the string field that I want to search, in different keywords. Then, I take each of this keyword and I create different keyword versions by striping the firsts characters of each keyword with a minimum of 3 characters.

Suppose that we have this collection:
```
{ name: "LöRem ípsÜM-dÓlôr" }
```

Then I create the keywords without diacritics (https://jsperf.com/diacritics/43) and in lowercase. To create the keywords, I split the original _name_ field by white space, comas, dots, dashes, etc:
```
{
    name: "LöRem ípsÜM-dÓlôr",
    nameSearch: ["lorem", "ipsum", "dolor"]
}
```

Finally I create all the keywords representations for partial search by matchings strings that starts with given strings:
```
{
    name: "LöRem ípsÜM-dÓlôr",
    nameSearch: [
        "lorem", "orem", "rem",
        "ipsum", "psum", "sum",
        "dolor", "olor", "lor
    ]
}
```

Now, I create an index for the "nameSearch" field, and I can do partial matches really efficient with this query:
```
{ nameSearch: { $all: [ /^\Qrem\E/, /^\Qpsum\E/ ] }
```

This increases the database size a little bit, but I think it's really worth it because it allows really good partial text searches. This is a good approach for searching short names, but not for searching in large amount of words. That will be better with full text matching and _containsAll_ condition I think.

But the problem comes that parse does not allow calling _ParseQuery.containsAll_ with regular expressions. So here it comes my feature.

### The feature
I've created a new query condition to match all strings that starts with the given strings. I think this is really useful to make the partial keyword matching explained above. 

This new condition is called _containsAllStartsWith_. Basically is a combination of _containsAll_ and _startsWith_ conditions. Given an array of strings, they are converted to regular expressions to match at the beginning of strings like in _startsWith_ and delivered to _containsAll_ condition.

The problem comes that the SDK sends the query as a JSON like this:
`{ nameSearch: { $all: [ { $regex: "^\Qrem\E"}, {$regex: "^\Qpsum\E" ] }`

This command is not supported by MongoDB, because $ is not allowed inside the $all condition. So I've updated the Parse Server to convert from regular expression JSON representation to a regular expression itself. `{ $regex: "^\Qrem\E"}` to `/^\Qrem\E/`.

So Parse Server converts from this:
`{ nameSearch: { $all: [ { $regex: "^\Qrem\E"}, {$regex: "^\Qpsum\E" ] }`
to this:
`{ nameSearch: { $all: [ /^\Qrem\E/, /^\Qpsum\E/ ] }`
which runs perfectly on MongoDB.

### Application use
To use this kind of search in an application it's really easy.

Suppose the user enters a text in a search field
`"lÒr Psû"`

The application should take this string and follow the same process to remove the diacritics and split as when creating the keywords. That will result in this:
`["lor", "psu"]`

Finally, that array should be delivered to _containsAllStartWith_ condition like this:
```
List<String> stringQueries = Arrays.asList("lor", "psu");
ParseQuery<> query = ...
query.containsAllStartsWith("nameSearch", stringQueries);
```

This call will end with a query in MongoDB like this:
`{ nameSearch: { $all: [ /^\Qlor\E/, /^\Qpsu\E/ ] }`

### Thanks!
I'm happy to try to contribute to this library. I hope that this can be useful for some other people.

Please, let me know if you want to know anything else 😃 

### Other SDK contributions and necessary changes
**Parse-SDK-JS**: https://github.com/parse-community/Parse-SDK-JS/pull/437
**parse server**: https://github.com/parse-community/parse-server/pull/3864


Parse/src/main/java/com/parse/ParseEncoder.java
```diff
@@ -120,6 +122,12 @@ public Object encode(Object object) {
         return ((ParseQuery.RelationConstraint) object).encode(this);
       }
 
+      if (object instanceof Pattern) {
+        JSONObject json = new JSONObject();
+        json.put("$regex", object.toString());
```
natario1: Would it be possible to avoid `Pattern`s and move this logic to `ParseQuery` just like all other $ constraints?

I mean using `ParseQuery.addCondition(key, “$regex”, string)`. That way we keep consistencies and leave encoder untouched.

natario1:
Left some comments but other than this, nice work IMO. Let’s see how the other PRs are received as well, I guess `parse-server` should be the first.

Parse/src/main/java/com/parse/ParseQuery.java
```diff
@@ -1699,6 +1701,31 @@ public void getInBackground(final String objectId, final GetCallback<T> callback
...
+   * Add a constraint to the query that requires a particular key's value to contain each one of
+   * the provided list of values entirely or just starting with given strings.
+   *
+   * @param key
+   *          The key to check. This key's value must be an array.
+   * @param values
+   *          The values that will match.
+   * @return this, so you can chain this call.
+   */
+  public ParseQuery<T> whereContainsAllStartsWith(String key, Collection<String> values) {
```
natario1: `ParseQuery.whereAllStartWith` ?

Parse/src/main/java/com/parse/ParseQuery.java
```diff
@@ -1699,6 +1701,31 @@ public void getInBackground(final String objectId, final GetCallback<T> callback
    * {@code ParseQuery}.
    * <p/>
    * This only works on keys whose values are {@link ParseObject}s or lists of {@link ParseObject}s.
```
eduardbosch: I thought so also when I copied the comment from `containsAll`, but was not sure to change it.

Thanks to comment this 👍  Someone could write a good documentation.

Parse/src/main/java/com/parse/ParseQuery.java
eduardbosch: At the beginning I was thinking about `containsAllStartingWith`. Maybe this could be a better approach.

natario1: It took me a while to understand what this does. Personally I think the cleanest and simplest wording is to keep `whereStartsWith`:

- `whereStartsWith(key, String prefix):` check that value in `key` (must be `String`) starts with prefix
- `whereStartsWith(key, Collection<String> prefixes)`: check that values in `key` (must be `Array`) start with prefixes in the given collection (implicit: all of the constraints must be satisfied)

Parse/src/main/java/com/parse/ParseEncoder.java
eduardbosch: I think that could be better to avoid update the ParseEncoder. I didn't researched a lot.

Now, I think could be better to update the `constraintsAllStartsWith` to:
```
  public ParseQuery<T> whereContainsAllStartsWith(String key, Collection<String> values) {
    checkIfRunning();

    ArrayList<KeyConstraints> startsWithConstraints = new ArrayList<>();
    for (String value : values) {
      KeyConstraints keyConstraints = new KeyConstraints();
      keyConstraints.put("$regex", buildStartsWithRegex(value));
      startsWithConstraints.add(keyConstraints);
    }

    return whereContainsAll(key, startsWithConstraints);
  }
```

This might auto compile as expected. If people prefer to revert the ParseEncoder changes, I'll update the PR 👍

eduardbosch:
I've updated this section with correct links.

### Other SDK contributions and necessary changes
**Parse-SDK-JS**: https://github.com/parse-community/Parse-SDK-JS/pull/437
**parse server**: https://github.com/parse-community/parse-server/pull/3864

Parse/src/main/java/com/parse/ParseQuery.java
eduardbosch: Could be a good name, but I think it's good to have the `All` word to know that all strings in collection must match. But I like your option too.

Let's see what other prefer.

Parse/src/main/java/com/parse/ParseQuery.java
natario1: Yeah, I think you can remove/reword both to their real meaning.

Parse/src/main/java/com/parse/ParseEncoder.java
natario1: It makes sense. Also since whereContainsAll is documented to work only with `ParseObject`s, you could call directly `builder.addCondition(key, "$all", startsWithConstraints)` and return this

eduardbosch: I've updated the function to use KeyConstraints.

I keep using `whereContsinaAll` because was not restricted to `ParseObject`s. The docs were wrong.

Parse/src/main/java/com/parse/ParseQuery.java
eduardbosch: Comments updated.

eduardbosch:
@natario1 I've added containsAllStartingWith compatibility for offline search in LDS, but obviously, it's terribly slow. I think should be here only for compatibility.

The code could be optimized a little bit to avoid some regexp calls, but I think it will continue being so slow.

I think that for offline search, the best approach is to use Full Text matching (whereEqualsTo) or at most, partial text matching with (whereContains). Even using this approach, LDS is really slow for searching in thousands of records as parse loads all the records into memory (#279) and performs the matches from Java code.

I think that Parse it not really designed for offline search of efficient partial text matching 😞

flovilmart:
It’s more an issue with mongodb than Parse. Regular expressions are very costly

eduardbosch:
Well, maybe this could be improved by saving ParseObjects in LDS in a way that data could be queried without having to load all data into memory and perform the matches in plain Java.

Is SQLite LIKE expression more efficient than performing the matches with plain Java? I don't know. Or maybe use another offline data management approach like Realm that it's really efficient in that.

You know better Parse SDK than me, and how it's built, and I'm sure that will be really difficult to implement a more efficient LDS data management for a huge amount of data. But that will make working with thousands of ParseObjects in LDS great. Now is a little bit a pain, but it works 👍 

Thanks for all your help as always!! 😃

eduardbosch:
Yeah, indexing will be great.

Yes, I've been looking deeply at Parse LDS SQLite db file and Parse offline store to know how data is saved and queried in LDS. I've seen that is saved as plain JSON, so the only way to perform queries right now is by loading into memory and perform the matching as it's done.

It's a pity that this makes a huge volume of data nearly unusable with some kind of queries in LDS. Maybe a good approach is to maintain another more efficient offline db that only saves the texts that could be searched with the associated ParseObject ids and its Class names so the search it's done outside of LDS and then we use the matching ParseObject ids to query into LDS.

flovilmart:
Yeah, or any list of keys that you wanna index for future searches. They could be pre-emptively set like: LDS.prepareStoreForQuery(storeName, query)

In this method we’d create a new table (if needed) with json objects as well as the columns extracted from the JSON. This way, you could use your query on that store and benefit from the indexing

eduardbosch:
Well, I think this approach will still be slow if there are thousands of records in `LDS.prepareStoreForQuery(storeName, query)`, right?

One nice approach will be to use an offline db like realm as LDS, but as you said, you need to know the schema, but maybe the user can create the schema and then integrate it automatically. That will be great.

Now I've tested the approach to save objects with `[String textForSearching, String parseObjectId, String parseClassName]` into realm and then query for objects that contain a given text inside `textForSearching` in the realm db. Then, with the matching results, I create `ParseObject.createWithoutData(parseClassName, parseObjectId)` and I fetch from local `fetchFromLocalDatastore()` and the results are instant, compared to 19 seconds if I use `ParseQuery.fromPin()` for aproximatelly 5k objects.

flovilmart:
Rhat’s Pretty much what I suggested, using a separate properly indexed storage for enduring fast queries and then resolving the objects either because they are stored in the optimized storage, or on the main pin. Wether you use SQLite or realm is an implementation detail at that point, I’m sure you could do the same with both

flovilmart:
And the prepareQuery would basically lookup the schema required for running the query the most efficiently because it knows the schema right?

eduardbosch:
I don't understand your last question @flovilmart sorry 😅

Do you mean that `prepareQuery` looks the query constraints and builds a schema that matches these constraints and saves objects with that conditions, so later on, when user performs the same query, it's done in this new table and not like is done right now loading all objects and performing the matches in Java?

flovilmart:
Thst’s The idea! Then the storage need probably an abstraction later do we can plug realm or SQLite :)

dplewis:
@rogerhu This was added in the latest version of parse-server. How does it look?

rogerhu:
Thanks for PR'ng.  Merging..

rogerhu:
rebased https://github.com/parse-community/Parse-SDK-Android/pull/839

eduardbosch:
Great! Thanks for merging @rogerhu and all the others that helped address this PR 😀

PR #675, opened 2017-06-01
Add client support for scheduling push notifications

### LINKED ISSUE #278: Add support for sending delayed push notifications
Sending push notifications from the REST API allows specifying `push_time`, which allows scheduling a push to be delivered to a device at a later date. This property is missing from our Android SDK and should be added for parity.

See:
- https://parse.com/docs/rest/guide#push-notifications-scheduling-pushes
- https://github.com/ParsePlatform/Parse-SDK-iOS-OSX/pull/610
- https://github.com/ParsePlatform/Parse-SDK-dotNET/pull/94

natario1:
This adds `ParsePush.setPushTime()` which is already present in other client SDKs.

Seems that `parse-server` will be eventually supporting scheduled pushes so we’ll be ready when it happens.

Fixes #278 .

### DIFF
--- a/Parse/src/main/java/com/parse/ParsePush.java
+++ b/Parse/src/main/java/com/parse/ParsePush.java
@@ -73,6 +73,7 @@ public class ParsePush {
             : new ParseQuery<>(new ParseQuery.State.Builder<ParseInstallation>(state.queryState()));
         this.expirationTime = state.expirationTime();
         this.expirationTimeInterval = state.expirationTimeInterval();
+        this.pushTime = state.pushTime();
         this.pushToIOS = state.pushToIOS();
         this.pushToAndroid = state.pushToAndroid();
         // Since in state.build() we check data is not null, we do not need to check it again here.
--- a/Parse/src/main/java/com/parse/ParsePush.java
+++ b/Parse/src/main/java/com/parse/ParsePush.java
@@ -55,6 +55,7 @@ public class ParsePush {
       private ParseQuery<ParseInstallation> query;
       private Long expirationTime;
       private Long expirationTimeInterval;
+      private Long pushTime;
       private Boolean pushToIOS;
       private Boolean pushToAndroid;
       private JSONObject data;
@@ -93,6 +94,18 @@ public class ParsePush {
       public Builder expirationTimeInterval(Long expirationTimeInterval) {
         this.expirationTimeInterval = expirationTimeInterval;
         expirationTime = null;
+        return this;
+      }
+
+      public Builder pushTime(Long pushTime) {
+        if (pushTime != null) {
+          long now = System.currentTimeMillis() / 1000;
+          long twoWeeks = 60*60*24*7*2;
+          checkArgument(pushTime > now, "Scheduled push time can not be in the past");
+          checkArgument(pushTime < now + twoWeeks, "Scheduled push time can not be more than " +
+              "two weeks in the future");
+        }
+        this.pushTime = pushTime;
         return this;
       }
 
@@ -151,6 +164,7 @@ public class ParsePush {
     private final ParseQuery.State<ParseInstallation> queryState;
     private final Long expirationTime;
     private final Long expirationTimeInterval;
+    private final Long pushTime;
     private final Boolean pushToIOS;
     private final Boolean pushToAndroid;
     private final JSONObject data;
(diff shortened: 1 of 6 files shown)


Parse/src/main/java/com/parse/ParsePush.java
```diff
@@ -96,6 +98,18 @@ public Builder expirationTimeInterval(Long expirationTimeInterval) {
         return this;
       }
 
+      public Builder pushTime(Long pushTime) {
+        if (pushTime != null) {
+          long now = System.currentTimeMillis() / 1000;
+          long twoWeeks = 60*60*24*7*2;
+          checkArgument(pushTime > now, "Scheduled push time can not be in the past");
+          checkArgument(pushTime < now + twoWeeks, "Scheduled push time can not be more than " +
```
Jawnnypoo: Is the two week restriction something imposed by the server?

natario1: Yes see [here](http://docs.parseplatform.org/rest/guide/#scheduling-pushes). Same code is in iOS version of this feature so it should be ok

Parse/src/main/java/com/parse/ParsePush.java
```diff
@@ -422,6 +442,14 @@ public void clearExpiration() {
   }
 
   /**
+   * Sets a UNIX epoch timestamp at which this notification should be delivered, in seconds (UTC).
+   * Scheduled time can not be in the past and must be at most two weeks in the future.
+   */
+  public void setPushTime(long time) {
```
hermanliang: Could it be `public void setPushTime(Date date)`? 
(`pushTime = date.getTime()/1000`)

or change the time unit to milliseconds to match Java's timestamp
(`pushTime = time/1000`)

Because unit of UNIX epoch timestamp is seconds, but developers probably using `Date.getTime()` or `System.currentTimeMillis()` to get timestamp (unit is milliseconds), then they will fail if not converted.

Parse/src/test/java/com/parse/ParsePushStateTest.java
```diff
@@ -151,6 +154,55 @@ public void testExpirationTimeIntervalNormalInterval() {
...
+
+    assertEquals(time, state.pushTime().longValue());
+  }
+
+  @Test(expected = IllegalArgumentException.class)
+  public void testPushTimeInThePast() {
+    ParsePush.State.Builder builder = new ParsePush.State.Builder();
+
+    ParsePush.State state = builder
+        .pushTime(System.currentTimeMillis() - 1000)
```
hermanliang: It is wrong if pushTime unit not changed as I previously commented.

Parse/src/main/java/com/parse/ParsePush.java
natario1: Yeah seconds are a terrible choice, but we already have `setExpirationTime(long)` and `setExpirationInterval(long)` in seconds, it would be confusing to have this in millis or in Date.
Note sure what to do, maybe we should add a `Date` version for all three methods. What do you think?

hermanliang: Oh yes, I didn't notice that. We should use the same time unit (seconds) for all relevant methods.
I think it's okay to add a `Date` version.

Parse/src/test/java/com/parse/ParsePushStateTest.java
hermanliang: time unit is not converted
`.pushTime(System.currentTimeMillis() / 1000 - 1)`

Parse/src/test/java/com/parse/ParsePushStateTest.java
```diff
@@ -151,6 +154,55 @@ public void testExpirationTimeIntervalNormalInterval() {
...
+        .data(new JSONObject())
+        .build();
+  }
+
+  @Test(expected = IllegalArgumentException.class)
+  public void testPushTimeTwoWeeksFromNow() {
+    ParsePush.State.Builder builder = new ParsePush.State.Builder();
+
+    ParsePush.State state = builder
+        .pushTime(System.currentTimeMillis() + 60*60*24*7*3)
```
hermanliang: `.pushTime(System.currentTimeMillis()/1000 + 60*60*24*7*3)`

natario1:
Ah, I didn't think about it 🙄 I just wanted to address a very old issue and get on par with iOS. It should do no harm though

flovilmart:
Yeah, people may think that 1 we can send push from the clien SDK’s without the master key, and 2 push scheduling is fully supported (only partially IIRC)

natario1:
Yeah, I think you remember correctly, my bad. I also didn't take it into account in the docs PR [here](https://github.com/parse-community/docs/pull/440/files#diff-08c4917989f0933cf095dd507f9ade73) where I just wanted docs to be on par with other clients SDK docs. Same story.

@flovilmart should we officially mark as deprecated all the `ParsePush.send()` stuff, or do you expect that some day there might be an option in parse-server that allows client pushing?

flovilmart:
Enabling client push is dangerous, and already was in 2014, the recommendation from the Parse team was to disable it: http://blog.parse.com/learn/engineering/the-dangerous-world-of-client-push/

PR #676, opened 2017-06-01
Refactor ParseFieldOperations and give its own class

natario1:
I see no reason why `ParseFieldOperation` and `ParseFieldOperations` should be two classes in the same java file. This has no benefit, looks bad practice, and sometimes makes the build fail (it happens to me when I click on the test folder to run all the tests, with error along the lines of `can’t find class ParseFieldOperations`).


Jawnnypoo:
Personally, I think we should at some point also remove all the `/* package */` comments at the beginning of source files, as it breaks the javadocs of the class. But that is another PR for another time

natario1:
Good point @Jawnnypoo , yes, let’s do that in a single PR for the whole SDK.

PR #677, opened 2017-06-01
A few updates

Jawnnypoo:
- Updates build tools, gradle 3.5, okhttp, android gradle plugin, and a few others
- Removes `provided` dependency for Stetho. This seems to have no purpose anymore
- Fetch support library from `maven.google.com` instead of the SDK installer

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,10 +12,9 @@ android:
   components:
     - tools
     - platform-tools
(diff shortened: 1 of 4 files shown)


Jawnnypoo:
That's one part of it. Google also mentioned at I/O that all the future updates to the support lib will only be available on maven.google.com

PR #678, opened 2017-06-02
Fix bug in count queries retrieving object data

### LINKED ISSUE #495: count operation fetches all the objects.
https://github.com/ParsePlatform/Parse-SDK-Android/blob/master/Parse/src/main/java/com/parse/ParseRESTQueryCommand.java#L65-L71

https://github.com/ParsePlatform/Parse-SDK-Android/blob/master/Parse/src/main/java/com/parse/ParseQuery.java#L1318

While encoding parameters to generate a http request for `count`, `setLimit(0)` is ignored. As a result, it fetches every object (at most 100 objects).

natario1:
This fixes #495 , an issue with `count()` queries that were fetching the whole objects as well as the count. I assume count queries will be much faster now.

### DIFF
--- a/Parse/src/main/java/com/parse/ParseRESTQueryCommand.java
+++ b/Parse/src/main/java/com/parse/ParseRESTQueryCommand.java
@@ -19,6 +19,15 @@ import java.util.Set;
 
 /** package */ class ParseRESTQueryCommand extends ParseRESTCommand {
 
+  /* package */ final static String KEY_ORDER = "order";
+  /* package */ final static String KEY_WHERE = "where";
+  /* package */ final static String KEY_KEYS = "keys";
+  /* package */ final static String KEY_INCLUDE = "include";
+  /* package */ final static String KEY_LIMIT = "limit";
+  /* package */ final static String KEY_COUNT = "count";
+  /* package */ final static String KEY_SKIP = "skip";
+  /* package */ final static String KEY_TRACE = "trace";
+
   public static <T extends ParseObject> ParseRESTQueryCommand findCommand(
       ParseQuery.State<T> state, String sessionToken) {
     String httpPath = String.format("classes/%s", state.className());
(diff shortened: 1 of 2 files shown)


Jawnnypoo:
Were you able to inspect the JSON response and verify only the counts being returned now?

natario1:
@Jawnnypoo yes, now the response is something like `{ count: 70, results: [] }`. Before you would get data for all the 70 objects, parsed and then dropped 😕

Jawnnypoo:
Wonder if this has been broken from day one? Good fix 👍

Parse/src/main/java/com/parse/ParseRESTQueryCommand.java
```diff
@@ -41,38 +50,41 @@
...
+    // parse-server supports it. Currently with our APIs, when counting, limit will always be 0,
+    // but that logic is in ParseQuery class and we should not do that again here.
+    int limit = state.limit();
+    if (limit >= 0) {
+      parameters.put(KEY_LIMIT, Integer.toString(limit));
     }
 
     if (count) {
-      parameters.put("count", Integer.toString(1));
+      parameters.put(KEY_COUNT, Integer.toString(1));
```
hermanliang: The default `limit` is `-1` if not set.
https://github.com/parse-community/Parse-SDK-Android/blob/master/Parse/src/main/java/com/parse/ParseQuery.java#L347
So, if there is no `limit` in the query (will be assigned `-1`), then there should be a logic that sets `limit` to `0`?
`parameters.put(KEY_LIMIT, Integer.toString(0));`

hermanliang: Actually, we don't need to fetch any objects when counting, right?
So, the the `limit' should always be set to `0` here.

Parse/src/main/java/com/parse/ParseRESTQueryCommand.java
```diff
@@ -41,38 +50,41 @@
...
+      parameters.put(KEY_INCLUDE, ParseTextUtils.join(",", includeds));
+    }
+
+    // Respect what the caller wanted for limit, even if count is true, because
+    // parse-server supports it. Currently with our APIs, when counting, limit will always be 0,
+    // but that logic is in ParseQuery class and we should not do that again here.
+    int limit = state.limit();
+    if (limit >= 0) {
+      parameters.put(KEY_LIMIT, Integer.toString(limit));
     }
```
hermanliang: I think it can be moved back to original logic.
If count is `true`, always set limit to `0`.

hermanliang: http://docs.parseplatform.org/rest/guide/#counting-objects

natario1: Your link is a good reference because it says

`With a nonzero limit, that request would return results as well as the count.`

So `parse-server` supports counting and finding altogether. We don’t currently support this with our exposed APIs, but we might want one day in the future.
So since `parse-server` supports it and `ParseRESTQueryCommand` is a low level interface to reach the API, I think it should allow any limit.
The important thing is that any call to `ParseQuery.count()` will use `limit = 0`. And that is already true https://github.com/parse-community/Parse-SDK-Android/blob/master/Parse/src/main/java/com/parse/ParseQuery.java#L1300 : 

`final State<T> state = copy.setLimit(0).build();`

So the logic `if (count) limit = 0` is already in the `count()` methods. The issue was that `ParseRESTQueryCommand` was ignoring `limit = 0` imposed by `ParseQuery`.

Parse/src/main/java/com/parse/ParseRESTQueryCommand.java
natario1: Right. But with our exposed APIs in ParseQuery limit is already 0 for `count` actions https://github.com/parse-community/Parse-SDK-Android/blob/master/Parse/src/main/java/com/parse/ParseQuery.java#L1300 see other comment

Parse/src/main/java/com/parse/ParseRESTQueryCommand.java
hermanliang: @natario1 Thanks for your explanation and I agree your opinion. Moving this logic out would be the minimum modification to fix this issue.

PR #684, opened 2017-06-21
Add withinPolygon to Query

dplewis:
https://github.com/parse-community/parse-server/pull/3866

I haven't used java in a while so any help will be great

### DIFF
--- a/Parse/src/main/java/com/parse/OfflineQueryLogic.java
+++ b/Parse/src/main/java/com/parse/OfflineQueryLogic.java
@@ -481,6 +481,23 @@ import bolts.Task;
   }
 
   /**
+   * Matches $geoWithin constraints.
+   */
+  private static boolean matchesGeoWithinConstraint(Object constraint, Object value)
+      throws ParseException {
+    if (value == null || value == JSONObject.NULL) {
+      return false;
+    }
+
+    @SuppressWarnings("unchecked")
+    HashMap<String, List<ParseGeoPoint>> constraintMap =
+      (HashMap<String, List<ParseGeoPoint>>) constraint;
+    List<ParseGeoPoint> points = constraintMap.get("$polygon");
+    ParsePolygon polygon = new ParsePolygon(points);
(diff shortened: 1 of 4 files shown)


Jawnnypoo:
@flovilmart Do we support this on iOS? Seems like something we would want parity on

flovilmart:
The PR is open on the iOS SDK as well, we're also trying to figure out how to keep track of SDK required changes properly (and even better automatically)

addisonElliott:
Why has this pull request not been merged?

natario1:
What's the status of this in other repos?

@dplewis would it be easy for you to add Local Datastore support? It should be a few lines additions in `OfflineQueryLogic` I think. There is also `OfflineQueryLogicTest` with samples. But I am not sure if it's an easy task, what do you think?

dplewis:
@natario1 @addisonElliott `withinPolygon` is in all other SDKs. For Local Datastore I need an efficient pointInPolygon function. I have some time to day to update the android sdk

natario1:
@dplewis that would be great! Thanks for the effort.

natario1:
@dplewis what do you think about refactoring whereWithinPolygon to accept a `ParsePolygon`? I think it would be better now, not sure if you already planned

dplewis:
@natario1 Do you mean something like this?

```
public ParseQuery<T> whereWithinPolygon(String key, Object value) {
    ParsePolygon polygon = null;

    if (value instanceof ParsePolygon) {
      polygon = (ParsePolygon)value;
    }
    if (value instanceof List) {
      polygon = new ParsePolygon((List<ParseGeoPoint>)value);
    }
    builder.whereGeoWithin(key, polygon.getCoordinates());
    return this;
  }
```

dplewis:
I will change it to support both. 

The query builder will still have to use a list of coordinates for now until this PR goes through https://github.com/parse-community/parse-server/pull/4067

natario1:
Great... thanks man!

We should add docs for this and #696 some day

flovilmart:
Can you open issues on the docs repo for it?

PR #692, opened 2017-07-14
Feature

oazya500:
when save Object with filed contains List off new ParseObject . is saved by tow sequence save . 

if the filed is relation with new ParseObject  it is fail save and show "unable to encode an association with an unsaved ParseObject" . so if you make it like List filed


Jawnnypoo:
Did you select the wrong branch here? What you described as an issue does not match up with the changes here (and none of the commits are by you)

PR #696, opened 2017-08-09
Add Type Polygon to Schema and PolygonContains to query

dplewis:
https://github.com/parse-community/parse-server/pull/3944

I've also added a containsPoint function that can be used for live queries.

ParseGeoPoint was missing equality and tests

### DIFF
--- a/Parse/src/main/java/com/parse/ParsePolygon.java
+++ b/Parse/src/main/java/com/parse/ParsePolygon.java
@@ -1,0 +1,228 @@
+/*
+ * Copyright (c) 2015-present, Parse, LLC.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+package com.parse;
+
+import android.location.Criteria;
+import android.location.Location;
+import android.os.Parcel;
+import android.os.Parcelable;
+
+import org.json.JSONArray;
+import org.json.JSONException;
+
+import java.util.Locale;
+import java.util.List;
+
+import bolts.Continuation;
+import bolts.Task;
+
+/**
+ * {@code ParsePolygon} represents a set of coordinates that may be associated with a key
+ * in a {@link ParseObject} or used as a reference point for geo queries. This allows proximity
+ * based queries on the key.
+ *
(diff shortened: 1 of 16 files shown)


Parse/src/main/java/com/parse/ParsePolygon.java
```diff
@@ -0,0 +1,228 @@
...
+            inside = !inside;
+        }
+    }
+    return inside;
+  }
+
+  /**
+   * Throws exception for invalid coordinates.
+   */
+  static List<ParseGeoPoint> _validate(List<ParseGeoPoint> coords) {
```
natario1: Could you remove the _? This can just be `private` I think.

Parse/src/main/java/com/parse/ParsePolygon.java
```diff
@@ -0,0 +1,228 @@
...
+   * Get coordinates.
+   */
+  public List<ParseGeoPoint> getCoordinates() {
+    return coordinates;
+  }
+
+  /**
+   * Get converts coordinate to JSONArray.
+   */
+  public JSONArray coordinatesToJSONArray() throws JSONException{
```
natario1: Could you make this package protected? We are not exposing JSONs anywhere in the SDK

Parse/src/main/java/com/parse/ParsePolygon.java
dplewis: I was thinking about using this for offline query in the future and wanted it to be exposed. I can still remove the _.

natario1: Sure, that's fine

natario1:
What do you mean by `TODO: Polygon`? What's missing?

dplewis:
I realized parse server doesn't support PFPolygon in PFConfig. I'll make a separate PR for it in the future. This PR is good 😊

natario1:
Looks good to me. I have no way to test offline behavior right now but the tests look good. I'd say let's merge this and move to #684 ?

@rogerhu

rogerhu:
great... go for it

sharadbhushank:
Is the below statement in the doc still valid?

> If more than 10MB of data is waiting to be sent, subsequent calls to #saveEventually() or ParseObject.deleteEventually() will cause old saves to be silently discarded until the connection can be re-established, and the queued objects can be saved.

PR #707, opened 2017-08-17
Include dependency & references shield in README

addisonElliott:
Add a dependency shield that determines if the dependencies in the repository are up to date. If not, the shield will indicate out of date and this will notify contributors that the libraries need to be updated.

In addition, a references shield is included which shows the number of projects that reference this project.

These two badges are included in the Android-SDK-iOS-OSX project and it is nice to stay consistent between the various projects of Parse. Badges are obtained via VersionEye.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,10 +78,10 @@ As of April 5, 2017, Parse, LLC has transferred this code to the parse-community
  [maven-link]: https://maven-badges.herokuapp.com/maven-central/com.parse/parse-android
  
  [dependencies-svg]: https://www.versioneye.com/java/com.parse:parse-android/badge.svg?style=flat-square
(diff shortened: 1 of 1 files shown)


rogerhu:
Now can you help bump our version too. :)

Rebase needed.

addisonElliott:
Sure thing, I can update the version. That should be a separate pull request, right?

Why is a rebase needed, I'm not sure I understand.

rogerhu:
Your READMe.md file probably conflicts with the version that I just merged in... there are some conflicts currently (see resolve conflicts button)

addisonElliott:
Thank you, I noticed the button right after I sent my message...

It should be good to go now.

Issue #711, opened 2017-08-18
Android O notification channels


iernie:
I couldn't find an issue about this, so I am making one.

Given that Android O is just around the corner, the library will need to be updated to use notification channels in order for it to work on the next version of Android, and I guess a way of configuring the name of that channel. If not, no push notifications will show up on Android O and up.

`NotificationCompat.Builder(Context context)` has even been deprecated in favor for `NotificationCompat.Builder(Context context, String channelId)`

Read more about notification channels here:
[https://developer.android.com/preview/features/notification-channels.html](https://developer.android.com/preview/features/notification-channels.html)

PR #713, opened 2017-08-19
Add Notification Channels configuration for Android O

iernie:
In response to my own issue #711 I tried to add the support myself.

I added the new meta-tag `com.parse.push.notification_channel` which defaults to `"parse_push"`.
If the Android version SDK is 26 or above, the channelId is set to the notification. The project is also compiled against SDK 26 in order to get the notification api.

I have no idea how to test this, so any help or pointers are appreciated.

### DIFF
--- a/Parse/src/main/java/com/parse/NotificationCompat.java
+++ b/Parse/src/main/java/com/parse/NotificationCompat.java
@@ -97,7 +97,7 @@ import android.widget.RemoteViews;
         }
       }
       if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
-        postJellyBeanBuilder.setChannelId(b.mNotificationChannel.getId());
+        postJellyBeanBuilder.setChannelId(b.mNotificationChannelId);
       }
       return postJellyBeanBuilder.build();
     }
@@ -128,7 +128,7 @@ import android.widget.RemoteViews;
     Bitmap mLargeIcon;
     int mPriority;
     Style mStyle;
-    NotificationChannel mNotificationChannel;
+    String mNotificationChannelId;
 
     Notification mNotification = new Notification();
 
@@ -196,8 +196,8 @@ import android.widget.RemoteViews;
     /**
      * Set the notification channel of the notification, in a standard notification.
      */
-    public Builder setNotificationChannel(NotificationChannel channel) {
-      mNotificationChannel = channel;
+    public Builder setNotificationChannel(String notificationChannelId) {
+      mNotificationChannelId = notificationChannelId;
       return this;
     }
 
--- a/Parse/src/main/java/com/parse/NotificationCompat.java
+++ b/Parse/src/main/java/com/parse/NotificationCompat.java
@@ -57,7 +57,7 @@ import android.widget.RemoteViews;
     @Override
     public Notification build(Builder builder) {
       Notification result = builder.mNotification;
-      NotificationCompat.Builder newBuilder = new NotificationCompat.Builder(builder.mContext);
+      NotificationCompat.Builder newBuilder = new NotificationCompat.Builder(builder.mContext, builder.mChannelId);
       newBuilder.setContentTitle(builder.mContentTitle);
       newBuilder.setContentText(builder.mContentText);
       newBuilder.setContentIntent(builder.mContentIntent);
@@ -95,6 +95,9 @@ import android.widget.RemoteViews;
           }
         }
       }
+      if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+        postJellyBeanBuilder.setChannelId(b.mChannelId);
+      }
       return postJellyBeanBuilder.build();
     }
   }
@@ -117,6 +120,7 @@ import android.widget.RemoteViews;
     private static final int MAX_CHARSEQUENCE_LENGTH = 5 * 1024;
 
     Context mContext;
+    String mChannelId;
 
     CharSequence mContentTitle;
     CharSequence mContentText;
@@ -138,8 +142,9 @@ import android.widget.RemoteViews;
      *      RemoteViews. The Context will not be held past the lifetime of this
      *      Builder object.
      */
-    public Builder(Context context) {
+    public Builder(Context context, String channelId) {
       mContext = context;
+      mChannelId = channelId;
 
       // Set defaults to match the defaults of a Notification
       mNotification.when = System.currentTimeMillis();
--- a/Parse/src/main/java/com/parse/NotificationCompat.java
+++ b/Parse/src/main/java/com/parse/NotificationCompat.java
@@ -18,6 +18,7 @@ package com.parse;
 
 import android.annotation.TargetApi;
 import android.app.Notification;
+import android.app.NotificationChannel;
 import android.app.NotificationManager;
 import android.app.PendingIntent;
 import android.content.Context;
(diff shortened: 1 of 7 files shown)


build.gradle
```diff
@@ -22,11 +22,11 @@ allprojects {
 }
 
 ext {
-    compileSdkVersion = 25
+    compileSdkVersion = 26
     buildToolsVersion = "25.0.3"
```
natario1: Should this be changed? Also in travis

iernie:
What I forgot to mention here is that I only set the channel id for the notification, and then it's up to the user to actually make a channel with that id in their app.

natario1:
So I'll leave you my own view on this... Currently I'm not comfortable with the facts that:

- developer must also call `notificationManager.createNotificationChannel(id)` to have this work
- the point of channels is that you can have multiple ones, but with manifest we practically only support a single channel

I think that:

- Manifest support should be removed. That makes sense only in a single-channel setup 
- The broadcast receiver should just have a `protected NotificationChannel getNotificationChannel(Context, Intent)`, defaulting to something like `new NotificationChannel("parse_push", "Push notifications", IMPORTANCE_DEFAULT)`
- After we call `getNotificationChannel()`, we should also care about creating the channel if it does not exist

This way developer can create different channels based on different pushes inside the Intent, with a reasonable default, and also it works out of the box because we create the channel ourselves. Developer also has the chance to customize the `NotificationChannel` with lights, vibration etc.

What do you guys think?

It would be nice to merge this soon since O is rolling out

iernie:
I agree with you @natario1. I've tried to modify my code based on your feedback.

natario1:
Thanks @iernie !

I like the design, but currently I think this would break for example [here](https://github.com/parse-community/Parse-SDK-Android/pull/713/files#diff-1d118da97676b8ac95f9176106e130e8R53). When API < O, you are passing around a null NotificationChannel, but the NotificationChannel does not exist in these platforms. Right?

I think you should ensure that any reference to NotificationChannel happens in the `if (version > O)` blocks... For example you could do

```java
    NotificationCompat.Builder parseBuilder = new NotificationCompat.Builder(context);
    parseBuilder.setContentTitle(title)
        .setContentText(alert)
        .set ...;
    if (androidVersion > O) {
        NotificationChannel channel = getNotificationChannel(); // Not called before O
        parseBuilder.setNotificationChannel(channel); // This will use the channel id
    }
```

Also [this call](https://github.com/parse-community/Parse-SDK-Android/pull/713/files#diff-fe851bae137007977bad1bfd2e548179R170) to `getNotificationChannel()` is redundant, we already call it from `getNotification()`

natario1:
Then in `ParseNotificationManager` you could just pass the notification and, if O, get the channel with `nm.getNotificationChannel(notification.getChannelId())` and create it

iernie:
`nm.getNotificationChannel(notification.getChannelId())` won't work because the NotificationManager hasn't created the notification channel yet. If I don't pass the channel to `ParseNotificationManager` the channel object is lost once the notification is built.

iernie:
What could be done is get NotificationManager and create the channel when you call `getNotificationChannel()`

iernie:
@natario1 I've made some changes. `getNotificationChannel()` is only called on SDK >= 26 and null is never returned. It also adds it to the NotificationManager immediately. So there is minimal change to the existing code.

Parse/src/main/java/com/parse/ParsePushBroadcastReceiver.java
```diff
@@ -261,6 +264,27 @@ protected void onPushOpen(Context context, Intent intent) {
...
+   * @return
+   *      The notification channel
+   */
+  @TargetApi(Build.VERSION_CODES.O)
+  protected NotificationChannel getNotificationChannel(Context context, Intent intent) {
+    NotificationManager nm = (NotificationManager) context.getSystemService(Context.NOTIFICATION_SERVICE);
+    NotificationChannel notificationChannel = new NotificationChannel("parse_push", "Push notifications", NotificationManager.IMPORTANCE_DEFAULT);
+
+    // Android doesn't create a new channel if the properties of the channel hasn't changed
+    nm.createNotificationChannel(notificationChannel);
```
natario1: Could you move the channel creation outside, maybe in `getNotification()`, right after we call `getNotificationChannel()` ? If someone overrides this method, the creation won't happen

iernie: Done

iernie:
I made the `createNotificationChannel()` method overrideable as well in case somebody wants to add multiple channels. Of course, they could just do that from within their app as well, but It's nice to have the option I guess.

Parse/src/main/java/com/parse/NotificationCompat.java
```diff
@@ -124,6 +128,7 @@ public Notification build(Builder b) {
     Bitmap mLargeIcon;
     int mPriority;
     Style mStyle;
+    NotificationChannel mNotificationChannel;
```
natario1: Sorry man, I just noticed this... I am not sure if this might throw at runtime on older APIs, even if not used. Could we just store `String mNotificationChannelId` here?

build.gradle
natario1: As before, shouldn't we build with something starting with 26? Here and in travis file... No?

Parse/src/main/java/com/parse/NotificationCompat.java
iernie: You are right again, this will crash. Sending just the Id is better.

build.gradle
iernie: I didn't want to change the build tools since I don't know what the routines are here on that, but I can change to 26.0.1

I could also upgrade the support library to 26.0.1, but then I'd have to raise the minSDK to 14, since they removed support for <14

natario1:
Any insight as to why tests are failing?

iernie:
Ahh! `Robolectric does not support API level 26.`

iernie:
The last 4 failing tests are because metaData is null for some reason in those tests after upgrading to robolectric 3.4.2. Any pointers on how to solve this are welcome as I am running out of ideas.

iernie:
I've tried everything I can think of to get the tests to work with robolectric 3.4.2 and API 26, but I don't feel like I have enough experience with the testing framework or robolectric to get this to work, so any help is appreciated here.

Now that Android Oreo is launched, it would be nice to get this merged :)

Jawnnypoo:
No tests were written to test out the Android O based functionality, so I would opt to just keep Robolectric on the version it is on for now.

iernie:
The problem is that if the compileSDK isn't 26 then the code fails to compile as NotificationChannel doesn't exist. When compileSDK is 26 then robolectric 3.3 fails because it doesn't support API 26. And with robolectric 3.4.2 I haven't been able to get the last 4 tests running.

natario1:
This is not going to fix anything, but you should also update the `ROBOLECTRIC_SDK` constant in the `TestHelper` class to 26. Not every test class is referencing it by the way.

iernie:
What... I swear after testing all kinds of combinations of configs and versions, I can't believe that this finally went through. Tests still run on API 25 and robolectric 3.3.2 which is the last 3.3.x version that supports API 26.

natario1:
Nice catch. Going to merge, thanks @iernie for bringing up this issue!

iernie:
@natario1 Happy to help! :)

Jawnnypoo:
Eventually I think it would be great to move away from testing with Robolectric and just running tests on the emulator. Robolectric causes lots of headaches with how it breaks things. But thanks for suffering through it, @iernie !

iernie:
Haha, yes I think I was on the brink of replacing the whole test framework with something else to get it to work. I found the documentation of Robolectric lacking. There has to be a better way like you said to just test on the emulator.

iernie:
@natario1 Just a small thing, but I think you chose the wrong PR in the release notes :)

natario1:
You are right, will fix :-)

iernie:
Thanks 😄

PR #716, opened 2017-08-22
Update to support 26, gradle 4.1. Changes min to 14

Jawnnypoo:
This should probably be a bump to 1.16.0, since the min version will be bumped from 9 to 14


natario1:
Is this reasonably safe? I mean, no one using 9-13 anymore?

iernie:
According to https://developer.android.com/about/dashboards/index.html#Platform API <10 && 11-14 has a usage of less than 0.1% while API 10 has a usage of 0.7%.

If Google deprecated API <14 in their support library I think it's safe to do the same. At some point you have to stop supporting outdated software.

natario1:
I think the same, other popular SDKs do the same, for instance Facebook SDK stands at 15. I was asking because right now this is not strictly needed, we are only using annotations that would still be working fine AFAIK.

On the other hand, the no-dependencies, support-everyone policy has probably done some harm to this repo in the past (thinking about the current push status). So yeah, fine for me.

We should update Parse-UI and other stuff as well

mchun:
This is how planned obsolescence works.

PR #717, opened 2017-08-23
Changed endpoint classes/_User to users

Kiwifed0r:
Change the old endpoint classes/_User to users in order to make it consistent.


Jawnnypoo:
Looks like this will be backwards compatible, right? It affects the sign up process?
 http://docs.parseplatform.org/rest/guide/#signing-up

Kiwifed0r:
@Jawnnypoo iOS and JavaScript use this endpoint for sign up, so this should be backwards compatible. It basically just fixes a legacy issue in the Parse-Android-SDK

PR #723, opened 2017-08-24
Check for JSONObject.NULL before casting to String [Fixes #209]

### LINKED ISSUE #209: ParseUser$State java.lang.ClassCastException: org.json.JSONObject$1 cannot be cast to java.lang.String
I updated sdk from 1.10.0 to 1.10.3, now on ParseUser.logOut() I have this error apparently at random.
happened on my development app as well as production app.
Everything was working fine before the upgrade.
if need anything more I will try to be helpful.

this is part of MyUser class

``` java

@ParseClassName("_User")
public class MyUser extends ParseUser {

//more methods

public synchronized String getUserId() {
        if (getObjectId() == null)
        try {
            save();
            } catch(ParseException e) {
            String tempUserId = preferences.getString("temp-user-id", "");
                if (tempUserId.isEmpty()) {
                    tempUserId = ("unknown-user" + randomNumber);
                    preferencesEditor.putString("temp-user-id", tempUserId).commit();
                }
                return tempUserId;
            }

        return getObjectId();
    }

public static MyUser getCurrentUser() {
        return (MyUser) ParseUser.getCurrentUser();
    }


public static void logOut() {
        unsubscribeToPush(getCurrentUser().getUserId());
        ParseUser.logOut();   //line 211
        getCurrentUser().setAppVersion();
        subscribeToPush(getCurrentUser().getUserId());
    }

private static void unsubscribeToPush(String userId) {
        ParsePush.unsubscribeInBackground("user_" + userId);
    }
}
```

```
com.mypackage E/AndroidRuntime: FATAL EXCEPTION: main
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime: java.lang.ClassCastException: org.json.JSONObject$1 cannot be cast to java.lang.String
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at com.parse.ParseUser$State.sessionToken(ParseUser.java:134)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at com.parse.ParseUser.logOutAsync(ParseUser.java:1000)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at com.parse.ParseUser.logOutAsync(ParseUser.java:991)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at com.parse.CachedCurrentUserController$4$1$1.then(CachedCurrentUserController.java:186)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at com.parse.CachedCurrentUserController$4$1$1.then(CachedCurrentUserController.java:179)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at bolts.Task$15.run(Task.java:825)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at bolts.BoltsExecutors$ImmediateExecutor.execute(BoltsExecutors.java:105)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at bolts.Task.completeAfterTask(Task.java:816)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at bolts.Task.continueWithTask(Task.java:628)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at bolts.Task.continueWithTask(Task.java:639)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at bolts.Task$13.then(Task.java:731)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at bolts.Task$13.then(Task.java:719)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at bolts.Task$15.run(Task.java:825)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at bolts.BoltsExecutors$ImmediateExecutor.execute(BoltsExecutors.java:105)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at bolts.Task.completeAfterTask(Task.java:816)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at bolts.Task.continueWithTask(Task.java:628)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at bolts.Task.continueWithTask(Task.java:603)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at bolts.Task.onSuccessTask(Task.java:719)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at bolts.Task.onSuccessTask(Task.java:709)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at bolts.Task.onSuccessTask(Task.java:743)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at com.parse.CachedCurrentUserController$4$1.then(CachedCurrentUserController.java:179)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at com.parse.CachedCurrentUserController$4$1.then(CachedCurrentUserController.java:176)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at bolts.Task$15.run(Task.java:825)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at bolts.BoltsExecutors$ImmediateExecutor.execute(BoltsExecutors.java:105)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at bolts.Task.completeAfterTask(Task.java:816)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at bolts.Task.continueWithTask(Task.java:628)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at bolts.Task.continueWithTask(Task.java:639)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at com.parse.CachedCurrentUserController$4.then(CachedCurrentUserController.java:176)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at com.parse.CachedCurrentUserController$4.then(CachedCurrentUserController.java:170)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at com.parse.TaskQueue.enqueue(TaskQueue.java:69)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at com.parse.CachedCurrentUserController.logOutAsync(CachedCurrentUserController.java:170)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at com.parse.ParseUser.logOutInBackground(ParseUser.java:955)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at com.parse.ParseUser.logOut(ParseUser.java:982)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at com.mypackage.nova_framework.user.MyUser.logOut(MyUser.java:211)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at com.mypackage.info_page.InfoPageActivity$5.onClick(InfoPageActivity.java:311)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at com.android.internal.app.AlertController$ButtonHandler.handleMessage(AlertController.java:166)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at android.os.Handler.dispatchMessage(Handler.java:99)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at android.os.Looper.loop(Looper.java:137)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at android.app.ActivityThread.main(ActivityThread.java:4441)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at java.lang.reflect.Method.invokeNative(Native Method)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at java.lang.reflect.Method.invoke(Method.java:511)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:784)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:565)
10-21 16:33:55.585 3764-3764/com.mypackage E/AndroidRuntime:     at dalvik.system.NativeStart.main(Native Method)
```

jhansche:
The same technique is already used in `ParseObject.getString()`, but because this is a ParseObject.State subclass, there is no getString() method. The documentation explicitly states that JSONObject.NULL may be contained in the data.

Before the change, the unit test fails with:
```
com.parse.ParseUserTest > testSessionTokenFromNull FAILED
    java.lang.ClassCastException: org.json.JSONObject$1 cannot be cast to java.lang.String
        at com.parse.ParseUser$State.sessionToken(ParseUser.java:139)
        at com.parse.ParseUser.getSessionToken(ParseUser.java:313)
        at com.parse.ParseUserTest.testSessionTokenFromNull(ParseUserTest.java:137)
```

### DIFF
--- a/Parse/src/main/java/com/parse/ParseDecoder.java
+++ b/Parse/src/main/java/com/parse/ParseDecoder.java
@@ -71,6 +71,10 @@ import org.json.JSONObject;
       return convertJSONArrayToList((JSONArray) object);
     }
 
+    if (object == JSONObject.NULL) {
+      return null;
+    }
+
     if (!(object instanceof JSONObject)) {
       return object;
     }
--- a/Parse/src/test/java/com/parse/ParseDecoderTest.java
+++ b/Parse/src/test/java/com/parse/ParseDecoderTest.java
@@ -74,6 +74,12 @@ public class ParseDecoderTest extends ResetPluginsParseTest {
   }
 
   @Test
+  public void testNull() {
+    Object object = ParseDecoder.get().decode(JSONObject.NULL);
+    assertNull(object);
+  }
+
+  @Test
   public void testParseFieldOperations() throws JSONException {
     JSONObject json = new JSONObject();
     json.put("__op", "Increment");


natario1:
I am looking at this and am trying to identify the issue... It would be better to catch the bug rather than do a check because there's a bug somewhere.

Session token is read only, if you attempt to call `object.put("sessionToken", ...)` the sdk will throw no matter the content. So JsonObject.NULL comes from the SDK itself and that's an error

jhansche:
Well, even ParseObject.mergeFromServer() may blindly insert a non-String object into that key -- and I'm sure there are plenty of other places as well that may do that.

We could add some debugging code to `ParseUser.State.Builder.put(String,Object)` to surface the root cause, but I haven't been able to reproduce the issue myself, so I'm not sure how long that debugging code would have to be in place before the root cause is identified.

natario1:
I have a strong feeling that something wrong is happing in [this class](https://github.com/parse-community/Parse-SDK-Android/blob/master/Parse/src/main/java/com/parse/ParseUserCurrentCoder.java).

But before fixing I am trying to write a failing test. If anyone would like to help investigating, it would be great.

jhansche:
@natario1 OK, I was able to reproduce the issue by using Charles Proxy to rewrite the server response of the `/users/me` API call.  The rewrite replaces the `"sessionToken":"r:..."` JSON string with `"sessionToken":null`, and 💥  the same crash mentioned in #209 happens immediately after the become() call:

```
java.lang.ClassCastException: org.json.JSONObject$1 cannot be cast to java.lang.String
    at com.parse.ParseUser$State.sessionToken(ParseUser.java:134)
    at com.parse.ParseUser.isAuthenticated(ParseUser.java:206)
```

I'll build in the put() override mentioned above, so I can trace it back to wherever the value is getting set.

Note that I do not think the class you linked to is relevant.  The first thing I notice is that ParseCurrentUserCoder's KEY_SESSION_TOKEN key name is `session_token`, which does not match the `sessionToken` key used elsewhere.

natario1:
@jhansche that's cool, but I don't think parse-server is ever returning null session tokens. Even automatic users have one, right?

natario1:
I really think it's the coder I mentioned. This reproduces the error.

```java
  @Test
  public void testEncodeDecodeWithNullSessionToken() throws Exception {
    ParseUser.State state = new ParseUser.State.Builder()
        .sessionToken(null)
        .build();
    ParseUserCurrentCoder coder = ParseUserCurrentCoder.get();
    JSONObject object = coder.encode(state, null, PointerEncoder.get());
    ParseUser.State.Builder builder =
            coder.decode(new ParseUser.State.Builder(), object, ParseDecoder.get());
    state = builder.build();
    assertNull(state.sessionToken());
  }
```

Going to set up a fix... Thanks for helping @jhansche

jhansche:
I added the `ParseUser.State.Builder.put()` override mentioned above, and this is the stack trace I get when that value is being put into the ParseUser.State map:
```
java.lang.IllegalArgumentException: Cannot receive NULL here
    at com.parse.ParseUser$State$Builder.put(ParseUser.java:103)
    at com.parse.ParseUser$State$Builder.put(ParseUser.java:66)
    at com.parse.ParseObjectCoder.decode(ParseObjectCoder.java:127)
    at com.parse.NetworkUserController$5.then(NetworkUserController.java:130)
    at com.parse.NetworkUserController$5.then(NetworkUserController.java:125)
```

jhansche:
OK, so you think it's when persisting to disk and then restoring from disk?  The docs for ParseCurrentUserCoder suggest it will only be used when LDS is not enabled.  I do use LDS, so I'm not sure if that's the same issue that my app is running into.

The problem with PCUC that you mentioned, is that it only removes the value if it was non-null, but it will leave the key if it did exist but contained a null-like value, like `JSONObject.NULL`

```
    String newSessionToken = json.optString(KEY_SESSION_TOKEN, null);
    if (newSessionToken != null) {
      userBuilder.sessionToken(newSessionToken);
      json.remove(KEY_SESSION_TOKEN);
    }
```

Similarly, the encode() method lets the super class encode everything into the JSONObject, and then only replaces the session token key if it was non-null, but will not remove the JSONObject.NULL if it had already been encoded that way.

jhansche:
With the same put() override and running your test, this is the stacktrace:
```
java.lang.IllegalArgumentException: Cannot receive NULL here
	at com.parse.ParseUser$State$Builder.put(ParseUser.java:104)
	at com.parse.ParseUser$State$Builder.put(ParseUser.java:66)
	at com.parse.ParseObjectCurrentCoder.decode(ParseObjectCurrentCoder.java:175)
	at com.parse.ParseUserCurrentCoder.decode(ParseUserCurrentCoder.java:120)
	at com.parse.ParseObjectCurrentCoderTest.testEncodeDecodeWithNullSessionToken(ParseObjectCurrentCoderTest.java:79)
```

But I think the documentation for ParseUserCurrentCoder is incorrect then, because the coder is still used even with local datastore enabled:

```
      FileObjectStore<ParseUser> fileStore =
          new FileObjectStore<>(ParseUser.class, file, ParseUserCurrentCoder.get());
      ParseObjectStore<ParseUser> store = Parse.isLocalDatastoreEnabled()
          ? new OfflineObjectStore<>(ParseUser.class, PIN_CURRENT_USER, fileStore)
          : fileStore;
      ParseCurrentUserController controller = new CachedCurrentUserController(store);
```
But OfflineObjectStore seems to only use the passed-in legacy store in order to migrate old data to the new offline store.

natario1:
I have created #724 . Yes, exactly... So you are experiencing this with LDS enabled?

jhansche:
Yes, my app is seeing the same crash with LDS enabled.

natario1:
OK, sorry, didn't read. I think this was the root cause of the problem. But the fact that you are experiencing it with LDS enabled is confusing. `OfflineObjectStore` contains the file store, but never writes to it

jhansche:
Another change that fixes it across all cases, without having to check it on the way out:

```
      @Override
      public Builder put(String key, Object value) {
        if (value == JSONObject.NULL) value = null;
        return super.put(key, value);
      }
```

What's the reason for ever keeping JSONObject.NULL?

flovilmart:
@natario1 we're seeing that issue too, randomly.

natario1:
@flovilmart do you have traces? There's another bug somewhere then. If it's very frequent we could merge this, it's just not elegant at all, like a try catch to hide a broken contract.

flovilmart:
Yeah, we’ll investigate more Monday, we don’t have other traces besides what’s there already. I’d like us not to merge this one as it’s pushing the dust under the carpet.

CoderickLamar:
Hi All,

I am running into this issue as well with LDS enabled.

Currently running Android SDK: 1.16.3

Here is my stack trace:

```
Fatal Exception: java.lang.ClassCastException: org.json.JSONObject$1 cannot be cast to java.lang.String
       at com.parse.ParseUser$State.sessionToken(ParseUser.java:144)
       at com.parse.ParseUser.getSessionToken(ParseUser.java:324)
       at com.parse.NetworkQueryController.findAsync(NetworkQueryController.java:34)
       at com.parse.OfflineQueryController.findAsync(OfflineQueryController.java:34)
       at com.parse.AbstractQueryController.getFirstAsync(AbstractQueryController.java:25)
       at com.parse.ParseQuery.getFirstAsync(ParseQuery.java:1291)
       at com.parse.ParseQuery.access$1900(ParseQuery.java:90)
       at com.parse.ParseQuery$5$1.then(ParseQuery.java:1283)
       at com.parse.ParseQuery$5$1.then(ParseQuery.java:1279)
       at bolts.Task$15.run(Task.java:917)
       at bolts.BoltsExecutors$ImmediateExecutor.execute(BoltsExecutors.java:105)
       at bolts.Task.completeAfterTask(Task.java:908)
       at bolts.Task.continueWithTask(Task.java:715)
       at bolts.Task.continueWithTask(Task.java:726)
       at bolts.Task$13.then(Task.java:818)
       at bolts.Task$13.then(Task.java:806)
       at bolts.Task$15.run(Task.java:917)
       at bolts.BoltsExecutors$ImmediateExecutor.execute(BoltsExecutors.java:105)
       at bolts.Task.completeAfterTask(Task.java:908)
       at bolts.Task.continueWithTask(Task.java:715)
       at bolts.Task.continueWithTask(Task.java:690)
       at bolts.Task.onSuccessTask(Task.java:806)
       at bolts.Task.onSuccessTask(Task.java:796)
       at bolts.Task.onSuccessTask(Task.java:830)
       at com.parse.ParseQuery$5.call(ParseQuery.java:1279)
       at com.parse.ParseQuery$5.call(ParseQuery.java:1276)
       at com.parse.ParseQuery.perform(ParseQuery.java:1151)
       at com.parse.ParseQuery.getFirstAsync(ParseQuery.java:1276)
       at com.parse.ParseQuery.getFirstInBackground(ParseQuery.java:1240)
       at com.parse.ParseQuery.getFirst(ParseQuery.java:1022)

```
All of this is originating from:

ParseQuery.getQuery(ELDLocationFile.class)
                                .getFirst();

flovilmart:
@jhansche We're jumping back on this one. 

> What's the reason for ever keeping JSONObject.NULL?

I don't believe we should ever have a JSONObject.NULL. How does it handle when you put a null value in the database?

does it interpoate to null?

natario1:
The source of the bug is probably a encode/decode cycle, in which a legit null value gets encoded to JsonObject.NULL for storing (more or less), and then it is not correctly decoded to null again.

I fixed the LDS-off decoder in #724 , I would expect a similar issue somewhere in LDS-on classes... But can't help more than this

jhansche:
It would help to be able to track down _how_ that JsonObject.NULL value gets into the map. Given that we see this crash on the way out anyway, is there any objection to crashing loudly in `put()` using code similar to what I mentioned in https://github.com/parse-community/Parse-SDK-Android/pull/723#issuecomment-324766303?
That should at least give us some stack traces to figure out where it's coming from.  Even if that's only temporary until we find the root cause and fix it.

flovilmart:
@jhansche as it stands now, we could harden the logic around the sessionToken setting, from null to check also for JSONObject.NULL. I'm not a big fan of preventing the put of JSONObject.NULL at large as it hides a bigger issue. I believe we're on the right track, with the serializer into SQLite. Perhaps at one point, because a key is missing (with anon users) the session token is stored as null and not 'undefined'. When unserializing from the LDS' this may cause the JSON.NULL to pop in.

CoderickLamar:
@flovilmart 
Our app has anon users disabled

jhansche:
@flovilmart the reason blocking it at `put()` would be helpful is because we'll have a stacktrace showing when/how/why the value is being put into the map in the first place.  As it is, the only stacktrace we have to go on currently is the CCE on the way out, and that is what is "hiding the bigger issue."  Rejecting the `put()` would surface _how_ that value is getting into the map in the first place, which will lead us to the root cause, and ultimately the right way to fix it.

Right now I see a lot of "maybe" and "might" and "perhaps", but no suggestions for **proving** any of those theories.

flovilmart:
Anyway we can instrument the code with a proper printStackTrace if it ever happens? Right now we don’t know what puts it in, wether it’a malformed parse server response or a serialization issue? I’d be comfortable with having it at the put, preventing it, and print a stacktrace that would be easily reported

jhansche:
Right, that's what I'm talking about. My suggestion was to throw an exception (which will of course include the stacktrace), but if you'd rather just log it, that's at least a way to make an informed decision.  Throwing the exception will be more noticeable (because it will crash) and it is more likely to get reported here with the stacktrace.  Simply logging the exception would require pairing the #209 crash, with an earlier stacktrace that got printed to the logs, and that might make it more difficult to track down the root cause.

CoderickLamar:
Can you direct me to the line number where you're suggesting we put it?

I can go ahead and push it through our app (since we haven't been able to repro this locally), and I can attempt to get it logged remotely to see what we come up with.

mmimeault:
Same for us (me and @flovilmart). @jhansche I am looking at the SDK right now, but I'm not totally sure where I would put this log/debug/crash to be able to prove who is saving this JSON.Null. Thanks for you help.

jhansche:
From https://github.com/parse-community/Parse-SDK-Android/pull/723#issuecomment-324753801
I suggested overriding `ParseUser.State.Builder.put()` to add the check:
```java.lang.IllegalArgumentException: Cannot receive NULL here
	at com.parse.ParseUser$State$Builder.put(ParseUser.java:104)
	at com.parse.ParseUser$State$Builder.put(ParseUser.java:66)
```

flovilmart:
I’ll also double check that the server never send out null in any case for that particular key.

CoderickLamar:
@flovilmart @jhansche Awesome guys... I'll continue to debug on our end and report back any results as soon as I know.

CoderickLamar:
@flovilmart I'm not so sure this is coming from the server. It might be, I'm not familiar enough with the parse android SDK, but here is what happens:

User logs in and everything is fine... all calls to parse are working just fine
User continues to use the app (could be for hours at a time). 

While this is happening, any time a user generates what we classify as an "event", we run a background thread to upload these. 

Sometimes the background thread will not be able to save because either there will be an invalid session token, or we get this error.

mmimeault:
I added a stacktrace logger into this method:
https://github.com/parse-community/Parse-SDK-Android/blob/master/Parse/src/main/java/com/parse/ParseObject.java#L978-L1052
(I pushed that before seeing @jhansche , but at the end it does the same, since I added my log just before the builder.put, but thanks)

So here is the stacktrace:
```
       at com.parse.ParseObject.mergeFromServer(ParseObject.java:1045)
       at com.parse.ParseObject.mergeREST(ParseObject.java:1202)
       at com.parse.OfflineStore$11$2.then(OfflineStore.java:669)
       at com.parse.OfflineStore$11$2.then(OfflineStore.java:666)
       at bolts.i$5.run(Task.java:872)
       at bolts.c$a.execute(BoltsExecutors.java:105)
       at bolts.i.c(Task.java:863)
       at bolts.i.a(Task.java:661)
       at bolts.i.a(Task.java:672)
       at bolts.i$3.a(Task.java:766)
       at bolts.i$3.then(Task.java:754)
       at bolts.i$6.run(Task.java:917)
       at bolts.c$a.execute(BoltsExecutors.java:105)
       at bolts.i.d(Task.java:908)
       at bolts.i.b(Task.java:715)
       at bolts.i.b(Task.java:690)
       at bolts.i.c(Task.java:754)
       at bolts.i.c(Task.java:778)
       at com.parse.OfflineStore$11.then(OfflineStore.java:666)
       at com.parse.OfflineStore$11.then(OfflineStore.java:626)
........
```

mmimeault:
Maybe what I could do is put the stacktrace logging where @jhansche suggested, so I would know who is trying to set a null session in the State.
I'm not sure this is normal to have a null session token.

flovilmart:
@mmimeault the session token can be null if the user was never saved.

jhansche:
@flovilmart @mmimeault  to be clear, the problem is not whether `null` gets set into the map.  The problem is when `JsonObject.NULL` is set, _instead of_ a proper `null`.  So setting null is fine, but it should log/crash/report/whatever if anyone tries to put `JsonObject.NULL` in the map.  Because that means the json null did not get decoded properly.

jhansche:
https://android.googlesource.com/platform/libcore/+/master/json/src/main/java/org/json/JSONObject.java#100

Looking at [ParseDecoder.decode](https://github.com/parse-community/Parse-SDK-Android/blob/master/Parse/src/main/java/com/parse/ParseDecoder.java#L74) I see:

```
    if (!(object instanceof JSONObject)) {
      return object;
    }
```

And because `JSONObject.NULL` is declared as an anonymous `Object{}` (or in other sources, as a final class `JSONObject.Null{}`), it is not a subclass of JSONObject, so it returns JSONObject.NULL directly.

So the problem is that it doesn't convert `JSONObject.NULL` to `null`.  It's a ParseDecoder bug.

flovilmart:
This is where we converge as for the core reason for the issue, @mmimeault has an app going live soon with extra tooling in there. let's see the results of that and try to release something before next week.

mmimeault:
Ok so well, your solution @jhansche sounds good.

I applied the same fixes to a production build with a lot of logging. It seem to be when you logout and then relaunch the app following that logout. The logout persist a null token as I can see. (validate me). I haven't see any other case where the token was set to null from the logs I got during the weekend.

The SDK explicitly encode `null` to `JSONObject.NULL`, but do not handle the decode part an return directly that object per
```
  if (!(object instanceof JSONObject)) {
    return object;
  }
```

So from what I understand, it is totally possible to have a null session token in a some particular case. Saving them a JSONObject.NULL will always result into a crash when trying to retrieve that particular token.

So I'm fine to go with your solution/implementation that is pending since August.

@flovilmart --^

PR #724, opened 2017-08-24
Fix ParseUserCurrentCoder bug

natario1:
Fixes a bug that would make session token be `JSONObject.NULL` instead of `null`, causing #209 .


jhansche:
Looks like another unit test is failing because of this now.

natario1:
@rogerhu the second there will just hide the bug when present, this should actually go to the point and fix it. 

I think we could merge this and then, if the bug is still present from some other source, consider the other

rogerhu:
rebase

natario1:
What's wrong with travis? We just merged a few comments. Tests pass on my machine.. 😕

flovilmart:
There seems to be an issue in the ParseInstallationTests

natario1:
Looks related to the travis worker time zone, let's see. It's strange that it showed up now

PR #732, opened 2017-09-02
Push handling cleanup

### LINKED ISSUE #729: [Android O] IllegalStateException: Not allowed to start service Intent
Hi,

From Android O, background tasks must call `Context.startForegroundService` instead of `Context.startService`  ,but in this case you could just extend the `PushService` to `JobIntentService` to solve this.

Could you take a look at this issue please?

Migration guide:
https://developer.android.com/about/versions/oreo/android-8.0-changes.html#back-all

### Model
```
Device: Google Pixel
OS: Android 8.0.0
SDK Version : 1.15.8
```

### Error log
```
Fatal Exception: java.lang.RuntimeException
Unable to start receiver com.parse.GcmBroadcastReceiver: java.lang.IllegalStateException: Not allowed to start service Intent { act=com.google.android.c2dm.intent.REGISTRATION flg=0x10 pkg=com.example.sample cmp=com.example.sample/com.parse.PushService (has extras) }: app is in background uid UidRecord{ac05be6 u0a187 TRNB idle procs:1 seq(0,0,0)}


Caused by java.lang.IllegalStateException
--
  | android.app.ContextImpl.startServiceCommon (ContextImpl.java:1505)
  | android.app.ContextImpl.startService (ContextImpl.java:1461)
  | android.content.ContextWrapper.startService (ContextWrapper.java:644)
  | android.content.ContextWrapper.startService (ContextWrapper.java:644)
  | com.parse.ServiceUtils.runIntentInService (ServiceUtils.java:37)
  | com.parse.ServiceUtils.runWakefulIntentInService (ServiceUtils.java:68)
  | com.parse.GcmBroadcastReceiver.onReceive (GcmBroadcastReceiver.java:23)
  | android.app.ActivityThread.handleReceiver (ActivityThread.java:3252)
  | android.app.ActivityThread.-wrap17 (Unknown Source)
  | android.app.ActivityThread$H.handleMessage (ActivityThread.java:1677)
  | android.os.Handler.dispatchMessage (Handler.java:105)
  | android.os.Looper.loop (Looper.java:164)
  | android.app.ActivityThread.main (ActivityThread.java:6541)
  | java.lang.reflect.Method.invoke (Method.java)
  | com.android.internal.os.Zygote$MethodAndArgsCaller.run (Zygote.java:240)
  | com.android.internal.os.ZygoteInit.main (ZygoteInit.java:767)


```

natario1:
Following #729 , this does not solve the issue but makes it easier to think of a fix, and also introduce FCM in the future.

- General clean up, renaming some classes
- PPNS is ripped out
- PushHandler interface for handling payloads of different providers (GCM, FCM)
- PushService acts as a router and streamlines payloads to handlers in a single thread

To fix 729 I would probably act on ServiceUtils to launch PushService either as a Service or as a job depending on API level


Parse/src/main/java/com/parse/ManifestInfo.java
```diff
@@ -173,14 +167,14 @@ private static boolean usesPushBroadcastReceivers() {
       intentsRegistered++;
     }
 
-    if (intentsRegistered != 0 && intentsRegistered != NUMBER_OF_PUSH_INTENTS) {
+    if (intentsRegistered != 0 && intentsRegistered != 3) {
```
rogerhu: why delete this constant?

natario1: I had my brain in cleanup mode :) and this looked already clear from the method implementation to me. I can add again.

I can also add some tests to the patch but first wanted to know if there's consensus over this being merged

rogerhu: PR works for me.  It looks like you removed PPNS while also refactoring to a PushHandler interface with an eye towards allowing the FCM handler?   The only thing is that FCM/GCM shouldn't co-exist so we may need to move hasAnyGcmSpecificDeclaration() out too as part of the interface?  (canHandlePush?)

natario1: That sounds good! I will do it. In the future we might even make the handler public so people can provide their adapters like parse server.

By the way, is parse-server ready to send FCM-like messages? AFAIK currently the handlers here need to update the ParseInstallation `pushType` to tell parse server adapter how to send the push. What happens when we start setting `pushType = 'fcm'` ?

rogerhu: Yeah parse server may need to be upgraded.  But the push tokens should be interoperable so hopefully t

rogerhu: FCM is already handled:

https://github.com/parse-server-modules/parse-server-push-adapter/blob/dfa75b51eb3fb909572ad7ecaa11f29a9ecc917f/src/ParsePushAdapter.js#L15

rogerhu: I do think that the deviceType (https://github.com/parse-community/Parse-SDK-Android/blob/6e8804cd721513dff78c68ea9b99f27830e41e3c/Parse/src/main/java/com/parse/ParseInstallation.java#L288) is what gets used over GCM/FCM right now..

See https://github.com/parse-server-modules/parse-server-push-adapter/blob/ef59a8c8eec303ad9547e02223765a3699cc5032/src/PushAdapterUtils.js#L20

natario1: The actual send implementation is still GCM based though? https://github.com/parse-server-modules/parse-server-push-adapter/blob/master/src/GCM.js I don't know if that's a problem and how we should manage

natario1:
This got bigger than I expected but seems solid to me. Makes it easy to fix #729 (I hope), and to add FCM (I hope).

I was very tempted to rip out [PushRouter](https://github.com/parse-community/Parse-SDK-Android/blob/master/Parse/src/main/java/com/parse/PushRouter.java) and [PushHistory](https://github.com/parse-community/Parse-SDK-Android/blob/master/Parse/src/main/java/com/parse/PushHistory.java) but resisted, though I'm not 100% sure they are still useful. I hope this can be done in the future.

I also think we could add `PushService` to our own manifest to ease the developer set-up. The developer would still be able to choose not to support push, by not registering any `ParsePushBroadcastReceiver`. But I don't know what would happen when merging manifests that both have a push service declared, so I gave up.

rogerhu:
Thanks for doing this!

PR #735, opened 2017-09-10
Use JobScheduler for push

### LINKED ISSUE #729: [Android O] IllegalStateException: Not allowed to start service Intent

natario1:
Wraps PushService initialization in the `PushServiceUtils` class. On Android Oreo, it uses the Job scheduler APIs to launch `PushService26`. This is already declared in our manifest so no changes for developers.

FIxes #729 

- As far as I know `PushService26` being a declared service must be public. Correct me if I am wrong
- Can rename to `PushServiceOreo` if you think it's better
- As far as I know JobScheduler APIs are not testable with Robolectric (it won't schedule), we would need androidTests
- Unit tests aside, if someone could test on real setup that would be great


Parse/src/main/AndroidManifest.xml
```diff
@@ -13,5 +13,11 @@
     <uses-permission android:name="android.permission.INTERNET"/>
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE"/>
 
-    <application />
+    <application>
+        <service
+            android:name=".PushService26"
```
rogerhu: PushServiceApi26?

natario1: Yeah better.

Parse/src/main/java/com/parse/PushService.java
```diff
@@ -103,45 +156,26 @@
     }
   }
 
-  /* package */ static void unregisterServiceLifecycleCallbacks(
-      ServiceLifecycleCallbacks callbacks) {
+  /* package */ static void unregisterServiceLifecycleCallbacks(ServiceLifecycleCallbacks callbacks) {
     synchronized (PushService.class) {
       serviceLifecycleCallbacks.remove(callbacks);
-      if (serviceLifecycleCallbacks.size() <= 0) {
```
rogerhu: any reason for getting rid of this block?

Parse/src/main/java/com/parse/PushServiceApi26.java
```diff
@@ -0,0 +1,116 @@
...
+    return did == JobScheduler.RESULT_SUCCESS;
+  }
+
+  // We delegate the intent to a PushHandler running in a streamlined executor.
+  private ExecutorService executor;
+  private PushHandler handler;
+
+  // Our manifest file is OK.
+  static boolean isSupported() {
+    return true;
```
rogerhu: should you check whether the job service is declared?

Parse/src/main/java/com/parse/PushService.java
natario1: Yeah I just think it made no sense to null this. These ServiceLifecycleCallbacks are made just for internal use in tests (PushServiceTest) so we can design them as we wish, I think the old logic was way too verbose

Parse/src/main/java/com/parse/PushServiceApi26.java
natario1: The original PushService has a check here because it must be declared by the developer in each app. With PushServiceApi26 I have declared it in our own manifest, which should be merged with the application manifest when building the app. So I guess that's a true.

PR #745, opened 2017-09-20
Better setup / teardown for JobService

### LINKED ISSUE #744: Crash in PushServiceApi26
Hi

I know you guys have been working on this service lately and wanted to let you know that I am experiencing a crash when getting push on API 26.

The push notification did arrive on my phone, I did see it and could click it. I couldn't see any error from my phone on that regard, but I did get this message in Fabric immediately after.

```
Fatal Exception: java.lang.NullPointerException: Attempt to invoke interface method 'void com.parse.PushHandler.handlePush(android.content.Intent)' on a null object reference
       at com.parse.PushServiceApi26$1.run(PushServiceApi26.java:92)
       at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1162)
       at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:636)
       at java.lang.Thread.run(Thread.java:764)
```

Good thing is that you don't notice this crash as a user, but would be nice to see what the issue is. SDK 1.16.2.

natario1:
We don't rely anymore on `onCreate` / `onDestroy` to set up and tear down the job service. The service executes jobs in a separate thread so onDestroy is called before we would expect.

Fixes #744

### DIFF
--- a/Parse/src/main/java/com/parse/PushServiceApi26.java
+++ b/Parse/src/main/java/com/parse/PushServiceApi26.java
@@ -25,6 +25,7 @@ import android.util.SparseArray;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 
@@ -44,7 +45,8 @@ public final class PushServiceApi26 extends JobService {
     // Execute in the next second.
     Bundle extra = new Bundle(1);
     extra.putParcelable(INTENT_KEY, intent);
-    int did = scheduler.schedule(new JobInfo.Builder(JOB_SERVICE_ID, new ComponentName(context, PushServiceApi26.class))
+    ComponentName component = new ComponentName(context, PushServiceApi26.class);
+    int did = scheduler.schedule(new JobInfo.Builder(JOB_SERVICE_ID, component)
         .setMinimumLatency(1L)
         .setOverrideDeadline(1000L)
         .setRequiresCharging(false)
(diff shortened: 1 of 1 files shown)


Jawnnypoo:
Maybe I am missing something, but it looks like no matter what, in the `finally` block, we are saying the job is done and calling `jobFinished(jobParameters, false);`. Shouldn't this `true/false` be dependent on if the push was a success or not? Otherwise, the push could fail, and we would never reschedule the push to retry

natario1:
@Jawnnypoo by success you mean that `handler.handlePush()` does not throw?

Jawnnypoo:
Yeah, it seems like there, we should be checking to see if an exception was thrown by the handler, and if so, reschedule. But maybe that is incorrect thinking, since more than likely the only reason a push would fail would be cases where it would always fail, and it should therefore not be rescheduled.

natario1:
I got it, I think your second scenario makes sense. handler.handle() never reaches developer code, it is our own stuff. So if that throws there's something wrong with our handlers and we would end in a schedule loop

Jawnnypoo:
Makes sense 👍

Jawnnypoo:
Is there something we can do about this codecov thing?

natario1:
@Jawnnypoo not easily. I had made a `PushService26Test` for one of these PRs, but then discarded because it requires Oreo, and we currently can't run Oreo tests due to Robolectric issues. I don't remember the exact reason right now but I can take a look

natario1:
I have been playing with `androidTest`s recently, that would solve a couple issues, but running them on Travis has been quite painful

rogerhu:
I'm seeing issues in the current branch...sometimes the push handler crashes because handler is null.

Jawnnypoo:
@rogerhu Are you seeing this in the tests, or in an app using the latest?

rogerhu:
In the app.  Let's merge and tag release.  I will do more digging to understand what the issue is.

PR #749, opened 2017-10-10
Bump to 1.16.3

rogerhu:
Apparently my admin access got taken away too... this is so annoying.

@natario1 @Jawnnypoo


flovilmart:
@rogerhu what’s the issue?

montymxb:
We changed the teams up, removing the other android team. We just need to regrant Admin instead for the new single android team. I'll do that real quick...sorry @rogerhu !

rogerhu:
Thanks for the quick fix!

Issue #679, opened 2017-06-03
Organizing directories


natario1:
I’d like to address #98 , but before doing that, I’d like to discuss file/cache storage as it currently is.
Most of all I would like to get rid of the deprecated `Parse.getParseDir()` which is still used here and there. There are:

- `Parse.getParseDir()` annotated as deprecated, and I agree
- `Parse.getParseCacheDir()`
- `Parse.getParseFilesDir()`

Here’s where we use them:

- `getParseDir() + "/LocalId"`: used by LocalIdManager
- `getParseDir() + "/CommandCache"`: used by `ParseCommandCache`
- `getParseDir() + "/currentUser"`: used by `ParseCorePlugins`
- `getParseDir() + "/currentInstallation"`: used by `ParseCorePlugins`
- `getParseDir() + "/currentConfig"`: used by `ParseCorePlugins`
- `getParseCacheDir() + "/applicationId"`: used by `Parse`
- `getParseCacheDir() + "/GCMRegistrar"`: used by `GcmRegistrar`
- `getParseCacheDir() + "/files"` : used by ParseFileController for `ParseFile` temporary and cache files
- `getParseFilesDir() + "/push"` : used by `PushRouter`

So we should move all `getParseDir()` (deprecated) to either `getParseCacheDir()` or `getParseFilesDir()`. But looking at this, I see no pattern. What should go to `cacheDir`, and what to `filesDir`?

I know the implementation won't be easy to avoid breaking current caches, so we will have to support legacy dirs for a while etc. But here I would just like to understand the intended usage of `cacheDir` and `filesDir`. We actually store no files, these are all just caches of something. Even `ParseFile` are stored in `cacheDir/files`...

**In short** 
We need a new home for LocalId, CommandCache, currentUser, currentInstallation and currentConfig. Should it be `cacheDir` or `filesDir` ? And actually, couldn't we use `cacheDir` for everything?

natario1:
Suggestions:

- LocalId: move to `cacheDir`
- CommandCache: move to `cacheDir`
- currentUser: move to `filesDir`
- currentInstallation: move to `filesDir`
- currentConfig: move to `filesDir`
- applicationId: I don’t think this belongs to cacheDir, move to `filesDir`

and leave the others where they are. If I got this correctly this means that when the user clears app cache, `currentUser`, `currentInstallation`, `currentConfig`, `applicationId` and push history are kept. The rest is cleared.
What do you think?

[...]

PR #750, opened 2017-10-10
Remove deprecated getParseDir

Jawnnypoo:
Remove access and replace usages with the non-deprecated method

### DIFF
--- a/Parse/src/main/java/com/parse/Parse.java
+++ b/Parse/src/main/java/com/parse/Parse.java
@@ -486,15 +486,6 @@ public class Parse {
     return true;
   }
 
-  /**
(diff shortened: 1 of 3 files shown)


natario1:
I think you should provide some legacy mechanism?

Like, if the old version was saving stuff in folderA, and the new version checks in folderB, the device is in a unexpected state. I haven't seen what kind of files were saved there though

Jawnnypoo:
This method was not public, and was only used by the SDK itself, and the `getParseCacheDir` points to the same folder, it just assures the directory exists beforehand. So, nothing really changes, this is just to remove some deprecation

natario1:
AFAIK they should be different, I remember wanting to do the same but I was kind of stuck #679 for this legacy thing..

Jawnnypoo:
I see, I'll just close this for now then, thanks

PR #769, opened 2017-12-19
Remove build tools and shorten travis script

Jawnnypoo:
This way we don't have to keep updating the travis configuration over and over, the dependencies are determined and downloaded at build time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,24 +1,12 @@
-branches:
-  only:
-    - master
-    - /^\d+\.\d+\.\d+$/ # regex
-
 language: android
-sudo: false
 
 jdk:
   - oraclejdk8
 
-android:
-  components:
-    - tools
-    - platform-tools
-    - build-tools-27.0.0
-    - android-27
-    - doc-27
-
 before_install:
-    - pip install --user codecov
+  - pip install --user codecov
+  - mkdir "$ANDROID_HOME/licenses" || true
+  - echo "d56f5187479451eabf01fb78af6dfcb131a6481e" > "$ANDROID_HOME/licenses/android-sdk-license"
 
 script:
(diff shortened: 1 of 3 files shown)


natario1:
Don't know how the echo part works, but it's cool. Deployments (tag triggers) work well without the branches section?

.travis.yml
```diff
@@ -1,24 +1,12 @@
-branches:
-  only:
-    - master
```
rogerhu: I used this to tag releases and launch deploys to Bintray...how come we're removing this?

Jawnnypoo: Ah, so it deploys a snapshot if its master and a release version if its a tag? I removed it because I thought it made since to have Travis builds run on pull requests, which it seems like presently it is preventing that from happening

rogerhu: ah interesting.  i was trying to use the deploy section below to only upload on release tagged branches (https://docs.travis-ci.com/user/customizing-the-build/#Deploying-your-Code).  is there a way to build on pull requests and still have this option?

Jawnnypoo: So we could do something in a script like:
```
# Check if tag is not empty
# http://stackoverflow.com/a/21407325/895797
if [ -n "$TRAVIS_TAG" ]; then
     echo "Deploying project";
     eval "./gradlew deploy"
else
     echo "Building project";
     eval "./gradlew build"
fi
```
and we would just run this `.sh` script in Travis

rogerhu: cool I like it =).  thanks =)

rogerhu: did you add yet?

rogerhu:
Need a rebase -- I had to add https://github.com/parse-community/Parse-SDK-Android/commit/182cc12a4311813e2ac8ef8604c781c5da0f470f to fix API 27 on Travis.

PR #774, opened 2018-01-19
Build for any SDK on Travis.

Jawnnypoo:
Trying this again. This time, leaving the build only restriction

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,22 +4,14 @@ branches:
     - /^\d+\.\d+\.\d+$/ # regex
 
 language: android
-sudo: false
 
 jdk:
   - oraclejdk8
 
-android:
-  components:
-    - tools
-    - platform-tools
-    - build-tools-27.0.0
-    - android-27
-    - doc-27
-
 before_install:
     - pip install --user codecov
-    - yes | sdkmanager "platforms;android-27"
+    - mkdir "$ANDROID_HOME/licenses" || true
+    - echo "d56f5187479451eabf01fb78af6dfcb131a6481e" > "$ANDROID_HOME/licenses/android-sdk-license"
 
 script:
   - ./gradlew clean testDebugUnitTest jacocoTestReport
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,7 @@ jdk:
 
 before_install:
     - pip install --user codecov
-    - mkdir "$ANDROID_HOME/licenses" || true
-    - echo "d56f5187479451eabf01fb78af6dfcb131a6481e" > "$ANDROID_HOME/licenses/android-sdk-license"
+    - yes | sdkmanager --licenses
 
 script:
   - ./gradlew clean testDebugUnitTest jacocoTestReport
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -4,22 +4,14 @@ branches:
...
-    - platform-tools
-    - build-tools-27.0.0
-    - android-27
-    - doc-27
-
 before_install:
     - pip install --user codecov
-    - yes | sdkmanager "platforms;android-27"
+    - mkdir "$ANDROID_HOME/licenses" || true
+    - echo "d56f5187479451eabf01fb78af6dfcb131a6481e" > "$ANDROID_HOME/licenses/android-sdk-license"
```
flovilmart: what is that licence key?

Jawnnypoo: It's a hash that signifies that you have read and agreed to the Android license. Here it is used in ButterKnife as well https://github.com/JakeWharton/butterknife/blob/master/.travis.yml

rogerhu: i don't think it's necessary?  i already added yes to the sdk manager..  besides Travis needs line 22 to pass

Jawnnypoo: The downside to what you added is that you will have to continue to update it each time the SDK gets a bump, such as when API 28 releases. With this setup, it accepts all licenses for the Android SDK and therefore will install anything from the sdkmanager that it needs.

rogerhu: Can we just do this?

https://stackoverflow.com/a/45782695

```bash
yes | sdkmanager --licenses
```

Jawnnypoo: I've never tried that. Let me see.

Jawnnypoo: Looks like Travis doesn't like that https://travis-ci.org/parse-community/Parse-SDK-Android/builds/335729767

rogerhu: Hmm it's using an older sdkmanager tools.

https://docs.travis-ci.com/user/languages/android/#Installing-a-newer-SDK-Platform-Tools-revision

When I run it locally, I see:

```
./sdkmanager
Usage: 
  sdkmanager [--uninstall] [<common args>] [--package_file <file>] [<packages>...]
  sdkmanager --update [<common args>]
  sdkmanager --list [<common args>]
  sdkmanager --licenses [<common args>]

```

Jawnnypoo:
Does this sound good @rogerhu ?

.travis.yml
rogerhu: I think you need to specify a build tools, which should hopefully give you a newer sdkmanager.

```
No build-tools version is specified in android.components. Consider adding one of:
build-tools-27.0.3

```

Jawnnypoo: Well yeah thats the nice thing about doing it the way I have it now. It does not require changing the build tools version in the travis config each time you change it in code. On top of that, if we let the SDK decide which build tools to use based on the Android plugin, then we would not know the build tools version to install ahead of time. Instead, if we accept the license the way it is now, it would accept all build tools versions

rogerhu: Ok well we should fix and change to use the more obvious sdkmanager line when Travis bumps the default build tools version.

PR #782, opened 2018-01-31
Allow setting of max retries

Jawnnypoo:
Presently, the number of retries of a Parse command is always set to 5, and is not changeable. This allows you to set the number of max retries via the Parse.Configuration. Example:

```
 val config = Parse.Configuration.Builder(this)
                .applicationId("test")
                .clientKey("blah")
                .server("https://example.com/parse")
                .maxRetries(2)
                .build()
```
or even disable retries by setting `maxRetries(0)`

### DIFF
--- a/Parse/build.gradle
+++ b/Parse/build.gradle
@@ -5,7 +5,7 @@ apply plugin: 'com.github.kt3k.coveralls'
 apply plugin: 'com.jfrog.bintray'
 
 group = 'com.parse'
-version = '1.16.5'
+version = '1.16.6-SNAPSHOT'
 
 ext {
     projDescription = 'A library that gives you access to the powerful Parse cloud platform from your Android app.'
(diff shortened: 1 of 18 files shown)


natario1:
+10 for removing ParsePlugins.Android, such a headache. Looks good to me.

addisonElliott:
If I get time, I'll check this out. 

@Jawnnypoo Wanted to let you know that this was on my (long) TODO list and I'm glad you submitted a PR for this! :thumbsup:

rogerhu:
@addisonElliott if you can check it out that would be much appreciated =)

addisonElliott:
Looks good to me. I pulled it locally and tested it out.

On a sidenote, what is the purpose of `userAgent` in the `ParsePlugins` class? It seems it is not used in Parse Android SDK and it is a private package, so it cannot be called. It may be worthwhile to create a function in Parse.java that calls this function. I can submit the PR if you would like.

Issue #783, opened 2018-02-01
Upload javadocs to maven


addisonElliott:
I submitted this issue regarding outdated API docs for Parse Android:
* parse-community/docs#486

As I mentioned in the issue, there is a service http://javadoc.io that provides documentation references to any docs on maven. I noticed that Parse Android is missing for any recent version and found out this is because javadocs are not uploaded to maven.

However, there is a javadocs uploaded for 1.15.5: http://www.javadoc.io/doc/com.parse/parse-android/1.15.5

# Proposed Idea
Submit change such that each time a new version of Parse Android is uploaded to maven, the javadocs are present with it.

Then changes can be made to the docs so that it links to the javadocs reference for Parse Android

I would love some feedback on this. What do you think?

# Solution
A bit of digging shows that Parse Android uses `bintrayUpload` to publish to maven.

According to bintray docs, we should be able to add the artifact `javadocJar` here https://github.com/parse-community/Parse-SDK-Android/blob/master/Parse/build.gradle#L252.

I'm hesitant to submit a PR for now because this cannot be tested on a separate branch (that I know of). Also, I'm not sure if adding the artifact will generate the docs automatically or if additional commands are needed.

My plan is to hopefully create a dummy repository with this plugin and mess around with it myself.

flovilmart:
THAT’s a sweet idea! Let us know if you need anything?

rogerhu:
You can but you can just run the command line and look inside the build/ dir to find the generated Jar file.

rogerhu:
You can also run bintrayUpload --verbose and watch to see if it does it correctly (the API calls will fail but you can still see if the upload was attempted)

PR #785, opened 2018-02-06
Add javadoc artifact to maven publication

addisonElliott:
When publishing to maven, the javadocs will be uploaded automatically along with the sources.

Fixes issue:
* https://github.com/parse-community/docs/issues/486
* #783 

**Note:** I was not able to test this yet since I am waiting on my Sonatype account to be accepted. I have a test repository that I will try this with. However, I have been reading articles and I believe this will work.

### DIFF
--- a/Parse/build.gradle
+++ b/Parse/build.gradle
@@ -249,7 +249,7 @@ publishing {
             groupId group
             // We have to specify it here because otherwise Bintray's plugin will assume the artifact's name is Parse
             artifactId artifact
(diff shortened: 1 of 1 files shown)


rogerhu:
Just run ./gradlew publishToMavenLocal and look in the ~/.m2/repository/com/parse/parse-android dir to see if it's there.  Also Sonatype doesn't matter since we upload to Bintray and then sync to Sonatype.

rogerhu:
I see it there now, thanks!

Issue #767, opened 2017-12-14
authData is not cleaned up when linkWithAsync() fails


mhartvick:
Whenever [linkWithAsync()](https://github.com/parse-community/Parse-SDK-Android/blob/master/Parse/src/main/java/com/parse/ParseUser.java#L1280) fails, [the anonymity is restored](https://github.com/parse-community/Parse-SDK-Android/blob/master/Parse/src/main/java/com/parse/ParseUser.java#L457), but the `authData` map still contains the value for the `authType` that just failed.

It's unexpected from a consumer point of view, `isLinked(myAuthType)` will be `true` right after failing to link the exact same `myAuthType`, and differs from the iOS behavior (according to a colleague of mine who tried to reproduce the issue on its platform)

flovilmart:
@rogerhu can you have a look please?

Jawnnypoo:
Seems like we can close this out now. Thanks for the contribution @mhartvick !

PR #787, opened 2018-02-09
Clean up authData when linkWithAsync() fails

mhartvick:
This was causing `isLinked(myAuthType)` to be true after a `linkWithAsync()` would fail for `myAuthType`, as explained in #767 

This PR would make the behavior on both iOS and Android the same.

### DIFF
--- a/Parse/src/test/java/com/parse/ParseUserTest.java
+++ b/Parse/src/test/java/com/parse/ParseUserTest.java
@@ -699,13 +699,13 @@ public class ParseUserTest extends ResetPluginsParseTest {
         partialMockUser.linkWithInBackground(authType, authData);
     linkTask.waitForCompletion();
 
-    // Make sure new authData is added
-    assertSame(authData, partialMockUser.getAuthData().get(authType));
     // Make sure we save the user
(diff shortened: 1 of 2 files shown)


Parse/src/test/java/com/parse/ParseUserTest.java
```diff
@@ -699,13 +699,13 @@ public void testlinkWithInBackgroundWithSaveAsyncFailure() throws Exception {
         partialMockUser.linkWithInBackground(authType, authData);
     linkTask.waitForCompletion();
 
-    // Make sure new authData is added
-    assertSame(authData, partialMockUser.getAuthData().get(authType));
```
mhartvick: This is the exact same test/comment from `testlinkWithInBackgroundWithSaveAsyncSuccess()` above, which makes me think it was expected in the test only because it was copied

flovilmart:
@rogerhu this seems reasonable as the iOS SDK is performing the same
https://github.com/parse-community/Parse-SDK-iOS-OSX/blob/master/Parse/Parse/PFUser.m#L896

rogerhu:
Great thanks.

PR #793, opened 2018-03-17
Added masterKey to parse client

bitterbit:
Hello All! :fireworks: 

Lately I found the need to set the masterKey for an android app I am writing.  Feel free to contact me for anything.

**Thanks!**
Gal.

### DIFF
--- a/Parse/src/main/java/com/parse/Parse.java
+++ b/Parse/src/main/java/com/parse/Parse.java
@@ -49,6 +49,7 @@ public class Parse {
       private Context context;
       private String applicationId;
       private String clientKey;
+      private String masterKey;
       private String server;
       private boolean localDataStoreEnabled;
       private OkHttpClient.Builder clientBuilder;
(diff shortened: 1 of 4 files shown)


flovilmart:
You should never need to set a masterKey in a client application as this master key has a full and unrestricted access to your database.

bitterbit:
What should I do then if I want to build a parse dashboard admin app?

In my opinion the client library should match the features that the server has. Yet it doesn’t mean masterKey should be required but only optional

flovilmart:
For your application to work, you,ll need to embed the masterKEy in the client, which is unsafe and we don't support nor encourage unsafe usage. If you wanna pursue this, you can very well use your fork at your own risks.

Jawnnypoo:
You also have access to the OkHttp Client builder when building the Parse instance, you can add an interceptor and add the master key header there. But yeah, at your own risk 😉

bitterbit:
If the parse key was passed only when and where needed and not in the initialisation, would that be ok?

flovilmart:
We won’t allow to change that, as it’s pretty clear this can tremendously be unsafe. As mentioned, you can setup a request interceptor and inject the masterKey to all requests if you need to

bitterbit:
The thing is that I was planning on making another pull request implementing querying a schema, but any schema related query must have a masterKey.
Will you accept such a pull request if there is no masterKey natively?

Jawnnypoo:
It's probably best to just fork the SDK and keep it as a separate thing, I don't think it would be good to expose for 99.9% of people using this SDK

flovilmart:
Yes @bitterbit, as @Jawnnypoo mentions, this is not something we want part of the client SDK

PR #800, opened 2018-04-12
FCM Support - GCM Module

Jawnnypoo:
This ended up being a pretty large effort and a substantial set of changes. 

The good news is that now that GCM is not anywhere in the core SDK, it removes quite a bit of code and allows us to lean more on the Google Play Services library to do the implementation. 

So, this update will **not** be backwards compatible, and will require us to solve the problem of shipping multiple artifacts with the library. I think maybe you have more experience with this process, @rogerhu ? I once again would advocate for switching over to jitpack.io to make the releases easy and doable by anyone who can tag a release. But I digress. 

I have tested the `gcm` and `fcm` modules in a sample project with a standard setup. Everything seems to work fine. Note that the instructions for setting up can be found in the README of the `gcm` and `fcm` modules respectively. We will need to update http://docs.parseplatform.org/tutorials/android-push-notifications/ to either reflect the changes, or point users to the FCM and GCM docs, leaning them more toward FCM since GCM is going away. 

Somewhat unrelated, but I think with these changes bumping us to a 1.17.0, we should consider removing or deprecating any other things we think need to be removed or deprecated. I for one think that we need to get rid of parsing the server url and application id and client key from the Android Manifest. It seems like as we add more and more methods to the Parse.Configuration.Builder, we would just want people to use this instead.


Parse/src/main/java/com/parse/PLog.java
```diff
@@ -10,7 +10,10 @@
 
 import android.util.Log;
 
-/** package */ class PLog {
+/**
+ * Parse Logger
+ */
+public class PLog {
```
Jawnnypoo: Had to make this public to allow for logging in the other modules

fcm/README.md
```diff
@@ -0,0 +1,69 @@
...
+    implementation 'com.parse:parse-android-fcm:latest.version.here'
+
+    implementation 'com.google.firebase:firebase-messaging:12.0.1'
+}
+```
+Then, follow Google's docs for [setting up an Firebase app](https://firebase.google.com/docs/cloud-messaging/android/client). Although the steps are different for setting up FCM with Parse, it is also a good idea to read over the [Firebase FCM Setup](https://firebase.google.com/docs/cloud-messaging/android/client).
+
+You will then need to register some things in your manifest, specifically:
+```xml
+<service
```
rogerhu: The parse-android-fcm should include this service in its AndroidManifest.xml file.  There should be manifest merging when building it as an AAR.

fcm/README.md
```diff
@@ -0,0 +1,69 @@
...
+    </intent-filter>
+</service>
+```
+where `MyFirebaseInstanceIdService` is your own custom class which extends `ParseFirebaseInstanceIdService`.
+
+Additional, you will register:
+
+```xml
+<service
+    android:name=".MyFirebaseMessagingService">
```
rogerhu: same here too?

gcm/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+Next:
+```xml
+<receiver
+    android:name="com.google.android.gms.gcm.GcmReceiver"
+    android:exported="true"
+    android:permission="com.google.android.c2dm.permission.SEND" >
+    <intent-filter>
+        <action android:name="com.google.android.c2dm.intent.RECEIVE" />
+        <category android:name="com.your.package.name.here" />
```
rogerhu: You should be able to use so you don't need to worry about mistyping.  Gradle swaps in the correct package name.

```
<!-- ${packageName} is substituted for the package name declared above.  -->
  <permission android:protectionLevel="signature" android:name="${packageName}.permission.C2D_MESSAGE" />
  <uses-permission android:name="${packageName}.permission.C2D_MESSAGE" />```

File: gcm/src/main/AndroidManifest.xml [hunk 5]
```diff
@@ -0,0 +1,18 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    package="com.parse.gcm">
+
+    <uses-permission android:name="com.google.android.c2dm.permission.RECEIVE" />
+    <uses-permission android:name="android.permission.WAKE_LOCK" />
+
+    <application>
+
+        <service
```
rogerhu: move ParseGCMListenerService and ParseGCMInstanceIDListenerService here too?

rogerhu:
@Jawnnypoo you can setup to build artifacts for different projects, this happens already with https://github.com/parse-community/ParseUI-Android

rogerhu:
Thanks for working on this!

fcm/src/main/java/com/parse/fcm/ParseFCM.java
```diff
@@ -0,0 +1,44 @@
...
+                .setRecurring(false)
+                .setReplaceCurrent(true)
+                // retry with exponential backoff
+                .setRetryStrategy(RetryStrategy.DEFAULT_EXPONENTIAL)
+                .setConstraints(
+                        // only run on a network
+                        Constraint.ON_ANY_NETWORK
+                )
+                .setService(ParseFirebaseJobService.class) // the JobService that will be called
+                .setTag("upload-token")        // uniquely identifies the job
```
rogerhu: can we set as a constant?

gcm/README.md
```diff
@@ -0,0 +1,98 @@
...
+### Installation
+
+Add dependency to the application level `build.gradle` file.
+
+```groovy
+dependencies {
+    implementation 'com.parse:parse-android:latest.version.here'
+    implementation 'com.parse:parse-android-gcm:latest.version.here'
+
+    implementation "com.google.android.gms:play-services-gcm:latest.version.here"
```
rogerhu: this shouldn't be necessary given you declared play-services-gcm as an api dependency

Jawnnypoo:
Okay @rogerhu Feedback has been addressed. Let me know if my comment above makes sense about not registering the services and receivers in the library projects. I think the reasoning would be the same as why it is not done automatically in the normal FCM implementation either, outlined here: https://firebase.google.com/docs/cloud-messaging/android/client

Basically, only one instance ID and messaging service is allowed per app, and forcing the Parse ones to already be registered would not allow users to define their own.

Jawnnypoo:
Something else to note is that I cannot update the build to use Gradle 4.4, with gradle build tools `com.android.tools.build:gradle:3.1.1`, there are errors being thrown when I build with those versions that I think have to do with either the artifact generation or the javadoc generation. I am not too familiar with these parts of the build process, so maybe you can look into this too @rogerhu ?

rogerhu:
@Jawnnypoo you'll just need to add

https://github.com/rogerhu/Parse-SDK-Android/blob/070c3afd1aec31efaf466da6877fd6cb6f0c84dc/Parse/build.gradle#L214-L262 to both the fcm and gcm directories.

https://github.com/parse-community/Parse-SDK-Android/pull/804/files simplifies the Gradle statements

rogerhu:
Rebase too :)

I suggest we use a common version share between the fcm and gcm version as well

Jawnnypoo:
> I suggest we use a common version share between the fcm and gcm version as well
Do you mean for the FCM/GCM dependencies?

rogerhu:
Yeah see commonLibVersion in https://github.com/parse-community/ParseUI-Android/blob/master/build.gradle

Jawnnypoo:
Okay, added the common lib version to all modules. I am not sure which parts of the Parse/build.gradle I need to apply to the `gcm` and `fcm` modules? Does it require all of the bintray plugin configuration, all of the artifactory info, etc? It seems like this would be a lot of duplication of the same thing if so. Is there a way we could pull this out into its own script, similar to what Square does here? https://github.com/square/sqlbrite/blob/master/gradle/gradle-mvn-push.gradle

Parse/build.gradle
```diff
@@ -5,7 +5,7 @@ apply plugin: 'com.github.kt3k.coveralls'
 apply plugin: 'com.jfrog.bintray'
 
 group = 'com.parse'
-version = '1.16.8-SNAPSHOT'
+version = rootProject.ext.commonLibVersion
```
rogerhu: keep this version?  i think we don't need to have the same version tied to the gcm/fcm plugins...

flovilmart: Do you plan to distribute the plugins separately? having everything at the same version helps understanding issues  :)

natario1:
Nice work @Jawnnypoo !

You should now remove PushServiceApi26 from Parse/manifest.

Speaking about cleanup, I remember being tempted to rip off PushRouter and PushHistory. PushHistory seems to be doing legacy operations, and PushRouter seem to exist only to deal with the history. Is it worth to take a look?

flovilmart:
@natario1 they are used for de-duplication (https://github.com/parse-community/Parse-SDK-Android/blob/master/Parse/src/main/java/com/parse/PushRouter.java#L28)

rogerhu:
I don't think so...I've seen cases of dup pushes.  I'd leave it in.

Jawnnypoo:
@rogerhu Can you take a look at the maven setup I did within the gradle folder? I am not sure if it is correct. (Latest commit)

Jawnnypoo:
Any ideas why the build may have failed here @rogerhu ?  Looks like it is related to the codecov setup?

rogerhu:
@Jawnnypoo no it's the Gradle 3.1.1 upgrade that caused stuff to break.  If I downgrade your current stuff seems to build.

rogerhu:
If you do gradlew build --stacktrace:

```
Caused by: groovy.lang.GroovyRuntimeException: Conflicting module versions. Module [groovy-all is loaded in version 2.4.7 and you are trying to load version 2.4.12
```

https://issuetracker.google.com/issues/77236219

It seems like `com.android.tools.lint:lint-gradle:26.1.1` that adds groovy-all..

Jawnnypoo:
Nice @rogerhu Looks like that worked. Anything else we need to get this merged?

PR #801, opened 2018-04-12
Add automated docs in Travis CI

addisonElliott:
Updates gradle to create a docs/api directory on Travis deployment. Additionally, clean up Travis CI to have two stages, test and release. This stays consistent with JS and iOS branches that received an update already.

Use Travis pages provider to upload the docs directory.

**Note:** In Parse-SDK-JS and Parse-SDK-iOS, the publish docs scripts clone the Git repository before placing changes and uploading again. I'm not sure there is not any point to it. The pages provider does a force-push that does not store incremental changes for the docs.

Also, there is a new feature `keep-history` that will keep the previous history if incremental docs are desired. If desired, let me know and I will add it.

More info can be found [here](https://docs.travis-ci.com/user/deployment/pages/) and [here](https://github.com/travis-ci/travis-ci/issues/7562)

This code was tested on fork where I setup Travis CI. Docs can be seen [here](https://addisonelliott.github.io/Parse-SDK-Android/) and build can be seen [here](https://travis-ci.org/addisonElliott/Parse-SDK-Android)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,9 +13,6 @@ before_install:
     - mkdir "$ANDROID_HOME/licenses" || true
     - echo "d56f5187479451eabf01fb78af6dfcb131a6481e" > "$ANDROID_HOME/licenses/android-sdk-license"
 
-script:
-  - ./gradlew clean testDebugUnitTest jacocoTestReport
-
 after_success:
   - ./gradlew coveralls
   - codecov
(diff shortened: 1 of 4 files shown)


Jawnnypoo:
Nice, this looks good.

.travis.yml
```diff
@@ -26,10 +23,22 @@ cache:
...
+      script: ./gradlew clean testDebugUnitTest jacocoTestReport
+    - stage: release
+      script: ./gradlew javadocRelease
+      deploy:
+        - provider: pages
+          skip_cleanup: true
+          github_token: $GITHUB_TOKEN # Set in travis-ci.org dashboard
+          local_dir: docs/
+          on:
+            all_branches: true
```
Jawnnypoo: Wouldn't we want to only do this on the master branch, @addisonElliott ?

addisonElliott: I spent an hour or two thinking the same thing. I used the same code that @flovilmart used when he added automated docs for JS and iOS. 

See here:
* https://github.com/parse-community/Parse-SDK-iOS-OSX/blob/master/.travis.yml#L38
* https://github.com/parse-community/Parse-SDK-iOS-OSX/blob/master/.travis.yml#L38

all_branches excludes PRs. It took me awhile to find documentation saying that. My guess is the thought process is that the only new branches on Parse-SDK-Android will be the master branch commit or new tags.

flovilmart:
Thanks @AddisonElliot!

PR #804, opened 2018-04-13
Simplify our dependency creation by using plugin

rogerhu:


build.gradle
```diff
@@ -12,6 +12,7 @@ plugins {
     id "com.jfrog.artifactory" version "4.4.18"
     id "com.jfrog.bintray" version "1.7.3"
     id 'com.github.ben-manes.versions' version '0.17.0'
+    id 'digital.wup.android-maven-publish' version '3.0.0' // tweak depending on Gradle version
```
rogerhu: this plugin version depends what Gradle version is used.  The api for gradle keeps changing..

Jawnnypoo:
I have used this one in the past, not sure what the differences are.

https://github.com/dcendents/android-maven-gradle-plugin

But this looks good 👍

rogerhu:
There are two Maven plugins, maven and maven-publish.  The one you've used (and I have done so myself) uses the maven plugin.

  Bintray's official upload script relies on the maven-publish plugin.  But it fails to produce dependencies in the metadata for Android.  This plugin takes care of that part.

Jawnnypoo:
Ahh okay, makes sense. Any ideas about the build failing? Seems like an odd error.

rogerhu:
I bumped

PR #809, opened 2018-04-30
Remove support for manifest configuration

Jawnnypoo:
The ability to configure the server in the manifest is a sort of relic of when the SDK was being used with Parse.com. In order to move forward with allowing for customization, as well as adding new plugins such as FCM support, I think it is best that in the 1.17.0 release we ditch the Manifest configuration. This is the best time to do it, since 1.17.0 has API breaking changes which, if users are using GCM, will require them to update their Manifest and code anyhow. 

We certainly would want to make sure it is clear in the release notes that this is occurring.

### DIFF
--- a/gcm/src/main/java/com/parse/gcm/ParseGCM.java
+++ b/gcm/src/main/java/com/parse/gcm/ParseGCM.java
@@ -9,15 +9,11 @@
 package com.parse.gcm;
 
 import android.content.Context;
-import android.os.Bundle;
-import android.support.annotation.Nullable;
+import android.support.annotation.NonNull;
 
-import com.firebase.jobdispatcher.Constraint;
 import com.firebase.jobdispatcher.FirebaseJobDispatcher;
 import com.firebase.jobdispatcher.GooglePlayDriver;
 import com.firebase.jobdispatcher.Job;
-import com.firebase.jobdispatcher.RetryStrategy;
-import com.parse.ManifestInfo;
 import com.parse.PLog;
 
 /**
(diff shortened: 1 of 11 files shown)


gcm/src/main/java/com/parse/gcm/ParseGCM.java
```diff
@@ -25,39 +24,25 @@
...
-                .setRetryStrategy(RetryStrategy.DEFAULT_EXPONENTIAL)
-                .setConstraints(
-                        // only run on a network
-                        Constraint.ON_ANY_NETWORK
-                )
-                .setService(ParseGCMJobService.class) // the JobService that will be called
-                .setTag(JOB_TAG_REGISTER)        // uniquely identifies the job
-                .build();
-
+        Job job = ParseGCMJobService.createJob(dispatcher, gcmSenderFromManifest(context));
```
rogerhu: You're getting read of manifest but keeping it here?

Jawnnypoo: Yeah, I tried to get rid of it for the GCM module, but it turns out we need the GCM sender ID at unpredictable times, such as when the token gets refreshed. So, we would either have to keep it in the manifest, or store it in SharedPreferences or something. I think it being in the manifest as it was before is the easiest configuration. Since most should be migrating away from the GCM module anyways, I think it is okay.

rogerhu:
Ok we need to update https://docs.parseplatform.org/android/guide/ too...

PR #811, opened 2018-05-02
Easier Push Overrides

Jawnnypoo:
Right now, if you wanted to have a custom notification, you would have to subclass `ParsePushBroadcastReceiver` and override methods. Right now, to create a custom notification, you would override `getNotification`, which returns a notification. Instead, it would be better if the user could override `getNotificationBuilder` and the user is able to just override what they want to within the builder, instead of being forced into rebuilding the notification completely.


gcm/README.md
```diff
@@ -12,7 +12,6 @@ Add dependency to the application level `build.gradle` file.
 
 ```groovy
 dependencies {
-    implementation 'com.parse:parse-android:latest.version.here'
     implementation 'com.parse:parse-android-gcm:latest.version.here'
```
rogerhu: https://api.bintray.com/packages/parse/maven/com.parse:parse-android/images/download.svg should be added as well

PR #814, opened 2018-05-02
Make LogLevel DEBUG instead of VERBOSE for FCM/GCM registration messages

rogerhu:


rogerhu:
(This is mostly because I set things to LogLevel.DEBUG, which is where Parse warning messages used to get flagged.)

PR #815, opened 2018-05-02
Move over to JitPack

Jawnnypoo:


Jawnnypoo:
@rogerhu Were you thinking something like this for the `latest.version.here` thing?

Jawnnypoo:
Going to repurpose this to moving over to JitPack

flovilmart:
Why to move?

Jawnnypoo:
I misunderstood, I was thinking we were going to move over to it for all dependencies, but we are likely going to just move over to it for snapshots.

PR #819, opened 2018-05-04
Fix for allowing for override of notification builder

Jawnnypoo:
This fixes the fact that the SDK was using a forked version of NotificationCompat and instead uses the support library version. We can then revert back to allowing for easy overrides of the ParsePushBroadcastReceiver, allowing users to override exactly what they want to. 

I would love to get this in quickly as a 1.17.1 if possible, since it does break some compatibility if people were overriding previously. Does that sound good to you @rogerhu ?


gcm/build.gradle
```diff
@@ -41,7 +41,8 @@ android {
 }
 
 dependencies {
-    api "com.google.android.gms:play-services-gcm:$googleLibVersionGCM"
+    // last version for GCM to be supported
+    api "com.google.android.gms:play-services-gcm:12.0.1"
```
rogerhu: it sounds like 15.0.0 brought in some 16.0 dependencies... hopefully this problem will be fixed soon

Jawnnypoo: That is strange. I sure hope so

akg1234:
Notification notification = getNotification(context, intent).build();

this get null point exception when the push is data only
better check getNotification(context, intent) !=null before apply the build() function

PR #820, opened 2018-05-05
Set installation to fcm now.

rogerhu:
The push adapter actually ignores the push type and falls back to the device type.  It should be OK to classify these pushes as FCM.  

i.e. this line falls back to using the android deviceMap:

https://github.com/parse-community/parse-server-push-adapter/blob/master/src/PushAdapterUtils.js#L20

The only accepted types are actually `android` and `fcm`:

https://github.com/parse-community/parse-server-push-adapter/blob/master/src/ParsePushAdapter.js#L35

We should move off the parse-server GCM module because the FCM sending needs a service account and database URL instead of a GCM sender ID.


Jawnnypoo:
Oh, nice! That is good to know!

PR #822, opened 2018-05-06
Revert "Set installation to fcm now. (#820)"

rogerhu:
This reverts commit f3d47f6eb1a03240ad48fd10b35d75b61d8c48ff.

Making this change requires parse-server to set push config to be `fcm` in addition to `android`:

i.e.

```javascript
"push": { "fcm": { "senderId": "XXXX", "apiKey": "XXXX" }, "android": { "senderId": "XXXX", "apiKey": "XXXX" }}
```

If your push configuration is set to only 'android' type, then there will be no corresponding push adapter and pushes of this type will fail.

I'm conflicted here... on one hand, it's better to keep track of FCM pushes separately.  On the other hand, is it reasonable to ask people to upgrade their Parse servers to include this change too?

@flovilmart @Jawnnypoo

### DIFF
--- a/fcm/src/main/java/com/parse/fcm/ParseFirebaseJobService.java
+++ b/fcm/src/main/java/com/parse/fcm/ParseFirebaseJobService.java
@@ -26,7 +26,6 @@ import com.parse.SaveCallback;
 public class ParseFirebaseJobService extends JobService {
 
     private static final String JOB_TAG_UPLOAD_TOKEN = "upload-token";
-    private static final String PUSH_TYPE = "fcm";
 
     static Job createJob(FirebaseJobDispatcher dispatcher) {
         return dispatcher.newJobBuilder()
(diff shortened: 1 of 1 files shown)


rogerhu:
1.17.1 uses GCM as the push type, which is good enough to verify things still work.  

I managed to get a working FCM adapter with the new HTTP reference (https://firebase.google.com/docs/reference/fcm/rest/v1) so strongly suggest we keep this.

flovilmart:
If we can do everything in the background without requiring any update to parse server that’s what we should do. Old clients may connect to new servers, therefore breaking changes on the server have to be avoided

rogerhu:
I'm reverting the constant.

Jawnnypoo:
Looks like this has some merge conflicts @rogerhu

rogerhu:
I already reverted with another PR.

PR #823, opened 2018-05-06
Change README

rogerhu:
@Jawnnypoo 

Is there a reason for overriding these classes?


Jawnnypoo:
Ah, no. I changed them to this in the docs repo. Meant to change them here too 😎

PR #826, opened 2018-05-19
Annotate nullable methods on ParseObject

kurtisnelson:
This makes the object linter/Kotlin friendly.


Jawnnypoo:
Nice, I was thinking this would be a good addition for Kotlin support. I'll probably expand on this for the put methods, as well as for other classes like ParseQuery

Jawnnypoo:
Should need any docs, but it's with noting that if someone is upgrading to the next version and using Kotlin, they might see build exceptions, since it will change things from String to String? etc

kurtisnelson:
That's good build breakage.

Jawnnypoo:
Of course! Just something to note and potentially a reason to make the next version a 1.18.0

PR #835, opened 2018-05-29
publish fcm / gcm on their subpath

### LINKED ISSUE #834: Android API documentation links to wrong package
The link on parseplatform.org to the android sdk documentation does not point to the correct package. Instead it points to the com.parse.gcm package.  All links I have been able to find to the api reference lead to this same place.

http://parseplatform.org/Parse-SDK-Android/api/

Apologies if this is not the correct place to report this.

flovilmart:
Fixes #834


Jawnnypoo:
Nice, thanks @flovilmart , lets see what happens

PR #843, opened 2018-06-07
Improved exception handling in EventuallyPin.findAllPinned()

DanGTZ:
The code in the callback starting at [L172](https://github.com/DanGTZ/Parse-SDK-Android/blob/26c80e68e79b373dd151287f1bdc3add6e9fa19b/Parse/src/main/java/com/parse/EventuallyPin.java#L172) doesn't check if the task is faulted, and assumes it's successful by iterating over the result's list.
So there is a potential null pointer exception at [L175](https://github.com/DanGTZ/Parse-SDK-Android/blob/26c80e68e79b373dd151287f1bdc3add6e9fa19b/Parse/src/main/java/com/parse/EventuallyPin.java#L175).
Also in other parts of the sdk onSuccessTask is used for the same case : https://github.com/parse-community/Parse-SDK-Android/blob/e86b920a001e9b184aa7d7f81cad00abcbe2afe8/Parse/src/main/java/com/parse/ParsePinningEventuallyQueue.java#L583

### DIFF
--- a/Parse/src/main/java/com/parse/EventuallyPin.java
+++ b/Parse/src/main/java/com/parse/EventuallyPin.java
@@ -166,7 +166,7 @@ class EventuallyPin extends ParseObject {
 
     // We need pass in a null user because we don't want the query to fetch the current user
     // from LDS.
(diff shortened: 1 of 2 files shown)


Jawnnypoo:
Nice stuff.

PR #854, opened 2018-07-19
Make things public for ParseLiveQuery

Jawnnypoo:
Since we are going to repackage ParseLiveQuery to be `com.parse.livequery`, we will need to make a few things publicly accessible. Luckily there are not too many things required.


Jawnnypoo:
Redoing since it would be difficult to merge

PR #855, opened 2018-07-20
Formatting

Jawnnypoo:
- Sets things as public as needed for repackaging Parse LiveQuery
- Renames the module from `Parse` to `parse` to make more standard
- Applies auto-formatting with default Android Studio rules to all files
- Updates the project for deploying with JitPack
- Removes references to library version, since it would no longer be applicable (version not hard coded within library, but rather defined via Git tags)

Obviously this PR is huge since it applies both a rename and auto formatting. I wanted to get this in when there were no other pull requests pending, since it would be extremely difficult to ever merge this into a PR. 

I think this library will be much, much easier to deploy changes to if we go with this solution and use JitPack for distribution. This way, all you need to deploy a new artifact is to be able to create tags in Git. No need for Maven Central, Bintray or any other configuration. This way we can iterate quickly and create releases with little manual effort needed. 

Once this is merged in, Parse-LiveQuery, ParseUI, ParseTwitterUtils and ParseFacebookUtils will need to be updated and version bumped to use the JitPack dependency instead of the bintray one. Once they are all in sync, it will be easy/simple to bring those libraries into this repo for them to exist locally, instead of having to have a repo for each library.


vovkab:
Any reason for killing whole git history, in particular, history for ParseObject?

How this even was approved?

PR #870, opened 2018-07-31
Null headers

### LINKED ISSUE #848: Using parse android sdk throw an exception
I tried initializing Parse Android SDK in my app but anytime I try to save an object parse throws a NullPointerException:
```
06-27 15:22:54.643 25869-25869/glassstones.net.creche E/APP: java.lang.NullPointerException: value for name X-Parse-App-Display-Version == null
```
I have used parse on several other projects and have never seen this exception thrown. This is how I'm initializing Parse:

```
ParseObject.registerSubclass(ParseMessage::class.java)
val config = Parse.Configuration.Builder(context)
                .clientKey("")
                .applicationId(appId)
                .server(baseUrl)
                .build()
        Parse.setLogLevel(Parse.LOG_LEVEL_DEBUG)
        Parse.initialize(config)
```

And this is how I'm saving the object:
```
val message = ParseMessage()
        message.content = "Blah blah"
        message.saveInBackground {
            if (it == null) { 
                Log.e("APP", "Good")
            } else { 
                Log.e("APP", it.message)
            } 
        } 
```
Dependencies:

```
parseVersion = "1.17.3"
androidBuildToolsVersion = "27.0.3"
androidMinSdkVersion = 21
androidTargetSdkVersion = 27
androidCompileSdkVersion = 27
```

Jawnnypoo:
It is unclear what can cause this, but if a header happens to have a null value, OkHttp throws an exception. We can instead just ignore these headers. This will in turn fix #848

### DIFF
--- a/parse/src/main/java/com/parse/ManifestInfo.java
+++ b/parse/src/main/java/com/parse/ManifestInfo.java
@@ -61,6 +61,7 @@ public class ManifestInfo {
                     versionName = getPackageManager().getPackageInfo(getContext().getPackageName(), 0).versionName;
                 } catch (NameNotFoundException e) {
                     PLog.e(TAG, "Couldn't find info about own package", e);
+                    versionName = "unknown";
                 }
             }
         }


flovilmart:
Which value is null?

Jawnnypoo:
For the case of #870 , it is the app version name. This value is controlled by the application, which could choose to make it null. More generally, an Android OS version could be modified to be null as well. It is a rare case and would probably require rooting the device, but there is no reason our SDK should cause crashes if any of these headers are null. Potentially some of the checks are unneeded, since things like the application ID would never be null at this point, but for the sake of consistency, we can use the same method to add all the headers and avoid any `value = null` exceptions being thrown from OkHttp.

flovilmart:
I'd rather have the single parameter return an 'unknown' or 'undefined' in this case. Making the method NonNull. As you mention we're polluting the code for edge cases.

Jawnnypoo:
Sure, we can do that. It just seems like we would never potentially want the SDK to crash an app just because it happened to pass something null as a header value. But I guess we can cross those bridges as SDK users encounter them. PR updated.

flovilmart:
Sure we never want to crash, unless we're in a weird state :)

PR #892, opened 2018-09-27
Feature kotlin delegates

danielsanfr:
In README.md, I described in which my PR will help the community.

Please let me know if I need to do anything else to approve this PR.


Jawnnypoo:
This looks pretty cool! I am wondering how proguard will affect this, since it is using the field name as the key for Parse, and proguard could rename these fields. Might be something we would need to add a proguard rule for. Do you have any experience with this @danielsanfr ?

danielsanfr:
I'm glad you liked it.
Normally I do not let ProGuard rename the properties of my models. Should you add this information to README.md?

Jawnnypoo:
Yep, looks good. Sorry for the delay, and thanks for contributing!

PR #894, opened 2018-10-04
ParseQuery extensions

danielsanfr:
With my last PR #892 now we can use a more secure way of creating queries.

I updated the README to show you how to do this.


Jawnnypoo:
Did you want to send this over again?

PR #926, opened 2019-02-14
Convert project to use AndroidX

Jawnnypoo:
This converts the project to SDK 28 and AndroidX, along with updating the OkHttp dependency to the latest version that supports our minSDK.

Next release should likely be a 1.19.0, since the upgrade path might not be compatible if others are not using AndroidX already.


rogerhu:
Lgtm

Jawnnypoo:
It's hard to tell, but yep, I did that too 👍

Issue #890, opened 2018-09-26
FCM Token not updated when app is updated from GCM to FCM


littleariel:
We are currently migrating our app from GCM to FCM and seeing an issue when users update from the current GCM version of the app to the newer FCM version. 
In that case, it seems that com.google.firebase.INSTANCE_ID_EVENT is never fired and the FCM token therefore not refreshed (seems confirmed by this: https://stackoverflow.com/questions/41330078/is-fcm-token-refreshed-on-app-update). Users going through this update path can't receive push notifications because of this. Users that do a fresh install are fine.

Is there an additional step needed in order to ensure existing installations cleanly register for FCM when updating from a non-FCM version?

flovilmart:
Did you try what is suggested and call ‘getToken()’ on the initial activity. what’s the issue with this?

With the token returned, set in on the current installation and save.

littleariel:
That worked! Thanks!
Not sure if the GCM->FCM migration docs should be updated to mention this?

flovilmart:
Perhaps, you can always open a Pr on the guide to mention it.

Issue #886, opened 2018-09-10
Parse notification not working with Android 9


matdev:
After completing a Parse installation for a new user of our app, the device token in our Parse's Installation table stays undefined and Parse's device notification are not working.

This is quite problematic for users of Android 9.

Is there a solution for this problem ?

I'm using com.parse:parse-android:1.16.7

rodrigosmarques:
Hi,
see this guide. You need to add the Parse-FCM. Have you added?
https://github.com/parse-community/Parse-SDK-Android/tree/master/fcm

g-martins:
Same here. It doesn't work from **Android O** or higher.
The application crashes when it tries to build the notification

`09-18 20:09:14.351 5620-5620/com.reelapps.RAShotListerApp E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.reelapps.RAShotListerApp, PID: 5620
    java.lang.RuntimeException: Unable to start receiver com.parse.ParsePushBroadcastReceiver: java.lang.NullPointerException: Attempt to invoke virtual method 'android.app.Notification android.support.v4.app.NotificationCompat$Builder.build()' on a null object reference
        at android.app.ActivityThread.handleReceiver(ActivityThread.java:3388)
        at android.app.ActivityThread.access$1200(ActivityThread.java:199)
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1661)
        at android.os.Handler.dispatchMessage(Handler.java:106)
        at android.os.Looper.loop(Looper.java:193)
        at android.app.ActivityThread.main(ActivityThread.java:6669)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858)
     Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'android.app.Notification android.support.v4.app.NotificationCompat$Builder.build()' on a null object reference
        at com.parse.ParsePushBroadcastReceiver.onPushReceive(ParsePushBroadcastReceiver.java:174)
        at com.parse.ParsePushBroadcastReceiver.onReceive(ParsePushBroadcastReceiver.java:125)
        at android.app.ActivityThread.handleReceiver(ActivityThread.java:3379)
        at android.app.ActivityThread.access$1200(ActivityThread.java:199) 
[...]

PR #927, opened 2019-02-18
Change to use non-deprecated methods for registering FCM push token

Jawnnypoo:
This allows us to no longer need to register two services to set up FCM push. We will need to update the docs as well in the docs repo. This should probably be in a 1.20.0 update, since it breaks compatibility if others were extending our instance ID service.


rogerhu:
https://medium.com/android-school/firebaseinstanceidservice-is-deprecated-50651f17a148 

Thanks!

NicksonYap:
Sorry to bring this up, but on

Parse-SDK-Android 1.20.0 (parse and fcm)
Samsung S9, Android 9 (pie)
SDK 28

`deviceToken` and `pushType` will be empty when added as new Installation

initially I did a workaround to manually get the token from `FirebaseInstanceId.getInstance().getInstanceId()` and then run `setDeviceToken` and save on the installation (took a very long time to figure this out)

Turns out, it's already handled by the library, but instructions were missing in docs (this PR removed it from the docs)

The following is required in manifest:

```
         <service
            android:name="com.parse.fcm.ParseFirebaseInstanceIdService"
            android:exported="true">
            <intent-filter>
               <action android:name="com.google.firebase.INSTANCE_ID_EVENT" />
            </intent-filter>
         </service>
```

and magically, the `deviceToken` and `pushType` will be set

Likely related to:
https://github.com/parse-community/Parse-SDK-Android/issues/886
https://github.com/parse-community/Parse-SDK-Android/issues/890

davimacedo:
@NicksonYap would you be willed to send a new PR with the change that can improve the docs regarding this?

FridaySG:
Thank you for catching that @NicksonYap :)

However, a separate service in the manifest is not necessary because ParseFirebaseMessagingService extends FirebaseMessagingService and therefore is called when a new deviceToken is generated. As seen here:

https://github.com/parse-community/Parse-SDK-Android/blob/7417db52564a763303eef6c79b2507cf3d5a5ef3/fcm/src/main/java/com/parse/fcm/ParseFirebaseMessagingService.java#L14

And referenced in the firebase docs:

https://firebase.google.com/docs/cloud-messaging/android/client#monitor-token-generation

The following should do the trick:


        <service
            android:name="com.parse.fcm.ParseFirebaseMessagingService">
            <intent-filter>
                <action android:name="com.google.firebase.MESSAGING_EVENT"/>
                <action android:name="com.google.firebase.INSTANCE_ID_EVENT" />
            </intent-filter>
        </service>

Although to be honest I'm not even sure the "INSTANCE_ID_EVENT" is necessary, as per the firebase docs linked above.

Here is a piece of code I added to my initial activity in order to continually monitor/update device token as necessary in the even of a bug resulting in the device token not being generated:

```
            FirebaseInstanceId.getInstance().getInstanceId()
                    .addOnCompleteListener(new OnCompleteListener<InstanceIdResult>() {
                        @Override
                        public void onComplete(@NonNull Task<InstanceIdResult> task) {
                            if (!task.isSuccessful()) {
                                Log.d("PUSH ID", "getInstanceId failed", task.getException());
                                return;
                            }

                            // Get new Instance ID token
                            String token = task.getResult().getToken();
                            installation.setDeviceToken(token);
                            installation.saveInBackground();
                            Log.d("PUSH ID", token);
                        }
                    });
```

PR #929, opened 2019-02-26
ParseQuery extensions

danielsanfr:
With my last PR #892 now we can use a more secure way of creating queries.

I updated the README to show you how to do this.

_Sorry for the delay in doing this PR again._


merlin-zaraza:
What is the release plan for this?

Jawnnypoo:
Thanks for the reminder, just released it as 1.20.0

PR #930, opened 2019-02-26
Expose useful constants

danielsanfr:
It is interesting to leave these constants exposed because they are heavily used in queries and this change does not cause any problems for the SDK.

Please, let me know if it's all right.


danielsanfr:
Hello everyone, please, let me know if you have any problems with this PR.

PR #939, opened 2019-04-10
Fixed fetchAllIfNeeded and fetchAllIfNeededInBackground limit

ovidiulaz7:
Fixed fetchAllIfNeeded and fetchAllIfNeededInBackground defaut Parse Query limit of 100 objects.

### DIFF
--- a/parse/src/main/java/com/parse/ParseObject.java
+++ b/parse/src/main/java/com/parse/ParseObject.java
@@ -967,7 +967,8 @@ public class ParseObject implements Parcelable {
         }
 
         final ParseQuery<T> query = ParseQuery.<T>getQuery(className)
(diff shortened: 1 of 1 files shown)


Jawnnypoo:
Curious what happens if it hits the 1,000 limit cap though

stromdotcom:
There isn't a limit on queries anymore, but there is the server limit I suppose.  

There is a still a problem here which is that the error message is cryptic if the number of results returned is not the same as the number of objects requested to be fetched.  

`Object id <objectId> does not exist`

That's what started us finding this problem in the first place.  I think either the fact that this is basically a query followed by a bunch of object updates under the hood should be documented, or at least the error message should explain a bit better what is going on here.

PR #943, opened 2019-04-17
Restore Bintray

rogerhu:
Adding this PR to get some feedback.  I've had mixed feelings for awhile about requiring people to add jitpack.io to use the Parse SDK.    It definitely saves developer time publishing artifacts but has created a lot of confusion since we still have the old artifacts available.

This change would move to use the new Maven publishing plugin.  Because Jetpack relies on the older Maven plugin and assumes `./gradlew install` to publish the artifacts, I've added a task that maps `./gradlew install` to `./gradlew publishToMavenLocal`.  For releases, we could automate the use of Bintray by bringing back `./gradlew bintrayUpload`.

We previously used artifacts as parse-android, fcm-android, so this will definitely change what we used to do for publishing artifacts on Bintray.

Wanted to get input on supporting Bintray again...

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,16 +7,20 @@ buildscript {
     dependencies {
         classpath 'com.android.tools.build:gradle:3.3.1'
         classpath 'org.kt3k.gradle.plugin:coveralls-gradle-plugin:2.8.2'
-        classpath 'com.github.dcendents:android-maven-gradle-plugin:2.1'
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
     }
 }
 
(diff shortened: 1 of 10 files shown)


Jawnnypoo:
Another big thing that was planned is to move all the repos into one repo, specifically moving the Parse Twitter and Facebook utils + ParseUI into this repo as their own modules, similar to what the iOS SDK has done. This would make it easier to test and make changes as well as have people contribute. Since LiveQuery is a niche thing, it should still remain in its own repo though.

Maybe it would be good to move to this mono repo first, so that if we are changing artifact distribution again, there is another valid reason for it.

PR #958, opened 2019-07-14
Feature/coroutines support

oliveiradev:
I think that use coroutines with parse sdk will be great, that support is very simple.

In README.md, I described in which my PR will help the community.


oliveiradev:
Awesome! Let me know if has something about coroutines that can be improved.

Jawnnypoo:
I just tagged 1.21.0 so it is available now!

PR #963, opened 2019-08-08
Non recursive pinning and other pinning functions

maaeps:
For my current project I needed some additional pinning features:
- pinning in a not recursive way (if pinning all child's is not necessary)
- forcing an unpin (I wrote in the JavaDoc that it is not recommended to use this)
- unpinning a ParseObject from all Pins (as well as a "force" variant again)

I would like to contribute this changes.

I tested the code in a test environment, a long term test was not done jet.

Please consider merging my changes.


maaeps:
As I had performance problems with the local datastore, I have not tested this features in a long term test.
Although I had no problems with this code changes, I close this request.

PR #979, opened 2019-10-22
expose client destroy

mtrezza:
Exposing `destroy` method to allow to re-initialize client.

There is an existing test case in which this is used.

https://github.com/parse-community/Parse-SDK-Android/blob/0cfc67e9a3bcedc714f0041ee60cafb40ed93520/parse/src/test/java/com/parse/ParseUserTest.java#L1637-L1638


mtrezza:
@Jawnnypoo what do you think, can we merge this?

PR #984, opened 2019-10-27
Parse Delegated Properties - List of Enum

bahaa-kallas:
While i am working on a project I noticed that we can't save ParseObject with list of enums as a value using the `listAttribute ` delegated property
for example if we have this Parse Class
```
@ParseClassName("Test")
class Test : ParseObject() {
  var enumList by listAttribute<UserGroups>()
}
```
and then we tried to save the object we get this error :
`java.lang.IllegalArgumentException: invalid type for ParseObject: class sy.edu.uok.admooncore.controllers.UserGroups` 
where `UserGroups `is enum class
i have checked the current set of delegated properties and i found that list delegated property it only support the common generic Types.
so i have added this delegated property and i tried to follow the same approach of storing enum values in the current `EnumParseDelegate ` (convert to lower case when saving and convert to upper case when retrieving)


danielsanfr:
Cool your implementation. I would just suggest adding a `try/catch`.

bahaa-kallas:
sorry for the late response 
@danielsanfr Thanks for your suggestion

rogerhu:
@Jawnnypoo thoughts of this PR?

bahaa-kallas:
as an update @rogerhu @Jawnnypoo 
i think this PR must be extended by making `ParseQuery` object capable of handling enum values as well
example:
`query.whereEqualTo("enumAttribute", SomeEnum.MY_VALUE)` (right now there is no overload of the function to accept Enum)
but i am still curious about why the current implementation of  `EnumParseDelegate` convert the enum value for lower case string because its common for enums that its stored as is (`MY_ENUM_VALUE`) and that's due to the fact that many programming languages offers the ability to restore the enum value especially in kotlin/java
for example :
`SomeEnum.valueOf("ENUM_VALUE_1")` it will return the enum value of the string (if it belongs to the Enum class)
i understand making this change will break compatibility previous versions of the SDK (if applied) but tbh this how it should be implemented from the beginning

Jawnnypoo:
Do we know if the iOS SDK allows for enums?

danielsanfr:
Hi @bahaa-kallas .
I was the one who implemented the `EnumParseDelegate`.
My suggestion to implement it this way, was because we are talking about a multi-platform SDK, and unfortunately so far I don't know any standard to represent a fully interoperable enum.
For example, in swift and C ++ (as I remember Qt) it is common to use CamelCase.

bahaa-kallas:
thanks for the reply @danielsanfr 
do you know how it's done in IOS sdk ? do we also convert the enum value to lowercase before storing it in the database.  (i never worked with IOS before and thus never tried the sdk there)

mtrezza:
Hi everyone, trying to revive discussion in this PR to either get it merged or close it.

@parse-community/android-sdk how would you evaluate the feature of this PR?

L3K0V:
Still needs some clarification how is done if it's done in the other SDKs. What do you think @mtrezza about closing it?

mtrezza:
Closing; this issues needs a proper discussion in a GitHub issue before a PR.

PR #985, opened 2019-10-27
Add update server url

mtrezza:
a) Added the possibility to set and get the current server URL without having to destroy and re-initialize the client. Example use case: domain failover handling.

b) Added docs note to `destroy` method that the client should not be destroyed while a rest call is in progress. This leads to an unhandled exception. Ideally the exception should be handled, so if anyone got the time they can look into that at some point.

### DIFF
--- a/parse/src/main/java/com/parse/Parse.java
+++ b/parse/src/main/java/com/parse/Parse.java
@@ -193,8 +193,9 @@ public class Parse {
 
     /**
      * Destroys this client and erases its local data store.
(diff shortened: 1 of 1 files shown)


mtrezza:
@Jawnnypoo any time estimate when this will be reviewed?

Jawnnypoo:
Did you verify that this works with your own project?

mtrezza:
@Jawnnypoo yes, we are even using this in production already.

PR #1004, opened 2020-02-12
Update README.md

eMelgooG:
Starting with Android 9 (API level 28), cleartext support is disabled by default so we need to set this property to true in order to be able to connect to the server.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,7 @@ The custom `Application` class must be registered in `AndroidManifest.xml`:
 ```xml
 <application
     android:name=".App"
+    android:usesCleartextTraffic="true"
     ...>
     ...
 </application>


Jawnnypoo:
I would not advocate for this. Instead, you should be hosting your Parse server on https. If you really needed cleartext to work, you could provide this in your own app, but in general, you should not be allowing cleartext traffic.

eMelgooG:
I still think it's worth mentioning because many people are using parse for testing and/or learning purposes without having a proper server and they might run into this problem as I did and many others around web and honestly I couldn't find a proper solution on the web and no one is mentioning this.

Jawnnypoo:
We could mention it with a comment or something denoting that you should only use `cleartext` while testing, and should use `https` when not testing.

eMelgooG:
That sounds like a great idea @Jawnnypoo . Let's do that.

Jawnnypoo:
Added to the main README, thanks for the PR @eMelgooG 👍

eMelgooG:
Looks great! Good job!

PR #1013, opened 2020-03-28
update to Bolts 1.5.0

rogerhu:


rogerhu:
I pushed a bolts-android version 1.15.0 to Bintray (but not sonatype) so let's see if this works.

PR #1018, opened 2020-04-18
Add bolts tasks to this library

rogerhu:
https://github.com/BoltsFramework/Bolts-Android is no longer really maintained/supported.  Moving it here gives us the remove the dependency and allow us to shift to possibly Kotlin coroutines long-term.


Jawnnypoo:
Awesome, can we make sure the source is formatting the same as the rest of the project? The project uses the default Android Studio formatting. This way, when users contribute, they can just use the auto-formatter. You should be able to scan the whole module and auto-format from within Android Studio 👍

PR #1022, opened 2020-04-21
Change to implementation

rogerhu:


Jawnnypoo:
I think this PR needs to be opened on top of master without the other changes too.

PR #1032, opened 2020-06-10
fix wrong appVersion

### LINKED ISSUE #902: Installation appVersion shows versionName instead of versionCode
**Issue:**
The Android SDK sets the app's `versionName` (e.g. `2.0`) instead of its `versionCode` (e.g. build `200`) in the `Installation` field `appVersion`.

**Expected:**
The build number (i.e. `versionCode`) should be set.

**Rationale:**
It makes sense to set the version code, because:
* server logic that is dependent on the build number can easily compare the `versionCode` being an integer, instead of `versionName` being a String.
* a build number more accurately identifies an app package than a version code.
* it contradicts the Parse iOS SDK which sets the app build number

However, correcting this inconsistency would be a breaking change for existing server logic.
Note: The API request header contains both, the build number and version name.

**Affected Code:**
https://github.com/parse-community/Parse-SDK-Android/blob/0cfc67e9a3bcedc714f0041ee60cafb40ed93520/parse/src/main/java/com/parse/ParseInstallation.java#L212

https://github.com/parse-community/Parse-SDK-Android/blob/0cfc67e9a3bcedc714f0041ee60cafb40ed93520/parse/src/test/java/com/parse/ParseInstallationTest.java#L292

mtrezza:
closes #902

**Note:** This is a breaking change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ### master
 
+> __BREAKING CHANGES__
+> - FIX: Corrected the `Installation` property `appVersion` to be the build version instead of the version name. This aligns the property with its equivalent in the Parse iOS SDK. See [#902](https://github.com/parse-community/parse-server/pull/902) for details. Thanks to [Manuel Trezza](https://github.com/mtrezza).
+
 - Update OkHttp version to allow for future Android API 30 compilation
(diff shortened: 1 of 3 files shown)


Jawnnypoo:
@mtrezza Please resolve the conflict

Issue #1028, opened 2020-05-16
Duplicate class bolts.* 


Azaratur:
This is my setting:
implementation "com.github.parse-community.Parse-SDK-Android:parse:1.24.1"
implementation "com.github.parse-community.Parse-SDK-Android:facebook:1.24.1"
implementation 'com.github.parse-community:ParseLiveQuery-Android:1.1.0'

And this is what i got:
Duplicate class bolts.Capture found in modules jetified-bolts-tasks-1.24.1.jar (com.github.parse-community.Parse-SDK-Android:bolts-tasks:1.24.1) and jetified-bolts-tasks-1.4.0.jar (com.parse.bolts:bolts-tasks:1.4.0)

Azaratur:
Temporally fixed by adding:
```
{
                exclude group: 'com.parse.bolts', module: 'bolts-tasks';
            }
```
and forking ParseLiveQuery, those are "pointing" to older ParseSDK

rogerhu:
Thanks for the bug report -- https://github.com/parse-community/ParseLiveQuery-Android/pull/107

pc-cesar:
i have the same erro error :Program type already present: bolts.AndroidExecutors

Libs:
 implementation 'com.github.parse-community.Parse-SDK-Android:parse:1.24.1'
    implementation "com.github.parse-community.Parse-SDK-Android:fcm:1.24.1"
    implementation 'com.github.parse-community.Parse-SDK-Android:facebook:1.24.1'
    implementation 'com.github.parse-community:ParseLiveQuery-Android:1.2.0'

i try --> exclude group: 'com.parse.bolts', module: 'bolts-tasks'; but not solved my problem

Azaratur:
Yes i still that issue too, but i did solve it whit this:
`implementation ("com.github.parse-community.Parse-SDK-Android:parse:1.24.1")
            {
                exclude group: 'com.parse.bolts', module: 'bolts-tasks';
            }
    implementation ("com.github.parse-community.Parse-SDK-Android:facebook:1.24.1")
            {
                exclude group: 'com.parse.bolts', module: 'bolts-tasks';
            }
    implementation ("com.github.parse-community.Parse-SDK-Android:bolts-tasks:1.24.1")
            {
                exclude group: 'com.parse.bolts', module: 'bolts-tasks';
            }
    implementation ('com.github.parse-community:ParseLiveQuery-Android:1.2.0')
            {
[...]

PR #1033, opened 2020-06-17
Fix bolts duplicate classes

mtrezza:
### Issue
The bolts-tasks module has been added natively to this library in #1018 without changing its name. This creates a naming conflict with transitively added legacy-bolts module.

### Solution
Renamed internalized package to `com.parse.boltsinternal`. This allows to keep the original module name `bolts-tasks` and if necessary other modules of the Bolts framework be internalized in the future without changing their names.

closes https://github.com/parse-community/Parse-SDK-Android/issues/1028


mtrezza:
Thanks for the fast release 👍

PR #1044, opened 2020-07-06
Add note in docs linking to local datastore issue

Jawnnypoo:
It is important for SDK users to understand the drawbacks of using Local Datastore when using the SDK. I think that we need to have a link to the open issue directly in the docs, so there are no surprises for users once they go down the route of using local datastore in their project. https://github.com/parse-community/Parse-SDK-Android/issues/279

### DIFF
--- a/parse/src/main/java/com/parse/Parse.java
+++ b/parse/src/main/java/com/parse/Parse.java
@@ -75,7 +75,8 @@ public class Parse {
      *   }
      * }
      * </pre>
-     *
+     * See <a href="https://github.com/parse-community/Parse-SDK-Android/issues/279">https://github.com/parse-community/Parse-SDK-Android/issues/279</a>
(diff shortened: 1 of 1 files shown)


mtrezza:
@TomWFox As the "docs expert", what is your opinion on linking Github issues directly in the docs? Anything we should consider?

TomWFox:
Seems like a good idea, probably for the iOS guide too.

I would say extracting the key points for the docs would be preferable but you could also provide a link for those who want to see the whole thread.

mtrezza:
@Jawnnypoo Do you think you could just briefly mention the key issue with the local datastore in the docs?

mtrezza:
I can then open an iOS PR to add the same comment there, thanks @TomWFox for suggesting.

mtrezza (approved):

PR #1045, opened 2020-07-07
Add README for bolts module

Jawnnypoo:
This adds clarification to why the module exists.


bolts-tasks/README.md
```diff
@@ -0,0 +1,4 @@
+## bolts-tasks
+Fork of [Bolts-Android](https://github.com/BoltsFramework/Bolts-Android), since it is no longer
+really maintained/supported. Moving it here gives us the remove the dependency and allow us to shift
```
mtrezza, line 3: > Moving it here gives us the remove the dependency

I think there are some words missing?

Jawnnypoo: Oops, fixed.

mtrezza: Do we already know with what we want to replace Bolts? So that if contributors make any changes they should already use the new approach?

PR #1047, opened 2020-07-10
RxJava 3 support

Jawnnypoo:
This adds a module which adds extension methods to transform `Task`s into RxJava types (Single and Completable). This allows for those who are familiar with RxJava and not Bolts to use Parse effectively. It is presented as a Kotlin extension, but can be called from Java as well. See the README for more info.


Jawnnypoo:
Some background: I have used these extensions in production apps before, they should be safe to merge for those who are interested in them. Let me know what you all think!

PR #1054, opened 2020-07-31
Rename functions that cause shadow members

### LINKED ISSUE #1016: Cannot use coroutines extension function save
Hello,
I've import parse sdk and parse coroutines with 
``` gradle
//Parse SDK
implementation "com.github.parse-community.Parse-SDK-Android:parse:1.23.1"
//Parse coroutines
implementation "com.github.parse-community.Parse-SDK-Android:coroutines:1.23.1"
```

And when I try to save a `ParseObject` with a coroutine, it use the parse method `save` and not the extension one. This is because they have the same name so the method win over the extension.

If I made a mistake when add dependency in my android project, I apologize but I tried with the _README.md_ which isn't clear enought.

Thank you.

P.S: In the [README.md](https://github.com/parse-community/Parse-SDK-Android/tree/master/coroutines#parseuser) it says for `signUp` :

> ``` kotlin
> launch { // Coroutine builder
>     user = ParseUser().apply {
>         setUsername("my name")
>         setPassword("my pass")
>         setEmail("email@example.com")
>     }.also {
>         signUp()
>     }
> }
> ```

but in fact the function is `singUp`.

oliveiradev:
### Problem

Some extension functions cause `shadow member function`, it means that the member function of the extended class always have the preference when the IDE shows suggestions.

### Porpuse

I renamed some extension function's name to use `coroutine` prefix, it solves the problem but I do not know if this name is the best option.

I'm open to suggestions

Resolve #1016

### DIFF
--- a/coroutines/README.md
+++ b/coroutines/README.md
@@ -18,11 +18,11 @@ Now we can call a parse query using a synchronous style, this is possible when w
 
 ```kotlin
 launch { // Coroutine builder
-    val cat = ParseQuery.getQuery(...).find()
+    val cat = ParseQuery.getQuery(...).coroutineFind()
     // get cats without callback
 }
 ```
-We use a coroutine builder because `find()` is a suspend function.
+We use a coroutine builder because `coroutineFind()` is a suspend function.
 
 We can also, use a function like a coroutine builder, it will be provider us a flexibility call our query without any extensions function.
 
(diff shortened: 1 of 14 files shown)


Jawnnypoo:
I think that `suspend` could be a better prefix. While we are making changes here, can we also change the package of all these functions to `com.parse.coroutines` with all these methods in the same package? (can still be in different files)

oliveiradev:
Good, I think that it makes sense

Jawnnypoo:
Thanks a bunch @oliveiradev !

Issue #1007, opened 2020-02-24
Android Application freezes up completely when using saveEventually()


tutorbear:
### Issue Description

saveEventually() is suppose to save objects using relations recursively just like saveInBackground does. The logic of my code is very simple and straight forward. I have a logged in user. I create a new ParseObject (Profile) and associate it with the user. Then when I call saveEventually The entire app freezes up. 

### Steps to reproduce

1) Signup and login.
2) Put this code on the onClick listener of a button in android.

      ` Parse user = ParseUser.getCurrentUser();`
       ` //A profile object`
       `profile = new ParseObject("StudentProfile");`
      `  profile.put("user", user);`
      `  profile.put("name", "John");`
       ` profile.setACL(new ParseACL(user));`
        `//Put profile object into user`
       ` user.put("profile",profile);`
        `user.saveEventually();`
### Expected Results

user saved along with new profile object while pointing towards it.

### Actual Outcome

App Freezes

### Environment Setup
  - parse-server version: 3.9.0
  - Hosted on Back4app

### Logs/Trace

Non , App freezes up and stops responding

Jawnnypoo:
Can you pinpoint the exact line that causes the freeze, and can you see if `saveInBackground` causes the same freeze?

ramiro-ciocca:
@jawnnypoo, I don't know if @tutorbear problem is related with the issues I have opened about LDS and saves.
I see that @tutorbear  is trying to save an object with a pointer to an usaved object, maybe that is his problem.
But, there is a real issue with LDS and saves.  If you LDS grows enough, and you try to save some complex object, with some pointers to other objects, then the app hangs.
I explained in #1006 and in #993 that what causes the hangs is a recursive method inside ParseObject that inefficiently traverse all the LDS with each save.

tutorbear:
@Jawnnypoo 
Hello.
SaveInBackgound() works fine, its save eventually thats causing the issue. I have tried it with both LDS enabled and disabled , on both occasion the app freezes up. And I have found the exactly whats wrong.

`Parse user = ParseUser.getCurrentUser();`
`//A profile object`
`profile = new ParseObject("StudentProfile");`

`// THIS LINE is the curprit----------------------->
`profile.put("user", user);`
`//-------------------------------------->

`profile.put("name", "John");`
`profile.setACL(new ParseACL(user));
`//Put profile object into user
`user.put("profile",profile);
`user.saveEventually();



No error is givin in the log cat, only this is shown
[...]

Issue #1006, opened 2020-02-23
Desesperately need cascadeSave like parameter for the Android SDK


ramiro-ciocca:
When Local Data Store is enabled and you have very complex ParseObjects with pointers to other ParseObjects, save methods insice ParseObject.java go mad and it can take minutes to complete a simple save.
I have tracked down the issue till line 662 of ParseObject.java, where is called collectDirtyChildren that enters and insane almost infinite loop traversing everty single attribute of every single nested object.
I understand that is a functionallity that many developers would use, but we need to say "we dont want to save children objects" to the save method, just like in JS SDK, with the cascadeSave=false.
I'm forced to write cloud code for every single save, or disallow LDS.
Please, if I could comment that line I think that would be the solution, but it is into an aar file, I dont know how to modify that.
Thanks!

ramiro-ciocca:
I have implemented the "cascadeSave=false" parameter for the Android SDK in a very hacky way.
I cloned the repos, then edited ParseObject.java, then commented out line 630, "collectDirtyChildren(object.estimatedData, dirtyChildren, dirtyFiles, seen, seenNew);".

That recursive collectDirtyChildren method is really messy and is the culpruit for all the hangs when saving any object.

The ParseObject class should provide a paramter to choose wether or not save the children objects, but most important, it should have that method fixed, it hangs every device when you try to save any complex object, with some nested children.  It is because of the LDS.  With my modification I can use the LDS and save objects without hangs.

tutorbear:
Dear ramiro-ciocca ,
Can you please share the modified sdk ? And also how can I use it my android project.
I Have the EXACT same issue.
A user class with a pointer to anthoer parse Object created locally,Whenver i try to save the parse user but saveEventually my app completely freezes up.

ramiro-ciocca:
I know, that save issue is a nigthmare, it hangs the app on every device or it takes minutes to complete a save.
And saveEventually doesn't even really save the data in the database, it kind of schedules the save until the device acquires connection, and at that time you will have the same hang issue.
I found out that the culprit was a buggy messy recursive method in ParseObject.java, that runs when you have the local data store is acitvated.  For every save for an object with some grade of complexity, lilke pointers to other objects (and no matter what kind of object, ParseUser, or generic) that method traverse the entire local data store looking for dirty children, on every single property of every pinned object, in objects not realted at all with your current save, I mean, the entire LDS, in an exagerated inefficient way.  
[...]

Issue #993, opened 2019-11-25
Making any call to 'ParseUser.saveInBackground()' on the UI Thread freezes app


bstillitano:
I have noticed that no matter what, even on a fresh app install, if you call `ParseUser.getCurrentUser().saveInBackground()' on a UI Thread the app will hang. The example I used was calling the function after updating a user value after toggling a Switch component:

```
fun setAccountPrivacy(on: Boolean) {
    ParseUser.getCurrentUser().put("privateAccount", on)
    ParseUser.getCurrentUser().saveInBackground()
}
```

I've written a basic custom activity to prove my hypothesis. Starting an Activity of this type will cause the app to hang and leave a blank screen as can be seen by simulator screenshot attached. Activity is as follows:

```
import androidx.appcompat.app.AppCompatActivity
import com.parse.ParseUser

open class CustomActivity : AppCompatActivity() {
    override fun onResume() {
        super.onResume()

        //Save Parse User
        ParseUser.getCurrentUser().saveInBackground()
    }
}
```

**After Running**
![Screenshot_1574686165](https://user-images.githubusercontent.com/40552246/69541849-46035000-0fde-11ea-9562-f57ce16a5f80.png)

**If commenting out the saveInBackground() function**
![Screenshot_1574686306](https://user-images.githubusercontent.com/40552246/69541976-8d89dc00-0fde-11ea-84c3-28f8365bb94a.png)

Jawnnypoo:
I believe `ParseUser.getCurrentUser()` can be an expensive operation. It is best to instead create a task/background operation that calls `getCurrentUser` and then saves, all in the background.

bstillitano:
@Jawnnypoo thanks for the suggestion as well as taking the time to comment and help. I’ll give this a shot this evening and let you know how I go and what a tried.

bstillitano:
@Jawnnypoo so I've mucked around with this and it hasn't resolved the issue. I did

```AsyncTask.execute {
    ParseUser.getCurrentUser().saveInBackground()
}```

I still end up with the same results. After some testing, the `ParseUser.getCurrentUser()` on its own doesn't seem to be causing the issue. It only hangs once I add the `saveInBackground()`. I'm at a real loss as this is causing me massive issues in getting a release out.

Jawnnypoo:
Did you try `ParseUser.getCurrentUser().save()` since you are already on a background thread with the AsyncTask?

ramiro-ciocca:
[...]

Issue #888, opened 2018-09-24
ParseObject.save() stuck in infinite loop.


mik9:
We have cloud function that returns object list (20 items, Post). It displayed in RecyclerView with pagination.
After I scroll a lot (600+ items) and try to create new object (Comment, that have reference to Post) it never returns from `save()`. I'm calling it on background thread (via RxJava).

After I attached profiler I saw it stuck in a loop:
<img width="1107" alt="10 42 47" src="https://user-images.githubusercontent.com/244760/45941405-bb445100-bfe6-11e8-924c-a5ce388ce9b6.png">
Or another:
<img width="1139" alt="10 47 41" src="https://user-images.githubusercontent.com/244760/45941539-4d4c5980-bfe7-11e8-86c3-b1b39aefb35c.png">

Used latest library version. Can share debug apk privately.

mik9:
After removed `.enableLocalDataStore()` issues disappears.

L3K0V:
As @shlusiak mention his #1058 PR most likely is improving even fixing this, so I close it.

PR #1058, opened 2020-08-28
perf: remove unnecessary extra child traversal in collectDirtyChildren

shlusiak:
Please see https://github.com/parse-community/Parse-SDK-Android/issues/1006#issuecomment-682448919 for the analysis of this.

The `collectDirtyChildren` forgets seen siblings when traversing, which causes extra traversals if those siblings point to the same objects and causes lots of garbage to be created in memory. It also has the potential of adding the same dirty object twice to the `dirtyChildren` collection, which may or may not be a set.

Removing the copy of the `seen` set may improve performance significantly.

Should hopefully massively improve the situations described in issues #1006, #1007, possibly #888 and #993 .


shlusiak:
The function in question actually has a few more issues:
- It uses a `HashSet` for `seen`, which means that the object's `equals()` and `hashCode()` functions will be used. So if two different ParseObjects implement equals over a set of attributes other than the objectId, this code will wrongly detect a cycle. Note that the `ParseTraverser` uses a `IdentityHashMap`.
- Whenever a ParseObject is visited, recursion is used to traverse this object. However, by calling `collectDirtyChildren`, a new `ParseTraverser` is instantiated with the current object as the root, which then of course has forgotten all of the previously visited objects and traverse them all again. This is obviously bad, I can't see how this would cause an infinite loop, but it is for sure exploding the algorithm to a point where it may never finish.

mtrezza:
@shlusiak Apologies for the late reply to your PR, we are finally looking to close all PRs. Could you merge master into this and resolve any conflicts?

mtrezza:
Thanks! Could you also add a test that to prevent the issue from occurring in the future?

mtrezza:
@shlusiak I see, looks like not worth the effort because difficult to test. So let's merge as is.

L3K0V (approved):
@mtrezza LGTM

mtrezza (approved):

parseplatformorg:
🎉 This change has been released in version [2.0.4](https://github.com/parse-community/Parse-SDK-Android/releases/tag/2.0.4)

Issue #1061, opened 2020-09-04
ConcurrentModificationException in ParseTraverser


shlusiak:
There is still the occasional `ConcurrentModificationException` in ParseTraverser.

Parse version: `1.25.0`

Stack trace:

```
Fatal Exception: java.util.ConcurrentModificationException
       at java.util.HashMap$HashIterator.nextNode(HashMap.java:1441)
       at java.util.HashMap$KeyIterator.next(HashMap.java:1465)
       at java.util.Collections$UnmodifiableCollection$1.next(Collections.java:1084)
       at com.parse.ParseTraverser.traverseInternal(ParseTraverser.java:101)
       at com.parse.ParseTraverser.traverseInternal(ParseTraverser.java:95)
       at com.parse.ParseTraverser.traverseInternal(ParseTraverser.java:102)
       at com.parse.ParseTraverser.traverse(ParseTraverser.java:140)
       at com.parse.OfflineStore.saveLocallyAsync(OfflineStore.java:666)
       at com.parse.OfflineStore.access$1300(OfflineStore.java:36)
```

The issue appears that while `ParseObject.keySet()` returns a `Collections.unmodifiableSet()`, the set is a read-through to the underlying HashMap, which is not synchronised. 

https://github.com/parse-community/Parse-SDK-Android/blob/797786ff42223094c20d2dadd7af9e616020175a/parse/src/main/java/com/parse/ParseObject.java#L1489-L1497

I believe we either need to make the `estimatedData` a synchronised Map, or we need to work on a copy of the set when traversing, to guard against concurrent modifications. Thoughts about making estimatedData a synchronized map?

https://github.com/parse-community/Parse-SDK-Android/blob/797786ff42223094c20d2dadd7af9e616020175a/parse/src/main/java/com/parse/ParseTraverser.java#L98-L103

mtrezza:
Closing via #1062

PR #1062, opened 2020-09-05
Create a copy of the ParseObject's keySet when iterating over it.

shlusiak:
Should fix the `ConcurrentModificationException` in #1061 by creating a local copy before iterating over the `ParseObject`'s keys, with the cost of potentially missing some newly added keys or traversing into a `null` value.

### DIFF
--- a/parse/src/main/java/com/parse/ParseObject.java
+++ b/parse/src/main/java/com/parse/ParseObject.java
@@ -1489,6 +1489,9 @@ public class ParseObject implements Parcelable {
     /**
      * Returns a set view of the keys contained in this object. This does not include createdAt,
      * updatedAt, authData, or objectId. It does include things like username and ACL.
+     *
+     * Note that while the returned set is unmodifiable, it is in fact not thread-safe, and creating a copy
+     * is recommended before iterating over it.
      */
     public Set<String> keySet() {
         synchronized (mutex) {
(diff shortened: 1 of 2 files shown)


JayMad18 (approved):

shlusiak:
Fixed merge conflicts. Previously the app could crash if keys are added or removed while the object is being traversed. With this change, at time of traversal a copy of the keyset is created that is traversed instead, fixing the race conditions of concurrent access. This does not resolve other issues of concurrent access, e.g. adding a key while traversal is happening will miss the newly added key, but also removing a key while traversal is happening would see a key where the value does not exist any more (`null`) and tries to traverse into it. That should work, as the traverseInternal checks for that, but it is not ideal.

L3K0V:
@mtrezza may you change the title from

`fix: create a copy of the ParseObject's keySet when iterating over it.` 

to 

`fix: race condition when keys are added or removed while the object is being traversed`

L3K0V:
> First-time contributors need a maintainer to approve running workflows. 

Can we have the CI running to see the checks passed @mtrezza?

L3K0V:
CONTRIBUTING.md is already updated. If you think it should be rephrased, let me know @mtrezza

shlusiak:
Ran `spotlessApply`, which seems to reorder imports and fixes comments to satisfy lint.

L3K0V:
> @L3K0V Does ./gradlew spotlessApply do more than lint? What do you think about renaming this to ./gradlew lint-fix, like we have in parse server, dashboard, etc for consistency? Or is spotlessApply a term an Android developer can be expected to understand?

Spotless is widely spread across Java and Kotlin developers. There's standard `lint` task, and I can make link run spotlessApply but it's not part of this PR. I will consider opening a separate one. 

Meanwhile I will be glad if this and #1058 are merged soon, as they seems to fix a lot of issues.

PR #1065, opened 2020-09-30
feature: Facebook coroutines wrapper

phillwiggins:
- Change version code location in gradle
- Add coroutines wrapper for facebook login

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,10 @@
 buildscript {
-    ext.kotlin_version = "1.3.72"
+
+    ext.versions = [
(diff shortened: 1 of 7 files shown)


Jawnnypoo:
Hey there, since this is a primarily Java project still, we don't want to ship Kotlin + Coroutines via `api` for anyone who depends on this library.

phillwiggins:
Would implementation be a better option in this instance?

Jawnnypoo:
I think we would potentially (unfortunately) make a different dependency altogether that adds the ktx bindings

PR #1076, opened 2021-01-15
fix TypeCastException when unlinking google account

timfreiheit:
fixes `kotlin.TypeCastException: null cannot be cast to non-null type com.parse.ParseException`

when trying to unlink the `ParseUser` from Google.

The `Task` returns `null` as the `error` when the task was completed successfully.
This is the same behavior as in the [ParseFacebookUtils](https://github.com/parse-community/Parse-SDK-Android/blob/master/facebook/src/main/java/com/parse/facebook/ParseFacebookUtils.java) but was handled incorrectly here.


Jawnnypoo (approved):
Thanks for the fix!

PR #1084, opened 2021-03-11
Fixed: resuming continuation twice when task fails

### LINKED ISSUE #1083: IllegalStateException: Already resumed in Task to coroutines
Issue is here https://github.com/parse-community/Parse-SDK-Android/blob/3a6fde5b8040568c3f2d0fcd753c43580b34a268/coroutines/src/main/java/com/parse/coroutines/ParseTaskExtensions.kt#L18

and down below, maybe in more places.

Problem it's missing an early return from the lambda, or the second continuation inside an else.

EDIT: typo

ecdpalma:
Fixes #1083.


Jawnnypoo (approved):
Nice fix!

Issue #1087, opened 2021-04-09
Support for custom object ids


martinpfannemueller:
Using the allowCustomObjectId option of parse-server, it is possible to provide custom object ids on the client side. It seems to me, that there is no support for this yet in the Android SDK. I already tried to implement this by simply creating POST requests instead of a PUT in case allowCustomObjectId is true and an objectId is provided (see https://github.com/martinpfannemueller/Parse-SDK-Android). However, as I do not have the complete overview of the SDK, this currently does not really work yet.

Maybe there is somebody with more experience in developing the Android SDK who can provide this relatively simple feature more easily. In case I make progress, I will reference this issue in a possible pull request.

cbaker6:
Though I haven’t used the Android SDK, I can tell from the code that the Android, IOS, and JS SDKs are designed the same. Because of this, the objectId in Android is probably never added to the encoded data sent to the parse server. I recommend looking at the JS PR I posted in response to your question I’m the iOS repo for guidance with fixing this here.

L3K0V:
This was released with [2.1.0](https://github.com/parse-community/Parse-SDK-Android/blob/master/CHANGELOG.md#210-2021-11-21)

PR #1088, opened 2021-04-12
feat: add support for custom objectId

martinpfannemueller:
I implemented the custom objectId feature mentioned in #1087 including unit tests. Is there something missing to get this merged?

Thanks!


martinpfannemueller:
I just merged the changes from the master. Again, locally everything runs just fine.

martinpfannemueller:
Still the same problem. This maybe has to do something with the OS or the development environment then. I am running the tests inside Android Studio on macOS. In general, do you have an idea how we can debug the two failing tests in the CI pipeline?

martinpfannemueller:
I just tried to update the unit tests as it was partly possible to reproduce the failing tests. It definitely has something to do with the Parse instance. Therefore, I tried to clean them up properly now in the new tests at the end.

mtrezza:
> It would definitely help if you could provide some guidelines how the mocking and the different plugins as well as the different states of the mocked objects work together in the tests, e.g., by providing a sample when using the local datastore or when a specific response is expected.

@L3K0V @Jawnnypoo could you do that or should I look into it?

Jawnnypoo:
I don't have context into that, sorry.

L3K0V (approved):
@mtrezza @martinpfannemueller LGTM

mtrezza (changes requested):
Could you please also add doc headers to the new public properties and methods?

L3K0V:
@martinpfannemueller gentle ping when you have some free time if is possible to add some JavaDoc as @mtrezza pleases, so this can be merged. Looking forward to see this merged 🤞

martinpfannemueller:
I just added some JavaDoc to allowCustomObjectId() in the configuration builder and the isAllowCustomObjectId() getter.

parse/src/main/java/com/parse/Parse.java
```diff
@@ -657,6 +669,20 @@ private Builder setLocalDatastoreEnabled(boolean enabled) {
                 return this;
             }
 
+            /**
+             * Enable the possibility to set a custom objectId for your ParseObjects.
```
mtrezza, line 673: Could you shorten this to:

> Allow to set a custom objectId for ParseObjects.

mtrezza:
There also seems to be a lint fix needed.

martinpfannemueller:
I updated the JavaDoc and ran spotlessApply

mtrezza:
Thanks, let just see whether the CI passes.

mtrezza (approved):
Looks good, thanks for this PR!

parseplatformorg:
🎉 This change has been released in version [2.1.0](https://github.com/parse-community/Parse-SDK-Android/releases/tag/2.1.0)

PR #1095, opened 2021-07-02
Update all dependencies

L3K0V:
This is first of many PRs which I want to introduce in order to get this SDK up to date. 
First step in my master plan is update all dependencies, Gradle version and used plugins. I prefer to make a small incremental steps so the PRs to be easy to review.

Updating of the Firebase Messaging should address https://github.com/parse-community/Parse-SDK-Android/issues/1071 and more precisely https://github.com/parse-community/Parse-SDK-Android/issues/1071#issuecomment-840608858

PR #1099, opened 2021-08-19
Migrate the source base to Kotlin

L3K0V:
@mtrezza in continuation of https://github.com/parse-community/Parse-SDK-Android/pull/1095 I started to port the current Java files to Kotlin. As I mentioned in the previous PR my goal would be to bring this SDK up-to-date.


mtrezza:
That is amazing and thanks for continuing this effort!

Please let us know whatever you need and how we can support you in that.

Would you mind opening an issue for this and reference this PR and the previous PR? So we can have conceptual discussions there (regardless of the PR) and implementation related discussions here. And maybe you could just roughly list the planned steps there, so we are all on the same page and make it easier for others to join the discussion. PRs tend to get less exposure than issues.

bahaa-kallas:
@L3K0V @mtrezza 
What really happened? Did you changed your mind about switching to kotlin? 
Was there any blockers ? or it was simply not possible to make it multi-platform ready.
I would love to know some details about your final decision.

L3K0V:
@bahaa-kallas migration to Kotlin of this huge source base is really big effort and I had few attempts. For sure it not for me alone doing it in my spare time. If I find someone with whom we can join forces it will be great. I also consider writing from scratch, but for this I need also to focus on it and it's tricky parts - encoding, decoding, traversals and all other challenges down the road.

Additionally I have a working PoC, brand new GraphQL-based SDK but it provides very limited functionality compared to this REST SDK, just because the GraphQL support on the Parse Server is very limited and kind of poor.

The KMM path will be very long and depends on the path taken. I suggest to continue and discuss everything else in #1100

PR #1101, opened 2021-09-19
Update README.md

mobilekosmos:
Updated gradle instructions according to new Android changes. Also added Kotlin version of code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,14 +12,15 @@ For more information about Parse and its features, see [the website](https://par
 
 ## Dependency
 
-Add this in your root `build.gradle` file (**not** your module `build.gradle` file):
+Add the maven repository where this library is stored (jitpack.io) in your root `settings.gradle` file:
 
-```gradle
-allprojects {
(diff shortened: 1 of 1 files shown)


mobilekosmos:
FAIL_ON_PROJECT_REPOS: If this mode is set, any repository declared directly in a project, either directly or via a plugin, will trigger a build error.

We want to declare the repositories for all project centraly, so to avoid modules adds repositories which are anyways already declared centrally this flag fits the purpose I think. If for some reasson a module needs an extra repository some day you can change that flag anytime.

https://docs.gradle.org/current/javadoc/org/gradle/api/initialization/resolve/RepositoriesMode.html

mobilekosmos:
[PR #1122](https://github.com/parse-community/Parse-SDK-Android/pull/1122)

mobilekosmos:
I got you now. Partially, you don't put dependencyResolutionManagement  in the gradlge file, I will update the PR.

PR #1102, opened 2021-09-29
Updated dependencies + changes accordingly.

mobilekosmos:
Important notes:
- minSDK set to API 28 as I prioritize security and adopting new standards over supporting old insecure platforms.
- Old module GCM removed.
- jacoco disabled as using the compiled version over jitpack leads to compilation errors which I could not fix so far.

### DIFF
--- a/bolts-tasks/build.gradle
+++ b/bolts-tasks/build.gradle
@@ -3,8 +3,8 @@
 // This source code is licensed under the MIT license found in the/
 // LICENSE file in the root directory of this source tree.
 
-apply plugin: 'java'
-apply plugin: 'maven'
+apply plugin: 'java-library'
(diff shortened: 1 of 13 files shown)


Jawnnypoo:
14 to 28 as the minSdk is a huge jump, I highly recommend setting it to something more reasonable like 21 or 23

mobilekosmos:
I deleted the PR because it is flawed

mobilekosmos:
@Jawnnypoo Jawnnypoo after making the updates it seems that now each module needs maven-publish, without it the modules are not getting published, so I'm wondering how come all modules from master are published on maven and accessible through jitpack but none of the modules besides bolts has the maven plugin in the build.gradle file? How is then the publishing working in the original file?

PR #1103, opened 2021-09-30
Updated dependencies + related changes.

mobilekosmos:
Notes: 
1. minSDK changed to 21 (In my opinion there is no reasson to support plattforms which are not even supported by their producers anymore and are full of security flaws, so I would recommend to go even higher than 21).
2. Jacoco stuff also updated but uncommented because I was getting compilation issues in my testing proyect which I was not able to solve.
3. Published modules are published without javadoc, TODO in code.


mobilekosmos:
The build is failing, I took a look but I'm not experienced with that system, locally it compiles without problem, and using it in a test proyect also work.
Btw. there is one warning in build output: "No build-tools version is specified in android.components. Consider adding one of:" Since some time this is not required anymore, only optional, is the build system up to date? It says The following versions are pre-installed:
build-tools-25.0.2, so this is a really old version already.

mtrezza:
Is there anything that can/should be taken from this PR? Or is it better to close it and start fresh, given the many changes that were merged in #1095?

L3K0V:
@mtrezza let's close it and I will encourage @mobilekosmos to check again and open PR with further improvements if any

mobilekosmos:
Oh noooo, I worked like 1 week updating all, #1095 doesn't even update the gradle plugin.

mtrezza:
reopened 🙂 do you want to bring this up-to-date with master?

mobilekosmos:
> [Android 12 was released in beta in mid-May 2021, and Google plans to officially retire Android 9 in the fall of 2021](https://www.tomsguide.com/uk/us/old-phones-unsafe,news-24846.html).

Despite above declaration I won't go to that extrem, since it's not my library I would say minSDK=26 (Android 8.0), that's like already 6 years old. Would it be my own library I would be more agressive and set mindSDK=28

Btw. I made another branch where I will try to merge the current master. Will then do a new PR.

L3K0V:
I'm also **for** updating the version to SDK 21 but also OkHttp should be updated as well for sure when updating the SDK version. @mobilekosmos do you want to handle this in your PR - updating do SDK 21 and updating the OkHttp to latest version?

mtrezza:
How we set the API versions liked requires a discussion. I opened https://github.com/parse-community/Parse-SDK-Android/issues/1123 for that.

Issue #1104, opened 2021-10-05
Deprecating Facebook webview 


kaushikmurthy:
https://developers.facebook.com/docs/facebook-login/android/deprecating-webviews

Since Facebook is deprecating the webview, I hope this is going to be taken care in the future SDK releases.
If it is already taken care, please let me know the version of the SDK.

mtrezza:
Closing via #1105

PR #1105, opened 2021-10-09
fix: upgrade Facebook SDK

sommye-ctr:
In accordance with #1104 I changed the Facebook for Android SDK version from `6.3.0` to `8.2.0`.
Also, as recommended by RoboElectric, I have updated its version to `4.2.1` to fix the parsing in windows.

### DIFF
--- a/facebook/build.gradle
+++ b/facebook/build.gradle
@@ -33,3 +33,4 @@ dependencies {
     testImplementation "junit:junit:4.13.2"
     testImplementation "org.mockito:mockito-core:3.9.0"
     testImplementation "org.robolectric:robolectric:4.6"
+}


mtrezza:
Thanks for the PR! Let's merge https://github.com/parse-community/Parse-SDK-Android/pull/1095 first so we also have the CI working.

L3K0V:
@mtrezza as this PR overlaps a lot with #1095, may I update the Facebook SDK there as well? What do you think?

mtrezza:
Interesting that this only required to bump the SDK but no code changes.

sommye-ctr:
I think I'll be working on a project using the Facebook SDK soon. I'll update if there's any problem.

mtrezza:
I should also be able to test in the next days. I merged it anyway for an alpha release and if really FB doesn’t work, then we are actually missing some tests.

Issue #401, opened 2016-02-19
Failed signup attempt with anonymous ParseUser leaves it in inconsistent state


yuzeh:
Hi,

Whenever a signup fails with an anonymous ParseUser, I've noticed a couple of things:
- The app will still have a session token for that user which has been invalidated by the server (any CloudFunction call will error with an "invalid session token" message).
- The user is not considered anonymous anymore! Calling `ParseAnonymousUtils.isLinked(ParseUser.current())` returns false.

Reproduction steps:
- Enable anonymous users.
- Save the current (anonymous) user to the backend. (Not sure if this is important, but this is what our app will do).
- Trigger a failed signup attempt. I've only tried signing up with a username that already exists.
- Read the value of `ParseAnonymousUtils.isLinked(ParseUser.current())`.

From my understanding, the last value should be true, but is actually false.

These things are troubling because we are using `ParseAnonymousUtils.isLinked(ParseUser.current())` to determine if the user is currently ~~logged in~~ registered (user is ~~logged in~~ registered iff that returns false).

Two questions:
- Is `!ParseAnonymousUtils.isLinked(ParseUser.current())` the right way to determine if a previously anonymous user is ~~logged in~~ registered?
- If not, what is the right way? If so, is the current behavior I'm seeing in fact a bug?

yuzeh:
Did some more investigating into the source code:
* Looks like saving the user to the backend before attempting signup is important to triggering this issue: https://github.com/ParsePlatform/Parse-SDK-Android/blob/7d908f3d1cbed3c89addfac49419c6c2f1859d62/Parse/src/main/java/com/parse/ParseUser.java#L604
* And removing anonymity before signup is intentional: https://github.com/ParsePlatform/Parse-SDK-Android/blob/7d908f3d1cbed3c89addfac49419c6c2f1859d62/Parse/src/main/java/com/parse/ParseUser.java#L420

Seems like the answer to my first question is no. My current workaround is to have custom client-side state to keep track of whether the user is logged in or not. Is there a better way?

Tom-Fox1:
Closing this as no one has reported this issue recently, and we don't have the resources to investigate these historic reports.

Please ask me to reopen if you experience this issue in the latest version of this SDK.

PR #1136, opened 2021-10-27
fix: failed signup attempt with anonymous ParseUser leaves it in inconsistent state

shlusiak:
### Issue Description
The key "authData" contains a `Map` of authentication related data. Because unfortunately `estimatedData` is a shallow copy of `state`, the object pointed to by `state["authData"]` and `estimatedData["authData"]` is the same object. Modifying the data in this map directly prevents us from being able to call `revert()` to revert the data back to the server state.

This is impotant e.g. when trying to revert the user after converting it from anonymous to registered fails, where anonymity has been stripped and there is no way to get it back. 

Related issue: #401

### Approach
This would just create another (shallow) copy of "authData", which in this case would be fine, because we are only modifying the first layer. By replacing the value of the map with a new map we ensure that the values in `state` and `estimatedData` are different and that `revert()` works.

### DIFF
--- a/parse/src/main/java/com/parse/ParseUser.java
+++ b/parse/src/main/java/com/parse/ParseUser.java
@@ -757,7 +757,7 @@ public class ParseUser extends ParseObject {
 
     /* package */ void putAuthData(String authType, Map<String, String> authData) {
         synchronized (mutex) {
-            Map<String, Map<String, String>> newAuthData = getAuthData();
+            Map<String, Map<String, String>> newAuthData = new HashMap<>(getAuthData());
             newAuthData.put(authType, authData);
             performPut(KEY_AUTH_DATA, newAuthData);
         }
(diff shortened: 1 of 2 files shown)


L3K0V (approved):
@shlusiak great job again, LGTM!

L3K0V:
@shlusiak can you check please the failing test

```
com.parse.ParseUserTest > testSaveAsyncWithLazyAndCurrentUser FAILED
1579
    java.lang.AssertionError at ParseUserTest.java:962
```

shlusiak:
@L3K0V I think I fixed the failing test. Let's see if it passes now.

L3K0V:
@mtrezza LGTM

mtrezza (approved):
Look good!

parseplatformorg:
🎉 This change has been released in version [2.0.5](https://github.com/parse-community/Parse-SDK-Android/releases/tag/2.0.5)

Issue #1158, opened 2022-04-05
Users are logged out after SDK upgrade


PLR2388:
### New Issue Checklist

- [X] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-SDK-Android/security/policy).
- [X] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [X] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-Android/issues?q=is%3Aissue).
- [X] I can reproduce the issue with the latest version of [Parse Server](https://github.com/parse-community/parse-server/releases) and the [Parse Android SDK](https://github.com/parse-community/Parse-SDK-Android/releases). 

### Issue Description

ParseUser.getCurrentUser() becomes null after an update from `1.25.0` to `3.0.0`. 

### Steps to reproduce

- Log in to an app with `1.25.0`
- Update your app with parse `3.0.0`
- See that you're logout

### Actual Outcome

I'm logout

### Expected Outcome

I should be login as before

### Environment

Parse Android SDK
- SDK version: `3.0.0`
- Operating system version: `API 30`

Server
- Parse Server version: `4.10.4` (I tried in a staging environment with `5.2.0` too)
- Operating system: `Ubuntu 20.04.2 LTS`
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): `Remote host`

Database
- System (MongoDB or Postgres): `MongoDB`
- Database version: `4.4.13`
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): `MongoDB Atlas`

qmetzler-luna:
After some more testing, the issue seems to happen when updating the lib from [2.1.0](https://github.com/parse-community/Parse-SDK-Android/releases/tag/2.1.0) to [3.0.0](https://github.com/parse-community/Parse-SDK-Android/releases/tag/3.0.0)

mtrezza:
If you look at the commits, are you able to tell which commit may cause this?

PLR2388:
I think the commit that may cause this is this [one](https://github.com/parse-community/Parse-SDK-Android/commit/6fad701158950168fcfcb8a3730ae844513e99d4) as the others before [3.0.0](https://github.com/parse-community/Parse-SDK-Android/releases/tag/3.0.0) are just renaming or adding documentation 🤔

L3K0V:
[...]

Issue #1159, opened 2022-04-16
Gradle dependencies out of date


mobilekosmos:
### New Feature / Enhancement Checklist

- [X] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-SDK-Android/security/policy).
- [X] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [X] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-Android/issues?q=is%3Aissue).

Mandy gradle dependencies are outdated, please update.

mtrezza:
@mobilekosmos Do you want to open a PR for this?

PR #1163, opened 2022-04-16
feat: update gradle dependencies

mobilekosmos:
- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/Parse-SDK-Android/issues?q=is%3Aissue).

Related issue: #1159

- [ ] Add tests
- [x] A changelog entry is created automatically using the pull request title (do not manually add a changelog entry)


L3K0V (approved):
Great job @mobilekosmos 

LGTM

mtrezza:
I just noticed that CI is failing, do we need a separate PR to fix something or is this related to this PR?

mobilekosmos:
It seems that all the mockito imports are not being resolved, could also see locally in my IDE, don't know why yet.

mobilekosmos:
How to fix the spotless failing? Wouldn't it be possible to let the CI do the formatting automatically?

mtrezza:
You should be able to run spotless to auto-correct what's possible locally and commit these changes.

mobilekosmos:
"First-time contributors need a maintainer to approve running workflows."

L3K0V (approved):

mtrezza:
Apologies for the delay, I started the CI.

mtrezza:
@mobilekosmos Could you please take a look at the failing CI?

mobilekosmos:
Despite such minor changes, the pain of a PR is big.

L3K0V (approved):

mtrezza (commented):
This is upgrading the Facebook SDK from 12 to 13. Does this imply a breaking change for developers? If so we'd need to make this a major version increase.

mobilekosmos:
> Does this imply a breaking change for developers? 
Shouldn't automated tests respond to this?

mobilekosmos:
Btw. it seems that Facebook's Android SDK 13 urges to "migrate" to it: 
[](https://github.com/facebook/facebook-android-sdk/blob/main/CHANGELOG.md)
[](https://developers.facebook.com/blog/post/2022/02/24/introducing-facebook-platform-sdk-13-related-updates/?locale=en_US)

mobilekosmos:
After seeing the amount of files changed with this PR I'm asking myself if it wouldn't be better, at least in the future, to accepto only one dependency update per each PR, more burocratic and painfull again, but maybe a better approach? Not sure yet.

mtrezza:
@mobilekosmos Yes, I would not mix a major version upgrade of the FB SDK with other dependency upgrades, because it's an essential part of the SDK with its own complexities. Depending on what the breaking change in the FB SDK 13 is, we may have to declare a breaking change for the SDK if developers cannot simply upgrade the SDK without having to change their code.

mobilekosmos:
The thing is if I understood it correctly apps using an old version of the facebook sdk will stop working anyways

mtrezza:
As I understand it, only FB Login will not work, but could an app use other functionality of the FB SDK?

mobilekosmos:
It would be a topic for another discussion, but the library should be modularized anyways, the main gradle Parse-SDK shouldn't contain ALL the modules it currently has, like google play services, etc. if you need facebook, twitter, etc. you must add parse-twitter separately and so on.
Regarding this PR, do what you want guys, I just wanted to help, but if it's so difficult I cannot invest more time.

mtrezza:
We need to know whether this needs to be merged as a breaking change, that's all as far as I understand.

mtrezza:
@L3K0V Could we get your opinion on whether upgrading the FB SDK by a major version should be considered a breaking change and we need to increment the major version of the Parse Android SDK as well?

When someone adds the Parse Android SDK, the FB SDK is added as a dependency well. Could they then just use any other feature of the FB SDK in their project, unrelated to the Parse SDK? If so, upgrading the Parse SDK may break their other FB related code?

mtrezza:
Thanks, so we have to do a major version increment when we merge this.

mtrezza:
I'm suggesting to fix https://github.com/parse-community/Parse-SDK-Android/issues/1158 first before we merge this, because any release without the migration logic is for most people likely a useless release.

mtrezza:
Good news, #1158 has been merged, so we can resume this PR.

mtrezza:
In my understanding, since the FB SDK is a transitive dependency, it should be "invisible" to the developer's app. So if a developer wants to directly call any FB SDK methods, they'd have to separately add the FB SDK as dependency to their app. Moreover, that dependency would be independent from the transitive dependency, in regard to version, compatibility, etc.

Does that sound right? If so, this PR will unlikely be a breaking change and we can merge.

L3K0V:
> `compile` / `api` are long deprecated in favor of `implementation`

That's not correct, only `compile` is deprecated, `api` is used for a transitive dependencies. 

Facebook SDK dependency can be changed from `api` to `implementation` and be used only within the library. In this scenario the developer should take care of adding its own version of the Facebook SDK and use it accordingly to its source code. I'm not aware how the Facebook SDK is configured to work and what will happen if there are two versions - one used by the app and another used by the Parse SDK and if the dependency resolution will cause some troubles.

The point having the Facebook as a `api` dependency I think is to push the user to use the version bundled with the Parse SDK and do the necessary adjustments to their own Facebook code.

Another approach will be to make the Facebook SDK `runtime` or `compileRuntime` dependency and to expect the SDK to be there in the app, but in such case the Parse SDK should take care to cover as many as possible Facebook versions and their breaking changes inside the implementation.

I cannot take a decision here, not until I research how is done in other similar cases and situations.

mman:
> > `compile` / `api` are long deprecated in favor of `implementation`
> 
> That's not correct, only `compile` is deprecated, `api` is used for a transitive dependencies.
> 
> Facebook SDK dependency can be changed from `api` to `implementation` and be used only within the library. In this scenario the developer should take care of adding its own version of the Facebook SDK and use it accordingly to its source code. I'm not aware how the Facebook SDK is configured to work and what will happen if there are two versions - one used by the app and another used by the Parse SDK and if the dependency resolution will cause some troubles.

AFAIK It will cause troubles if two different versions of Facebook SDK will end up being used. I have experienced this myself many times. The Android Studio compiler will eventually flag this as "duplicate class" error since Java can not use two .jar files providing the same classes in isolation. 

This can work in C/Objective-C/Swift world where `static` method are isolated per `.o` file and multiple versions of the same method can this way be linked into the same executable, bit it does not work like this in java world as far as I know.

mtrezza:
> That's not correct, only compile is deprecated, api is used for a transitive dependencies.

Yes, sorry for the confusion, should have phrased that better.

> I'm not aware how the Facebook SDK is configured to work and what will happen if there are two versions - one used by the app and another used by the Parse SDK and if the dependency resolution will cause some troubles.

> AFAIK It will cause troubles if two different versions of Facebook SDK will end up being used. ... The Android Studio compiler will eventually flag this as "duplicate class" error.

There can't be 2 different FB SDK versions used at the same time. Gradle will try to decide on a single version. If it cannot, it will throw an error. If a developer implements their own FB SDK with a higher version, gradle will override the FB SDK version in the Parse SDK. We have to tell gradle that we *did not test* the Parse SDK with any version higher than 12.1.0. We can only assume that it will be compatible with `>=12.1.0 <13.0.0` according to semver. If we don't specify an upper range and only say `>=12.1.0`, we are effectively misleading gradle (and the developer).

If a developer tries to implement a FB SDK that is incompatible with Parse SDK's dependency of the FB SDK, gradle *should* throw an error, that is nothing bad but necessary.

In addition, we are only implementing the FB Login SDK, that means changes are high that a developer adds their own FB SDK to the app, because all other FB modules are missing.

L3K0V:
@mman my point was about the resolution strategy of Gradle which version will pick of many and what will be the behaviour after that. In your experience and from your perspective what should be the approach here?

a) Facebook SDK as `runtime` dependency covering many cases in the Parse SDK
b) Facebook SDK as `api` dependency and the developer to use same version in their app + adjustments on their own FB code
c) Facebook SDK as `implementation` and hoping Gradle to pick up most suitable version between SDK and the app.

Maybe is a manner of how do we want to 'communicate' to developers, do we want to push them to use bundled version, or their own or something else.

L3K0V:
I haven't used before the versions constraints @mtrezza, where you can specify a range. It sounds promising but again maybe we should discuss whenever the Parse SDK bundle the Facebook SDK, bundle + export (`api`) or expect it runtime.

mtrezza:
> my point was about the resolution strategy of Gradle which version will pick of many and what will be the behaviour after that. In your experience and from your perspective what should be the approach here?

That's answered in the [dependency resolution management](https://docs.gradle.org/current/userguide/dependency_management.html#understanding_dependency_resolution) and my previous 2 comments.

> Maybe is a manner of how do we want to 'communicate' to developers, do we want to push them to use bundled version, or their own or something else

Gradle communicates that to the developer. Even if we specify an upper range, the developer can still decide to force override the FB SDK version in the Parse SDK to resolve the resolution conflict. But it should never happen in the background without the developer knowing.  The best way to communicate that is not via a changelog but via gradle.

L3K0V:
> How about we add the range in this PR and open a new issue to discuss the larger question you posed?

Sounds great 👍

mtrezza:
@mman What do you think?

facebook/build.gradle
```diff
@@ -35,7 +35,7 @@ android {
 }
 
 dependencies {
-    api "com.facebook.android:facebook-login:12.1.0"
+    api "com.facebook.android:facebook-login:13.1.0"
```
mtrezza, line 38: Change to 
```
api "com.facebook.android:facebook-login:[13.2.0, 14.0["
```

mtrezza (changes requested):

mtrezza:
@mobilekosmos Would you take a look at the small change requested and the failing CI so we can merge this?

mtrezza:
Does anyone want to continue this PR to get is merged?

rommansabbir:
> Does anyone want to continue this PR to get it merged? It seems to need just a rebase and a trivial code change. It seems that @mobilekosmos has [abandoned](https://github.com/parse-community/Parse-SDK-Android/pull/1163#issuecomment-1107830008) it.

I will try to start again with this PR, once I'm back to my working place.

mtrezza:
Cool, @rommansabbir can I add you to the Parse Android SDK review team? No obligations, but you would get a notification once a PR is ready for review and can do a review to get it merge-ready.

rommansabbir:
> Cool, @rommansabbir can I add you to the Parse Android SDK review team? No obligations, but you would get a notification as member of `@parse-community/android-sdk-review` once a PR is ready for review and can do a review to get it merge-ready.

Yes, that would be amazing!

rommansabbir:
@mtrezza I have updated FB SDK (api) version to a specific range and run all the tests from the existing tests, all passed but I'm unable to push my changes to @mobilekosmos HEAD:master to that the changes could refelect in this PR.

@mobilekosmos can you check, if it's allowed to push commits to this PR?
image

Or, else I just create a new PR and refer to this one? @mman @mtrezza

mtrezza:
New PR please

mtrezza:
Closing, replaced by https://github.com/parse-community/Parse-SDK-Android/pull/1172.

PR #1168, opened 2022-05-15
fix: users logged out after SDK upgrade due to different cache path

rommansabbir:
ParsePlugins ::
- `getCacheDir()`, `getFilesDir()` APIs now has support for backward compatibility before migration to v3.0.0

### DIFF
--- a/parse/src/main/java/com/parse/ParsePlugins.java
+++ b/parse/src/main/java/com/parse/ParsePlugins.java
@@ -10,12 +10,19 @@ package com.parse;
 
 import android.content.Context;
 import android.os.Build;
+import android.util.Log;
+
 import java.io.File;
+import java.io.IOException;
+import java.util.ArrayList;
+
 import okhttp3.Headers;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
 
-/** Public for LiveQuery. You probably don't need access */
+/**
+ * Public for LiveQuery. You probably don't need access
+ */
 public class ParsePlugins {
 
     private static final String INSTALLATION_ID_LOCATION = "installationId";
@@ -111,50 +118,50 @@ public class ParsePlugins {
                 }
                 // add it as the first interceptor
                 clientBuilder
-                        .interceptors()
-                        .add(
-                                0,
-                                chain -> {
-                                    Request request = chain.request();
-                                    Headers.Builder headersBuilder =
-                                            request.headers()
-                                                    .newBuilder()
-                                                    .set(
-                                                            ParseRESTCommand.HEADER_APPLICATION_ID,
-                                                            configuration.applicationId)
-                                                    .set(
-                                                            ParseRESTCommand
-                                                                    .HEADER_APP_BUILD_VERSION,
-                                                            String.valueOf(
-                                                                    ManifestInfo.getVersionCode()))
-                                                    .set(
-                                                            ParseRESTCommand
-                                                                    .HEADER_APP_DISPLAY_VERSION,
-                                                            ManifestInfo.getVersionName())
-                                                    .set(
-                                                            ParseRESTCommand.HEADER_OS_VERSION,
-                                                            Build.VERSION.RELEASE)
-                                                    .set(ParseRESTCommand.USER_AGENT, userAgent());
-                                    if (request.header(ParseRESTCommand.HEADER_INSTALLATION_ID)
-                                            == null) {
-                                        // We can do this synchronously since the caller is already
-                                        // on a background thread
-                                        headersBuilder.set(
-                                                ParseRESTCommand.HEADER_INSTALLATION_ID,
-                                                installationId().get());
-                                    }
-                                    // client key can be null with self-hosted Parse Server
-                                    if (configuration.clientKey != null) {
-                                        headersBuilder.set(
-                                                ParseRESTCommand.HEADER_CLIENT_KEY,
-                                                configuration.clientKey);
-                                    }
-                                    request =
-                                            request.newBuilder()
-                                                    .headers(headersBuilder.build())
-                                                    .build();
-                                    return chain.proceed(request);
-                                });
+                    .interceptors()
+                    .add(
+                        0,
+                        chain -> {
+                            Request request = chain.request();
+                            Headers.Builder headersBuilder =
+                                request.headers()
+                                    .newBuilder()
+                                    .set(
+                                        ParseRESTCommand.HEADER_APPLICATION_ID,
+                                        configuration.applicationId)
+                                    .set(
+                                        ParseRESTCommand
+                                            .HEADER_APP_BUILD_VERSION,
+                                        String.valueOf(
+                                            ManifestInfo.getVersionCode()))
+                                    .set(
+                                        ParseRESTCommand
+                                            .HEADER_APP_DISPLAY_VERSION,
+                                        ManifestInfo.getVersionName())
+                                    .set(
+                                        ParseRESTCommand.HEADER_OS_VERSION,
+                                        Build.VERSION.RELEASE)
+                                    .set(ParseRESTCommand.USER_AGENT, userAgent());
+                            if (request.header(ParseRESTCommand.HEADER_INSTALLATION_ID)
+                                == null) {
+                                // We can do this synchronously since the caller is already
+                                // on a background thread
+                                headersBuilder.set(
+                                    ParseRESTCommand.HEADER_INSTALLATION_ID,
+                                    installationId().get());
+                            }
+                            // client key can be null with self-hosted Parse Server
+                            if (configuration.clientKey != null) {
+                                headersBuilder.set(
+                                    ParseRESTCommand.HEADER_CLIENT_KEY,
+                                    configuration.clientKey);
+                            }
+                            request =
+                                request.newBuilder()
+                                    .headers(headersBuilder.build())
+                                    .build();
+                            return chain.proceed(request);
+                        });
                 restClient = ParseHttpClient.createClient(clientBuilder);
             }
             return restClient;
@@ -169,7 +176,7 @@ public class ParsePlugins {
         synchronized (lock) {
             if (installationId == null) {
                 installationId =
-                        new InstallationId(new File(getFilesDir(), INSTALLATION_ID_LOCATION));
+                    new InstallationId(new File(getFilesDir(), INSTALLATION_ID_LOCATION));
             }
             return installationId;
         }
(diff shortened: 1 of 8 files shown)


parse/src/main/java/com/parse/ParsePlugins.java
```diff
@@ -178,7 +178,17 @@ InstallationId installationId() {
...
             if (cacheDir == null) {
-                cacheDir = new File(applicationContext.getCacheDir(), "com.parse");
+                /*
+                Check for old reference file before migration to v3.0.0.
+                If old reference found, copy the ref to new directory and delete the ref.
+                 */
+                File oldRef = applicationContext.getDir("Parse", Context.MODE_PRIVATE);
+                if (oldRef.exists()) {
+                    cacheDir = new File(oldRef, "com.parse");
+                    oldRef.deleteOnExit();
```
L3K0V, line 188: I remember that because of the Android's JVM specifics and the application lifecycle it's not good to use `deleteOnExit()`. This will properly work on desktop JVM but on Android it's questionable. 

- https://groups.google.com/g/android-developers/c/2FOEz35k_Uc
- https://developer.android.com/reference/java/io/File#deleteOnExit()

> Note that on Android, the application lifecycle does not include VM termination, so calling this method will not ensure that files are deleted. Instead, you should use the most appropriate out of:

>  - Use a finally clause to manually invoke [delete()](https://developer.android.com/reference/java/io/File#delete()).
> - Maintain your own set of files to delete, and process it at an appropriate point in your application's lifecycle.
> - Use the Unix trick of deleting the file as soon as all readers and writers have opened it. No new readers/writers will be able to access the file, but all existing ones will still have access until the last one closes the file.

parse/src/main/java/com/parse/ParsePlugins.java
```diff
@@ -178,7 +178,17 @@ InstallationId installationId() {
     File getCacheDir() {
         synchronized (lock) {
             if (cacheDir == null) {
-                cacheDir = new File(applicationContext.getCacheDir(), "com.parse");
+                /*
+                Check for old reference file before migration to v3.0.0.
+                If old reference found, copy the ref to new directory and delete the ref.
```
L3K0V, line 183: I can see that you copy reference, but what's the idea behind @rommansabbir? Coping reference will no do the migration itself, right?

Where the actual data copy is done from the old to the new path aka. the migration?

L3K0V (changes requested):

L3K0V:
Here you can see an migration logic about the push notifications, I have expected to see something similar but much more simple: https://github.com/parse-community/Parse-SDK-Android/commit/7d0faa38bf3ccc85b715dc61965d5b82103c19a4#diff-6cb14853da236c30db3f14e4005ea40c5c45cb8fcb9fc49541693186620de1dcR60

parse/src/main/java/com/parse/ParsePlugins.java
```diff
@@ -187,7 +197,17 @@ File getCacheDir() {
     File getFilesDir() {
         synchronized (lock) {
             if (filesDir == null) {
-                filesDir = new File(applicationContext.getFilesDir(), "com.parse");
+                /*
```
mtrezza, line 200: Could you please deduplicate this logic into a separate method?

mtrezza (changes requested):

rommansabbir:
@mtrezza @mman @L3K0V First of all, sorry for messed up with the PR. This is my first contribution to a such big community of Parse.

About my logic for cache path migrations. Based on @mman migration thoughts (https://github.com/parse-community/Parse-SDK-Android/issues/1158#issuecomment-1110097992)

I came up with this ::

- First, looking for the all files (including sub directory) in the old "app_Parse" directory
- If files exists, moving the all files to the directory "com.parse" by following the old sub directory name. After that delete the old file (already copied) silently.
- For each file, migration will be failed if there is already a file exist in the new directory with the same file name or other known cases.

If the solution seems suitable for migration process, I will create a new PR (which will include changes in `ParsePlugins.java` and `ParseFileUtils.java`).

Let me know, Thanks.

mman:
@rommansabbir Yes, this is the approach I would take, but please note that since 3.0.0 is already in the wild, and since it forced people to re-login, there will be cases where there will be fresh files in new location `com.parse` as well as files in old location, the `app_Parse` directory. In that case, the new files should be used, and old just deleted...

mman:
@rommansabbir That is exactly what I have not yet had a time to investigate properly and in detail (read: This PR is actually trickier then it may appear on a first look).

rommansabbir:
@mman I think we can find out from where the old API is accessed and which `FILE_NAME` (File name) was used, eg. `File(ParsePlugins.get().getParseDir(), FILENAME_CURRENT_CONFIG)`. We can look for others `Constants` like `FILENAME_CURRENT_CONFIG`. So, when migration is ongoing, we can check for item name matches with our `Constants` list or not, if match - then move item to `*/files/com.parse/` else `*/cache/com.parse/`

mtrezza:
> So, when running migration, for a specific file, in which location I should point to move the spcific file, */cache/com.parse or */files/com.parse?.

The two storages have a different purpose and are managed differently. That determines which file should go where.
- The cache dir is authoritatively managed by the OS and only co-managed by the app. It must not be considered persistent, in other words, the data the app stores there can be deleted by the OS at any time. That also means that if the app doesn't properly delete files the OS will make sure the cache doesn't swell up too much.
- The files dir is exclusively managed by the app. It is persistent and *must* be managed by the app. If the app doesn't delete files there, the files will stay and the app's footprint will swell up until the OS will report a storage issue to the user.

rommansabbir:
````
ParsePlugins::installationId(){File(getParseDir(), INSTALLATION_ID_LOCATION)}
ParseCorePlugins::getCurrentUserController() {File(Parse.getParseDir(), FILENAME_CURRENT_USER)}
ParseCorePlugins::getConfigController() {File(ParsePlugins.get().getParseDir(), FILENAME_CURRENT_CONFIG)}
ParseCorePlugins::getCurrentInstallationController() {File(ParsePlugins.get().getParseDir(), FILENAME_CURRENT_INSTALLATION)}
ParseCommandCache::getCacheDir(){File(Parse.getParseDir(), "CommandCache")}
LocalIdManager(Parse.getParseDir()){File(root, "LocalId")}
PushRouter::getInstance(){File(ParsePlugins.get().getParseDir(), LEGACY_STATE_LOCATION)}
````
from where `ParsePlugins.get().getParseDir()` API has been accessed. All of those files managed by SDK itself, so when migration is ongoing, we can look for those listed files and move files to `*/files/com.parse/` directory, then delete the old files from `app_Parse` by covering known cases.

After that, all other files that left in `app_Parse` directory should move to `*/cache/com.parse/` directory by covering known cases.


> * The files dir is exclusively managed by the app. It is persistent and _must_ be managed by the app. If the app doesn't delete outdated files there the files will stay and the app's footprint will swell up until the OS will report a storage issue to the user.

Version: 2.0.6
````
ParseCorePlugins::getFileController(){
             ParseFileController(
                    ParsePlugins.get().restClient(), 
                    Parse.getParseCacheDir("files")
            ))
}
````
calling `Parse.getParseCacheDir("files")` API which point to `*/cache/com.parse/`. In this case, all files (eg. when storing an `ParseObject` to `LocalDataStore` that contain an _File_ itself call `something.mp3`) that are stored in the cache directory should point to `*/files/com.parse/` as  `something.mp3` is managed by SDK itself.

**Comment by mtrezza** [comment]
> All of those files managed by SDK itself, so when migration is ongoing, we can look for those listed files and move files to */files/com.parse/ directory, then delete the old files from app_Parse by covering known cases.

Sounds good

> After that, all other files that left in app_Parse directory should move to */cache/com.parse/ directory by covering known cases.

What are these other files?

How about the other modules like facebook, fcm, it is possible that they store anything in these dirs?

**Comment by rommansabbir** [comment]
@mtrezza I found those usages in the mentioned list. If any other module deal with _File_, at the end they all point to the same API `ParsePlugins.get().getParseDir()` to retreive `File` object.

> What are these other files?

I'm not sure about how many files (`File`) is accessed (CRUD) to old dir (`app_Parse`) by the SDK itself. So,
- Files that matches with our lists, will be moved to `*/files/com.parse/` but what about others files in the old dir? (If exists)
- We can assume those files should be moved to `*/cache/com.parse/` dir.

**Comment by mtrezza** [comment]
I wonder what these other files could be? I assume it's only the Parse SDK itself that writes to that dir, so any write logic, even for temporary files should be identifiable in code. Have you seen that there are other files, or is that just an assumption?

**Comment by mtrezza** [comment]
OK, so I guess let's go with your suggested migration concept and then just test it out.

**Comment by rommansabbir** [comment]
> You mean pre 3.x or even with the latest version of the SDK?

pre 3.x since, from 3.x API removed from codebase and using `*/files/com.parse/` and `*/cache/com.parse/`.

Also, Can I write the test in Kotlin?

**Comment by mtrezza** [comment]
Not sure about Kotlin, we probably want to keep all tests in a consistent language, what do you think @L3K0V?

**Comment by L3K0V** [comment]
Tests should be in Java for now.

**Comment by mtrezza** [comment]
@rommansabbir you get a 403 error pushing to the branch in your own fork?

**Comment by rommansabbir** [comment]
> you get a 403 error pushing to the branch in your own fork? If so, you need to authenticate yourself for GitHub to accept a push.

I think, Git was trying push to the remote branch as well as my forked branch. But yeah, code is psuhed to my fork's branch.

I have added test for the migration process. Still, one more thing left, calling new overloaded APIs `ParsePlugins.get().getCacheDir(true)` and `ParsePlugins.get().getFilesDir(true)` should be used instead of old APIs in the respective classes.

**Comment by rommansabbir** [comment]
Thanks, I will definitely ask you for further guidance. Can review the changes? So that, we can close this ASAP.

**Comment by mtrezza** [comment]
> it will be called everytimes, when cache or files API is called

I would call the migration routine only once on SDK initialization.

File: parse/src/main/java/com/parse/ParsePlugins.java [hunk 4]
```diff
@@ -169,27 +176,170 @@ InstallationId installationId() {
...
             }
-            return createFileDir(filesDir);
+            return _getFilesDir();
         }
     }
+
+    /**
+     * Start old data migrations to new respective locations ({@link #getFilesDir()}, {@link #getCacheDir()}).
+     */
+    public void runSilentMigration() {
```
L3K0V, line 244: Is there any reason this to be public?

parse/src/main/java/com/parse/ParsePlugins.java
```diff
@@ -169,27 +176,170 @@ InstallationId installationId() {
...
+        ArrayList<File> filesToBeMigrated = new ArrayList<>();
+        ParseFileUtils.getAllNestedFiles(
+            getOldParseDir().getAbsolutePath(),
+            filesToBeMigrated
+        );
+        runParseDirFilesMigrationToCacheDir(filesToBeMigrated);
+    }
+
+    public static String replaceLast(String text, String regex, String replacement) {
+        return text.replaceFirst("(?s)" + regex + "(?!.*?" + regex + ")", replacement);
```
L3K0V, line 254: Why is public and why is called replace last, but we have text replace first?

parse/src/main/java/com/parse/ParsePlugins.java
```diff
@@ -169,27 +176,170 @@ InstallationId installationId() {
...
+        }
+        boolean useFilesDir = false;
+        //Hard coded config file names list.
+        ArrayList<String> configFilesToMigrateToFilesDir = new ArrayList<>();
+        configFilesToMigrateToFilesDir.add("installationId");
+        configFilesToMigrateToFilesDir.add("currentUser");
+        configFilesToMigrateToFilesDir.add("currentConfig");
+        configFilesToMigrateToFilesDir.add("currentInstallation");
+        configFilesToMigrateToFilesDir.add("LocalId");
+        configFilesToMigrateToFilesDir.add("pushState");
```
L3K0V, line 281: For readability this can be converted to something like:

```java
String[] strArray = {"installationId", "currentUser", "currentConfig", "currentInstallation", ...};
List<String> configFilesToMigrateToFilesDir = Arrays.asList(strArray);
```

parse/src/main/java/com/parse/ParsePlugins.java
```diff
@@ -169,27 +176,170 @@ InstallationId installationId() {
...
+                //Parse the old sub directory name where the file should be moved (new location) by following the old sub directory name.
+                String oldSubDir = replaceLast(itemToMove.getAbsolutePath().replace(getOldParseDir().getAbsolutePath(), "").replace("/" + itemToMove.getName(), ""), itemToMove.getName(), "");
+                File fileToSave = new File(
+                    (useFilesDir ? applicationContext.getFilesDir() : applicationContext.getCacheDir())
+                        + "/com.parse/" +
+                        oldSubDir,
+                    itemToMove.getName());
+                //Perform copy operation if file doesn't exist in the new directory.
+                if (!fileToSave.exists()) {
+                    ParseFileUtils.copyFile(itemToMove, fileToSave);
```
L3K0V, line 302: Can we add log somewhere here what was migrated? I believe it will be interesting in a verbose mode for the developers.

parse/src/main/java/com/parse/ParsePlugins.java
rommansabbir: No, I don't see any valid to be public. Also, Migration process is now independent module and protected, so no more public access.

parse/src/main/java/com/parse/ParsePlugins.java
rommansabbir: Code refactored and reason is `Before returning the path, replace file name from the last, eg. dir name & file name could be same, as we want to get only dir name.`

parse/src/main/java/com/parse/ParsePlugins.java
rommansabbir: Thanks. Refactored.

parse/src/main/java/com/parse/ParsePlugins.java
rommansabbir: Yes, Added (`PLog.v()`) verbos log for events like _Copying_, _Already Exist_, _Deletion_, _Migration_Completion_.

rommansabbir (commented):
fix: parse cache dir migration routine on SDK initialization (only once) (#1168)

L3K0V (approved):
LGTM, but it would be great if someone else take a look as well.

rommansabbir (commented):
Fixed SpotlessJavaCheck Violations gradle build issue for PR (#1168)

mtrezza:
> LGTM, but it would be great if someone else take a look as well.

Would be nice to have a pre-release branch to test things out easier, to consider for the future.

@PLR2388 Since you originally reported the issue, do you think you could test out this fix to see whether the migration happens successfully?

rommansabbir:
@mman @mtrezza Any update on this PR?

mtrezza:
I would like to have someone test this out with an actual app before we merge it, just to validate that it works.

rommansabbir:
@mtrezza I will review the codes and will update you later.

mman:
@rommansabbir I will try to test until the end of this week...

rommansabbir:
@mtrezza According to the code comments:

- _Make sure the data on disk for Parse is for the current application_ on Parse SDK Init.
- _Verifies that the data stored on disk for Parse was generated using the same application that is running now_
- _Make sure the current version of the cache is for this application id else the applicationId file was malformed or something_
- _The application id has changed, so everything on disk is invalid._

mtrezza:
Thanks for investigating, so the `applicationId` there is essentially to validate that the cache belongs to the app. That makes sense I guess. So if an app would allow to switch between different application IDs, the cache would get invalidated.

mtrezza:
I can confirm that the migration works fine.

It moves from cache to files:
- currentConfig
- currentInstallation
- currentUser

It does not move from cache to files:
- installationId

Everything as expected, installation update and user requests also work fine.

mtrezza:
Changelog entry:
> ### Fix
> - users logged out after SDK upgrade due to different cache path; this fixes the bug that was introduced with release 3.0.0 which ignores SDK-internal data that is stored locally on the client side (#1168)

mtrezza:
@mman If you want to test that would be great, then we'll wait until the end of the week with the release.

mtrezza (approved):

rommansabbir:
> Great job, @rommansabbir! Would you be interested in joining the Parse Android SDK team? No commitment in terms of contribution, but it would be great if you could at times help out with things like PR reviews for example.

Yes, I would love to 😃

mman:
@rommansabbir @mtrezza I'd love to double check, the code looks good. What's the easiest way to build my app against this branch. I have tried naive approach to depend directly on git branch, but that apparently does not work. What steps are needed? Sorry for my lame question.

rommansabbir:
image

Add this branch dependency:

`implementation 'com.github.rommansabbir:Parse-SDK-Android:patch-1-SNAPSHOT'`

Now, you can install the latest SDK from this branch.

mtrezza:
You can also close the repo, checkout the PR branch, build to local maven and reference the builds.

mman:
Thanks @rommansabbir, I was missing the `-SNAPSHOT` in the github reference.

```
implementation 'com.github.rommansabbir:Parse-SDK-Android:patch-1-SNAPSHOT'
```

I have now tried our apps and looks like this PR does correctly move all files around.

Thanks, LGTM 👍

parseplatformorg:
🎉 This change has been released in version [3.0.1](https://github.com/parse-community/Parse-SDK-Android/releases/tag/3.0.1)

Issue #1187, opened 2023-02-20
Facebook Login not working when Facebook app is not installed


cjosepha:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-SDK-Android/security/policy).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-Android/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the latest version of [Parse Server](https://github.com/parse-community/parse-server/releases) and the [Parse Android SDK](https://github.com/parse-community/Parse-SDK-Android/releases). 

### Issue Description

When the Facebook app is not installed on the device and the app uses AndroidX fragments 1.5+, the Facebook Login flow is broken and the app login Activity is being recreated on top of the existing login Activity instance. Additionally, after closing the additional Activity, the FacebookCallback.onCancel() method is called regardless the webview login and permission was successful. This makes Parse SDK returning a successful task, but with still no ParseUser logged.

### Steps to reproduce

- Make sure Facebook app is not installed on the device
- Make sure your app uses AndroidX 1.5+ for fragment/activities. In my case: `implementation 'androidx.appcompat:appcompat:1.5.1'`
- Use Parse SDK 4.1.0 with Facebook:
  - `implementation "com.github.parse-community.Parse-SDK-Android:facebook:4.1.0"`
  - `implementation "com.github.parse-community.Parse-SDK-Android:parse:4.1.0"`
- Implement Facebook Login
- Launch the app and launch Facebook Login by calling `ParseFacebookUtils.logInWithReadPermissionsInBackground(...)`
- Enter valid credentials
- Validate

### Actual Outcome

- Additional login Activity created
- `ParseUser.getCurrentUser()` still returns null
- `onActivityResult()` is not called
- `ParseFacebookUtils.logInWithReadPermissionsInBackground(...)` does not return any task and callback is not called

### Expected Outcome
- `onActivityResult()` is called
- `ParseFacebookUtils.logInWithReadPermissionsInBackground(...)` returns a task with a ParseUser or callback success method is called

### Environment

Parse Android SDK
- SDK version: `4.1.0`
- Operating system version: `Android 13`

Server
- Parse Server version: `4.2.0`
- Operating system: `Linux/4.17.6`
[...]

PR #1188, opened 2023-02-20
feat: Add support for Facebook SDK 15.x

cjosepha:

- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/Parse-SDK-Android/issues/1187).

### Issue Description

Use more recent version of Facebook SDK to fix an issue where the login flow is broken if the Facebook app is not installed on the device. See Facebook SDK v14.1.1 changelog: https://github.com/facebook/facebook-android-sdk/blob/main/CHANGELOG.md#1411

Closes: #1187 

### Approach

Allow to use Facebook SDK dependency v14.1.1

### TODOs before merging

Nothing necessary

### DIFF
--- a/facebook/build.gradle
+++ b/facebook/build.gradle
@@ -35,7 +35,7 @@ android {
 }
 
 dependencies {
-    api "com.facebook.android:facebook-login:[13.2.0, 14.0["
+    api "com.facebook.android:facebook-login:[13.2.0, 14.1.1]"
     implementation project(":parse")
 
     testImplementation "junit:junit:$rootProject.ext.junitVersion"
--- a/facebook/build.gradle
+++ b/facebook/build.gradle
@@ -35,7 +35,7 @@ android {
 }
 
 dependencies {
-    api "com.facebook.android:facebook-login:[13.2.0, 16.0.0]"
+    api "com.facebook.android:facebook-login:[13.2.0, 17.0.0["
     implementation project(":parse")
 
     testImplementation "junit:junit:$rootProject.ext.junitVersion"
--- a/facebook/build.gradle
+++ b/facebook/build.gradle
@@ -35,7 +35,7 @@ android {
 }
 
 dependencies {
-    api "com.facebook.android:facebook-login:[13.2.0, 14.1.1]"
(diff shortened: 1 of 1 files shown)


facebook/build.gradle
```diff
@@ -35,7 +35,7 @@ android {
 }
 
 dependencies {
-    api "com.facebook.android:facebook-login:[13.2.0, 14.0["
+    api "com.facebook.android:facebook-login:[13.2.0, 14.1.1]"
```
mtrezza, line 38: The upper limit is set because the SDK has not been tested for compatibility with higher versions. If we know it's compatible with Facebook SDK 14.x we can set the upper range to <=15. If we know it's compatible with Facebook SDK 15 and 16 we can even set it to <=17.

mtrezza (changes requested):

facebook/build.gradle
cjosepha: I'm able to test the Facebook Login feature only, is that sufficient to say it's compatible? BTW last published version is 16.0.0 https://github.com/facebook/facebook-android-sdk/commits/main

I can try with this and let you know:
```suggestion
    api "com.facebook.android:facebook-login:[13.2.0, 16.0.0]"
```

cjosepha: @mtrezza tested Facebook Login with 16.0.0 and it's working.

mtrezza: Great; coupd you also test with version 15? Then we can set the upper bount to <17.

cjosepha: @mtrezza I already set the upper bound to <17
Version 14 was already tested OK as this PR was initially using this one
Version 15 has just been tested OK

mtrezza: Great, so I think you could change this to 
`api "com.facebook.android:facebook-login:[13.2.0, 17.0.0["`, which means SDK is supported up to excluding version 17.0.0

cjosepha: @mtrezza please check the commits, I've done this yesterday

mtrezza (approved):
Looks good!

parseplatformorg:
🎉 This change has been released in version [4.2.0](https://github.com/parse-community/Parse-SDK-Android/releases/tag/4.2.0)

Issue #764, opened 2017-12-07
ParseObject saved multiple times on slow internet connection


Kiwifed0r:
Hi,

I'm having an issue with parse and very slow network connection. This can be reproduced with the emulator when using cellular network and setting it to GSM and signal strength 'None'. Now, when saving a ParseObject, the call goes through and the ParseObject shows up in the dashboard but the client doesn't get the answer before the timeout. This means that Parse tries again and saves it another time and so on until the maximum number of retries is reached. And finally the callback is called reporting an error. Setting the network to LTE and signal strength 'Goog' at some time results in a positive callback for the most recent try.

One solution which I tried is to set the timeouts to 60 seconds. This works and the answer from the server reaches the client on very slow internet before the timeout.
Another possible solution would be to tell Parse not to retry.

Does anyone have any other suggestions how to solve this issue?
Does it make sense for Parse to retry saving an object after a failure? There's no way to be sure whether or not the object was saved.
For fetching data on the other hand retrying is fine.

This only happens when using https.

Parse version: 1.16.3

montymxb:
We'll have to see about this one, it's a bit tough when working under very poor or interrupted network conditions to ensure much of anything. I wonder if maybe preventing retries would even help, as you still may not be able to tell if the object was even created or not. It would be nice to deliberate and see if there is a reasonable solution for this.

Kiwifed0r:
I ran some tests on iOS and I was able to reproduce the issue on iOS aswell. The only difference is that the timeouts are different. It seems that the default timeouts on iOS are a lot higher (60 seconds) compared to Android (okhttp, 10 seconds?). To reduce the chance of this happening I will set the timeouts to 60 seconds for now. The connection has to be incredibly poor for this to happen with 60 seconds timeout and it shouldn't happen very often but still not impossible.

I played around with setting number of retries to 0 and it also prevents saving the object multiple times but then the problem is that we still don't really know what's going on and the 10 seconds timeout is too less.

The best 'solution' for now is to increase the okhttp timeout.

montymxb:
@Kiwifed0r hopefully you're able to keep the issue at bay with your timeout patch. It's not the best but if it works in your case it will have to do for the moment. Since, this could happen to _any_ sdk I'm wondering if a server side fix should be considered...

[...]

PR #1190, opened 2023-03-16
feat: Add idempotency feature to detect duplicate requests due to network conditions

DrMimik:

- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/Parse-SDK-Android/issues/764).

### Issue Description
Requests are duplicated on slow internet connections and during network interruptions.

Closes: [#764](https://github.com/parse-community/Parse-SDK-Android/issues/764)

### Approach
Add idempotency requestId header for client requests to enforce server side idempotency middleware on selected routes..

### TODOs before merging

- [x] Add tests

### DIFF
--- a/parse/src/main/java/com/parse/ParseRESTCommand.java
+++ b/parse/src/main/java/com/parse/ParseRESTCommand.java
@@ -20,6 +20,8 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.Map;
+import java.util.UUID;
+
 import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
@@ -37,8 +39,8 @@ class ParseRESTCommand extends ParseRequest<JSONObject> {
     /* package */ static final String HEADER_INSTALLATION_ID = "X-Parse-Installation-Id";
     /* package */ static final String HEADER_REQUEST_ID = "X-Parse-Request-Id";
     /* package */ static final String USER_AGENT = "User-Agent";
-    /* package */ static final String HEADER_SESSION_TOKEN = "X-Parse-Session-Token";
-    /* package */ static final String HEADER_MASTER_KEY = "X-Parse-Master-Key";
+    private static final String HEADER_SESSION_TOKEN = "X-Parse-Session-Token";
+    private static final String HEADER_MASTER_KEY = "X-Parse-Master-Key";
     private static final String PARAMETER_METHOD_OVERRIDE = "_method";
 
     // Set via Parse.initialize(Configuration)
@@ -50,6 +52,7 @@ class ParseRESTCommand extends ParseRequest<JSONObject> {
     /* package */ String httpPath;
     private String installationId;
     private String operationSetUUID;
+    private final String requestId = UUID.randomUUID().toString();
     private String localId;
 
     public ParseRESTCommand(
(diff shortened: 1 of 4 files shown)


parse/src/main/java/com/parse/ParseRESTCommand.java
```diff
@@ -215,6 +216,20 @@ protected void addAdditionalHeaders(ParseHttpRequest.Builder requestBuilder) {
...
+            // using header names so we don't override a parameter with the same key name,
+            // using all headers to insure the requestId generated doesn't conflict with the rest of the users
+            if (installationId != null)
+                jsonObject.put(HEADER_INSTALLATION_ID, installationId);
+            if (sessionToken != null)
+                jsonObject.put(HEADER_SESSION_TOKEN, sessionToken);
+            if (masterKey != null)
+                jsonObject.put(HEADER_MASTER_KEY, masterKey);
+            requestBuilder.addHeader(HEADER_REQUEST_ID, ParseDigestUtils.md5(toDeterministicString(jsonObject)));
+        } catch (JSONException e) {
```
mtrezza, line 230: The OS should provide a UUID API that takes care of seeding. Could you use the same logic for generating the UUID as we use in the other SDKs? For example the iOS SDK and JS SDK?

parse/src/main/java/com/parse/ParseRESTCommand.java
```diff
@@ -35,9 +35,10 @@ class ParseRESTCommand extends ParseRequest<JSONObject> {
     /* package */ static final String HEADER_OS_VERSION = "X-Parse-OS-Version";
 
     /* package */ static final String HEADER_INSTALLATION_ID = "X-Parse-Installation-Id";
+    /* package */ static final String HEADER_REQUEST_ID = "X-Parse-Request-Id";
     /* package */ static final String USER_AGENT = "User-Agent";
-    private static final String HEADER_SESSION_TOKEN = "X-Parse-Session-Token";
-    private static final String HEADER_MASTER_KEY = "X-Parse-Master-Key";
```
mtrezza, line 40: Revert, shouldn't be needed

mtrezza (changes requested):

parse/src/main/java/com/parse/ParseRESTCommand.java
DrMimik: I have rewritten the logic to use a UUID, as this is uniformed between SDKs.

But imo why isn't the UUID data based?, like this if the SDK calls ParseObject.save() it changes the ParseRequest which creates a new UUID, while using a data based UUID calling save won't change the UUID header.

Currently if I want the SDK to create an idempotent request I'd still set the HttpClient to use a longer timeout and increase the default max retries of requests.

DrMimik:
Should I force push the first commit? or use a new commit?.

parse/src/main/java/com/parse/ParseRESTCommand.java
DrMimik: I am not very good at JS, but could you direct me to where the iOS SDK impliments the logic.

mtrezza: > But imo why isn't the UUID data based?

It doesn't need to be based on object data. A UUID is a Universally **Unique** IDentifier, which ensures that every generated UUID is unique.

The part you'd have to take care about in the PR is that for a `ParseObject.save` that UUID is only generated one time and then reused if the request fails and the SDK auto-retries the request. Obviously, if you generated a new UUID with ever retry of the same request, then this wouldn't work.

DrMimik: The current approach works as follows if all requests were unsuccessfull:

`ParseObject object = new ParseObject("GameScore");`
// calls a request and retries according to the max retries (default is 4) set during initilization, with the same requestId header. (5 calls with the same header)
`object.save();`

// this second method call... calls a request and retries according to the max retries set, with the same requestId header but different than the first save() method call.
`object.save();`

mtrezza: The second `save` call should use a different UUID for the request ID. In fact, every `save` call should use its own UUID. Only auto-retries of each `save` call should use the same UUID.

DrMimik: Yes, It is using exactly that. My explanation was kind of confusing.

I'll add a second test using the code in my previous comment.

azlekov:
> Should I force push the first commit? or use a new commit?.

New commit is preferable to track changes and reviewing it easily.
Hoping to get this merged.

dplewis:
@mtrezza Is this ready to merge?

mtrezza (changes requested):
Lint failed

mtrezza:
@parse-community/android-sdk anyone wants to pick this up and submit a new PR?

abubkr-hago:
Hi, I am a new contributor, I would love to continue on this PR

mtrezza:
Sure, please go ahead.

PR #1191, opened 2023-03-17
feat: Implementing encrypted local storage for user sessions

DrMimik:

- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/Parse-SDK-Android/issues/1192).

### Issue Description
User data can be copied on rooted devices.

Closes: [#1192](https://github.com/parse-community/Parse-SDK-Android/issues/1192)

### Approach
Encrypting local user session using Jetpack security features to ensure better security for rooted devices.

### TODOs before merging

- [x] Add tests

### DIFF
--- a/parse/src/main/java/com/parse/EncryptedFileObjectStore.java
+++ b/parse/src/main/java/com/parse/EncryptedFileObjectStore.java
@@ -15,7 +15,10 @@ import java.io.IOException;
 import java.security.GeneralSecurityException;
 import java.util.concurrent.Callable;
 
-public class EncryptedFileObjectStore<T extends ParseObject> implements ParseObjectStore<T> {
+/**
+ * a file based {@link ParseObjectStore} using Jetpack's {@link EncryptedFile} class to protect files from a malicious copy.
+ */
+class EncryptedFileObjectStore<T extends ParseObject> implements ParseObjectStore<T> {
 
     private final String className;
     private final File file;
@@ -42,11 +45,27 @@ public class EncryptedFileObjectStore<T extends ParseObject> implements ParseObj
         return ParseCorePlugins.getInstance().getSubclassingController();
     }
 
+    /**
+     * Saves the {@code ParseObject} to the a file on disk as JSON in /2/ format.
+     *
+     * @param current ParseObject which needs to be saved to disk.
+     * @throws IOException              thrown if an error occurred during writing of the file
+     * @throws GeneralSecurityException thrown if there is an error with encryption keys or during the encryption of the file
+     */
     private void saveToDisk(ParseObject current) throws IOException, GeneralSecurityException {
         JSONObject json = coder.encode(current.getState(), null, PointerEncoder.get());
         ParseFileUtils.writeJSONObjectToFile(encryptedFile, json);
     }
 
+    /**
+     * Retrieves a {@code ParseObject} from a file on disk in /2/ format.
+     *
+     * @return The {@code ParseObject} that was retrieved. If the file wasn't found, or the contents
+     * of the file is an invalid {@code ParseObject}, returns {@code null}.
+     * @throws GeneralSecurityException thrown if there is an error with encryption keys or during the encryption of the file
+     * @throws JSONException            thrown if an error occurred during the decoding process of the ParseObject to a JSONObject
+     * @throws IOException              thrown if an error occurred during writing of the file
+     */
     private T getFromDisk() throws GeneralSecurityException, JSONException, IOException {
         return ParseObject.from(coder.decode(ParseObject.State.newBuilder(className), ParseFileUtils.readFileToJSONObject(encryptedFile), ParseDecoder.get()).isComplete(true).build());
     }
@@ -59,8 +78,8 @@ public class EncryptedFileObjectStore<T extends ParseObject> implements ParseObj
                 if (!file.exists()) return null;
                 try {
                     return getFromDisk();
-                } catch (GeneralSecurityException | JSONException | IOException e) {
-                    throw new RuntimeException(e);
+                } catch (GeneralSecurityException e) {
+                    throw new RuntimeException(e.getMessage());
                 }
             }
         }, ParseExecutors.io());
@@ -72,8 +91,8 @@ public class EncryptedFileObjectStore<T extends ParseObject> implements ParseObj
             if (file.exists() && !ParseFileUtils.deleteQuietly(file)) throw new RuntimeException("Unable to delete");
             try {
                 saveToDisk(object);
-            } catch (IOException | GeneralSecurityException e) {
-                throw new RuntimeException(e);
+            } catch (GeneralSecurityException e) {
+                throw new RuntimeException(e.getMessage());
             }
             return null;
         }, ParseExecutors.io());
--- a/parse/src/main/java/com/parse/EncryptedFileObjectStore.java
+++ b/parse/src/main/java/com/parse/EncryptedFileObjectStore.java
@@ -1,0 +1,94 @@
+package com.parse;
+
+import android.content.Context;
+
+import androidx.security.crypto.EncryptedFile;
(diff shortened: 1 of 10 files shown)


parse/src/main/java/com/parse/ParseObjectStoreMigrator.java
```diff
@@ -0,0 +1,71 @@
+package com.parse;
+
+import com.parse.boltsinternal.Continuation;
+import com.parse.boltsinternal.Task;
+
+import java.util.Arrays;
+
+public class ParseObjectStoreMigrator<T extends ParseObject> implements ParseObjectStore<T> {
```
azlekov, line 8: You could add some Javadoc with comments. 

Does this migrator needs to be here forever or just for about some migration period?

parse/src/main/java/com/parse/EncryptedFileObjectStore.java
```diff
@@ -0,0 +1,94 @@
...
+
+import org.json.JSONException;
+import org.json.JSONObject;
+
+import java.io.File;
+import java.io.IOException;
+import java.security.GeneralSecurityException;
+import java.util.concurrent.Callable;
+
+public class EncryptedFileObjectStore<T extends ParseObject> implements ParseObjectStore<T> {
```
azlekov, line 18: Can you add Javadoc on the public classes and their public methods you have introduced?

parse/src/main/java/com/parse/ParseObjectStoreMigrator.java
```diff
@@ -0,0 +1,71 @@
...
+
+    private ParseObjectStore<T> store;
+    private ParseObjectStore<T> legacy;
+
+    public ParseObjectStoreMigrator(ParseObjectStore<T> store, ParseObjectStore<T> legacy) {
+        this.store = store;
+        this.legacy = legacy;
+    }
+
+    private static <T extends ParseObject> Task<T> migrate(
```
azlekov, line 18: Is there a way to check whenever migrate is needed?

azlekov: If `from` is deleted and next time you call migrate the `onSuccessTask` what will happen? It will call again `onSuccessTask`? Is there chance `onFailedTask` or something similar to be called in such case?

parse/src/test/java/com/parse/AndroidKeyStoreProvider.kt
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    class RsaKeyPairGenerator : KeyPairGeneratorSpi() {
+        private val wrapped = KeyPairGenerator.getInstance("RSA", "BC")
+
+        private var lastSpec: AlgorithmParameterSpec? = null
+
+        override fun generateKeyPair(): KeyPair = wrapped.generateKeyPair().also { keyPair ->
+            null
+//            AndroidKeyStore.storedKeys[lastSpec!!.keystoreAlias] = KeyStore.PrivateKeyEntry(keyPair.private, arrayOf(keyPair.toCertificate()))
```
azlekov, line 131: Let's not have commented code. 

If this is a TODO or a FIXME or something important for reference later on, please consider adding a descriptive comment instead of deleting it.

parse/src/test/java/com/parse/EncryptedFileObjectStoreTest.java
```diff
@@ -0,0 +1,125 @@
...
+        ParseTaskUtils.wait(store.setAsync(user));
+
+        Context context = InstrumentationRegistry.getInstrumentation().getContext();
+        EncryptedFile encryptedFile = new EncryptedFile.Builder(context, file, new MasterKey.Builder(context).setKeyScheme(MasterKey.KeyScheme.AES256_GCM).build(), EncryptedFile.FileEncryptionScheme.AES256_GCM_HKDF_4KB).build();
+        JSONObject jsonAgain = ParseFileUtils.readFileToJSONObject(encryptedFile);
+        assertEquals(json, jsonAgain, JSONCompareMode.STRICT);
+    }
+
+    @Test
+    public void testGetAsync() throws Exception {
```
azlekov, line 83: Can we have a test covering the `migrate`? Pardon me if is covered already. If it is can you point it to me?

azlekov (commented):
Wow @DrMimik! Many thanks for this contribution. It's highly appreciated! Looks very clean and nice!

I have the feeling that this is something working for a long time somewhere else, correct me if I'm wrong. I'm guessing that because of the copyright comments.

I left just few questions and small javadoc suggestions. I'm wondering @mtrezza if the migration mechanism should be here forever or there is some policy about that. But if @DrMimik said already migrated stores are covered and there's no performance issue maybe can stay for a while.

parse/src/test/java/com/parse/AlgorithmParameterSpecExt.kt
```diff
@@ -0,0 +1,22 @@
+package com.parse
+
+/*
+ * Copyright 2020 Appmattus Limited
```
mtrezza, line 4: We need to make sure to consider the license implications. This repository is planned to be migrated to Apache 2, we should do that before merged this, otherwise we need to add this as a second license.

mtrezza (changes requested):

mtrezza:
We don't have a specific policy for the duration of providing migration mechanisms. It depends on the type of change. In this case I'd see the mechanism staying for several years, so indefinite at this point.

parse/src/main/java/com/parse/ParseObjectStoreMigrator.java
DrMimik: > Wow @DrMimik! Many thanks for this contribution. It's highly appreciated! Looks very clean and nice!
>
> I have the feeling that this is something working for a long time somewhere else, correct me if I'm wrong. I'm guessing that because of the copyright comments.
> 
> I left just few questions and small javadoc suggestions. I'm wondering @mtrezza if the migration mechanism should be here forever or there is some policy about that. But if @DrMimik say that already migrated stores are covered and there's no performance issue maybe can stay for a while.

Thank you guys for creating this awesome platform, I actuallty just copied the `ParseObjectStore`'s logic for creating a new ObjectStore and used Jetpack's crypto library.

Since Robolectric doesn't mock Android's `java.security.KeyStore` I copied a couple files from this [repo](https://github.com/appmattus/layercache/tree/main/testutils/src/main/kotlin/com/appmattus/layercache/keystore) (with their copyright XD ) to create tests using Robolectric.

DrMimik: > We don't have a specific policy for the duration of providing migration mechanisms. It depends on the type of change. In this case I'd see the mechanism staying for several years, so indefinite at this point. The reason is that it can be a years long process to migrate clientes once they are released to end-users.

I think the migrator should be available for new users until they migrate, or remove it on a major release.

parse/src/main/java/com/parse/ParseObjectStoreMigrator.java
DrMimik: Thanks for pointing that out, that is actually a bug which is going to be calling `from.getAsync()` recursively.

azlekov:
@mtrezza @DrMimik I believe here we are on the finish line, what do you think can we get this merged?

rommansabbir:
Hi, pardon me if my question is out of the context. Why we need to `Encrypt` the user session for Rooted Device? If a device is rooted, the device is compromised. And to decrypt the `Session Token`, someone have to decrypt the token by using decode method to get `JSON` payload. If so, then session token is already `Encrypted`, so why do we need to add an extra security layer for the `Session Token`?

mtrezza:
@parse-community/android-sdk anyone wants to pick this up and submit a new PR? It seems this was almost ready for merge.

rommansabbir:
As it was almost ready for merge, I will work on the PR very soon. Thanks for the attention.

Issue #1194, opened 2023-04-16
IllegalArgumentException when initializing Parse


hej2010:
### New Issue Checklist

- [X] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-SDK-Android/security/policy).
- [X] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [X] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-Android/issues?q=is%3Aissue).
- [X] I can reproduce the issue with the latest version of [Parse Server](https://github.com/parse-community/parse-server/releases) and the [Parse Android SDK](https://github.com/parse-community/Parse-SDK-Android/releases). 

### Issue Description

The app crashes when initializing Parse with the latest Gradle/Android Studio version when using Proguard/R8 with `minifyEnabled true` and `'com.android.tools.build:gradle:8.0.0'`.

### Steps to reproduce

Initialize Parse at app startup with `Parse.initialize(new Parse.Configuration.Builder(...));`

### Actual Outcome

The app crashes.

### Expected Outcome

The app does not crash.

### Environment

Parse Android SDK
- SDK version: `4.2.0`
- Operating system version: `Android 12`

### Logs

```
Caused by java.lang.IllegalArgumentException: Cannot register a type that does not implement the default constructor!
       at com.parse.ParseObjectSubclassingController.registerSubclass(ParseObjectSubclassingController.java:95)
       at com.parse.ParseObject.registerSubclass(ParseObject.java:291)
       at com.parse.ParseObject.registerParseSubclasses(ParseObject.java:1032)
       at com.parse.Parse.initialize(Parse.java:169)
       at com.parse.Parse.initialize(Parse.java:140)
```
I do not have any custom subclasses registered.

### Fixed by
Keeping all classes that are registered in `registerParseSubclasses()` by adding the line `-keep @com.parse.ParseClassName class com.parse.*` to `proguard-rules.pro`.

mtrezza:
In other words, gradle 8 is not supported OOTB by the Parse SDK, is that correct?

hej2010:
> In other words, gradle 8 is not supported OOTB by the Parse SDK, is that correct?

Yes (if you are using minifyEnabled).

mtrezza:
Pre gradle 8 minify would work, right?

hej2010:
Yes, everything worked until I updated to Gradle 8

[...]

PR #1196, opened 2023-05-05
fix: Missing Proguard rules for R8 in full mode

azlekov:

- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/Parse-SDK-Android/issues?q=is%3Aissue).

### Issue Description

Closes: #1194

### TODOs before merging

- [ ] Add tests

### DIFF
--- a/parse/release-proguard.pro
+++ b/parse/release-proguard.pro
@@ -1,7 +1,10 @@
 -keepnames class com.parse.** { *; }
+-keep @com.parse.ParseClassName class com.parse.*
 
 # Required for Parse
 -keepattributes *Annotation*
 -keepattributes Signature
+
 # https://github.com/square/okio#proguard
 -dontwarn okio.**
+-dontwarn org.codehaus.mojo.animal_sniffer.*
--- a/parse/release-proguard.pro
+++ b/parse/release-proguard.pro
@@ -1,5 +1,8 @@
--keepnames class com.parse.** { *; }
 -keep @com.parse.ParseClassName class com.parse.*
+-keepnames class com.parse.** { *; }
+-keepclassmembers public class * extends com.parse.** {
+   public <init>(...);
+}
 
 # Required for Parse
 -keepattributes *Annotation*
--- a/parse/release-proguard.pro
+++ b/parse/release-proguard.pro
@@ -8,6 +8,6 @@
 -keepattributes *Annotation*
 -keepattributes Signature
 
-# https://github.com/square/okio#proguard
--dontwarn okio.**
--dontwarn org.codehaus.mojo.animal_sniffer.*
+# Retracing stacktraces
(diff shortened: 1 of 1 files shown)


mtrezza:
Is this ready for merge?

mtrezza (approved):
Looks good

azlekov:
> Looks good
> 
> 
> 
> Is this ready for merge? I've changed this to a `feat` PR, since it seems to add support for Gradle 8 to the Parse SDK.

That's not quite correct. But fixes fine release issues with ProGuard and R8

mtrezza:
I assumed so because the issues don't seem to have these occurred so far. Is this an adaptation for Gradle 8 or is this also a bug for Gradle <8?

mtrezza:
Maybe we should start adding a gradle matrix to the CI where we stest with different gradle versions - would that make sense?

azlekov:
Hey @mtrezza the issues which the developers are reporting are runtime. Because of wrong ProGuard rules after magnification is done Parse classes are erased and when you run the app it crashes. I need to think about if there can be some tests covering the cases by building and running simple app using parse via some instrumentation test and check whatever it crash or not.

mtrezza:
Ok, so the minification is the core issue here. Other than that, wouod it still male sense to run the CI with multiple Gradle or Android versions? If so, I'll open a separate issue for that.

mtrezza:
@azlekov This PR got somewhat stale, can we merge this or is there anything that is worth discussing beforehand?

parse/release-proguard.pro
```diff
@@ -1,7 +1,13 @@
...
+   public <init>(...);
+}
 
 # Required for Parse
 -keepattributes *Annotation*
 -keepattributes Signature
+
 # https://github.com/square/okio#proguard
 -dontwarn okio.**
+-dontwarn org.codehaus.mojo.animal_sniffer.*
```
mtrezza, line 13: I assume this was added by mistake, or does the SDK have auch a dependency?

azlekov: Done, thanks!

azlekov:
@mtrezza I change the title to be more precise, it's not a feat, but a fix for proguard R8 in full mode as mentioned in the #1194

mtrezza:
Great! I've just slightly edited the title as it should describe the actual issue. Otherwise I think this is ready for merge?

mtrezza (approved):

azlekov:
> Great! I've just slightly edited the title as it should describe the actual issue. Otherwise I think this is ready for merge?

Yes, go go go! 🚗

parseplatformorg:
🎉 This change has been released in version [4.2.1](https://github.com/parse-community/Parse-SDK-Android/releases/tag/4.2.1)

PR #1208, opened 2024-08-15
feat: Add idempotency feature to detect duplicate requests due to network conditions

abubkr-hago:

- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/Parse-SDK-Android/issues/764).

### Issue Description
ParseObject saved multiple times on slow internet connection 

Closes: [#764](https://github.com/parse-community/Parse-SDK-Android/issues/764)

### Approach
Add `X-Parse-Request-Id` header to requests allowing parse-server to handle idempotency.

### TODOs before merging
- [x] Add tests
- [x] Documentation available here [parse-server-docs](http://parseplatform.org/parse-server/api/7.2.0/#idempotency-enforcement).


mtrezza (commented):
It seems that this always adds the req ID header to all requests. This should be optional because if the feature is not in use it just causes more traffic costs.

Issue #1217, opened 2024-08-18
Fix failing tests


mtrezza:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-SDK-Android/security/policy).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-Android/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the latest version of [Parse Server](https://github.com/parse-community/parse-server/releases) and the [Parse Android SDK](https://github.com/parse-community/Parse-SDK-Android/releases). 

### Issue Description

At some point, tests did not run anymore because they were not identified. The CI job itself passed, but it could have been noticed by a sudden decrease in test coverage of 0%. Issue was fixed in https://github.com/parse-community/Parse-SDK-Android/pull/1216, however this also means the CI is not failing due to 9 failing tests.

### Steps to reproduce

See https://github.com/parse-community/Parse-SDK-Android/actions/runs/10436358169/job/28901326229 for the failing tests, or for better overview https://github.com/parse-community/Parse-SDK-Android/pull/1216/commits/fdd94883ad0057fcd183b89bf7aea69ddb5bdac4 which shows the 9 tests.

### Actual Outcome

Tests fail, apparently due to the way the testing is conducted, not because the SDK code itself is faulty.

### Expected Outcome

Tests should pass.

mtrezza:
@abubkr-hago would you want to give this a try?

PR #1213, opened 2024-08-17
refactor: Bump AGP from 7.1.3 to 8.5.2

abubkr-hago:
### New Pull Request Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-SDK-Android/blob/master/SECURITY.md).
- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/Parse-SDK-Android/issues/1212).

### Issue Description
Current AndroidGradlePlugin is 7.1.3.

Closes: [#1212](https://github.com/parse-community/Parse-SDK-Android/issues/1212).

### Approach
Upgrade AGP and gradle-wrapper and modify code syntax accordingly.

### TODOs before merging
- [x] No tests required.
- [x] No documentation changes required.

PR #1218, opened 2024-08-18
fix: Tests failing due to disabling JUnitPlatform which is causing 0% coverage reports

abubkr-hago:

- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/Parse-SDK-Android/issues/1217).

### Issue Description
Some tests are failing after disabling JUnitPlatform.

Closes: [#1217](https://github.com/parse-community/Parse-SDK-Android/issues/1217).

### Approach
Fix failing tests without enabling JUnitPlatform.

### TODOs before merging

- [x] Fix ParseCountingUriHttpBodyTest.testWriteToWithNullOutput
- [x] Fix ParseCountingUriHttpBodyTest.testWriteTo
- [x] Fix ParseFileControllerTest.testSaveAsyncSuccessWithUri
- [x] Fix ParseFileTest.testSaveAsyncSuccessWithUri
- [x] Fix ParseUriHttpBodyTest.testInitializeWithUri
- [x] Fix ParseUriHttpBodyTest.testWriteTo
- [ ] Fix ParseCorePluginsTest.testQueryControllerDefaultImpl
- [ ] Fix ParseDecoderTest.testCompleteness
- [ ] Fix ParseDecoderTest.testCompletenessOfIncludedParseObject

### DIFF
--- a/parse/build.gradle
+++ b/parse/build.gradle
@@ -51,6 +51,7 @@ dependencies {
     api "com.squareup.okhttp3:okhttp:$okhttpVersion"
     api project(':bolts-tasks')
 
+    testRuntimeOnly('org.junit.vintage:junit-vintage-engine:5.10.2')
     testImplementation "org.junit.jupiter:junit-jupiter:$rootProject.ext.jupiterVersion"
     testImplementation "org.skyscreamer:jsonassert:1.5.0"
     testImplementation "junit:junit:$rootProject.ext.junitVersion"


mtrezza:
I think you have the wrong base for this PR, the base branch does not seem to be up-to-date?

abubkr-hago:
> I think you have the wrong base for this PR, the base branch does not seem to be up-to-date?

Yes, I continued working on the same branch.
I'll merge them once all tests pass.

abubkr-hago:
> Maybe we were using JUnit 5, so removing `useJUnitPlatform` now causes the tests to fail, even though they only fail because we now run on JUnit 4 instead of 5. So maybe the solution is not to fix the tests, but to find out why `useJUnitPlatform` did not execute the tests properly?

This seems to be a better solution, downgrading JUnit would probably cause more issues later on.

mtrezza:
I'm not sure whether the PR has effectively downgraded it, because of the 2 dependencies. Do you see which JUnit version the tests are running with?

abubkr-hago:
> I'm not sure whether the PR has effectively downgraded it, because of the 2 dependencies. Do you see which JUnit version the tests are running with?

It did, the current version is `4.13.2`, you can print it using `junit.runner.Version.id()`.

abubkr-hago:
**Good News**, I got a 1% coverage report after refactoring imports to `org.junit.jupiter.api` in one of the test files.

It seems we will need to refactor all test files to use `org.junit.jupiter` instead of `org.junit`.

abubkr-hago:
> **Good News**, I got a 1% coverage report after refactoring imports to `org.junit.jupiter.api` in one of the test files while useJUnitPlatoform is enabled.
> 
> It seems we will need to refactor all test files to use `org.junit.jupiter` instead of `org.junit`.

It turns out only one runtime can be used when testing, meaning this and the previous builds were running JUnit4 all the time.
But now JUnitPlatform doesn't ignore JUnit4 tests.
Also, we can't upgrade to JUnit5 since Robolectric doesn't support JUnit5 runtime.

mtrezza:
Interesting; so what is the solution? Stick with JUnit 4 for now?

abubkr-hago:
> Interesting; so what is the solution? Stick with JUnit 4 for now? Does it require a lot of changes?

I think we will have to stick with JUnit4 for now, coverage is still not reporting though, look at `Run tests` [here](https://github.com/parse-community/Parse-SDK-Android/actions/runs/10448178282/job/28968671075?pr=1218).
I'll fix this ASAP.

mtrezza:
Sounds good; could you remove the unrelated changes from this PR, so we can merge it?

abubkr-hago:
> Sounds good; could you remove the unrelated changes from this PR, so we can merge it?

Do you mean remove the changes related to upgrading different packages? and keep the changes related to fixing the failing tests?

There are still 3 failing tests I'm trying to figure out:
- [ ] Fix ParseCorePluginsTest.testQueryControllerDefaultImpl
- [ ] Fix ParseDecoderTest.testCompleteness
- [ ] Fix ParseDecoderTest.testCompletenessOfIncludedParseObject

mtrezza:
> Do you mean remove the changes related to upgrading different packages? and keep the changes related to fixing the failing tests?

Yes, only keep changes related to fixing coverage.

abubkr-hago:
`junit-vintage-engine` runtime fixes 0% coverage reporting issue, should I include fixing the failing tests in this PR?

This PR mentions this issue [#1217](https://github.com/parse-community/Parse-SDK-Android/issues/1217).

mtrezza:
Oh nice, with that, can we mix Junit 4 and 5 tests? Because as some point I guess we'll have to start migrating, or maybe require newly added tests to be written already in JUnit 5.

abubkr-hago:
Yes both JUnit4 and Jupiter tests can be mixed now.

mtrezza:
Very nice!

KutayKerem:
Whatsapp send +90 5461395461

10 Eyl 2024 Sal 15:59 tarihinde Manuel ***@***.***> şunu
yazdı:

mtrezza:
Could you take a look at the failing CI?

mtrezza:
Re-running the CI...

abubkr-hago:
Sorry for the late response, I got a job and got very busy.
I actually couldn't find what these errors are coming from. Is it because of the JVM or the OS of github VMs.

mtrezza:
If we could get this PR over the finish line, we could also merge https://github.com/parse-community/Parse-SDK-Android/pull/1213; @azlekov do you have any time to take a look at why this is failing?

azlekov:
Hello, @mtrezza let's have first the #1213 merged. Probably after that the project can be configured via https://github.com/mannodermaus/android-junit5 this plugin to have both JUnit 4 and JUnit 5. This plugin requires AGP to be higher then 8.0.0

Another approach would be to downgrade the tests to JUnit 4 but I remember when I was doing some version bump I migrate to JUnit 5 because I thought it should be supported well and because of some issues with the tests. Looks like JUnit 4 is still preferable for Android setup.

mtrezza:
Oh the intention was the other was around I believe. The other one has codecov disabled, so this one was opened, I believe?

